PR #1635, opened 2015-01-06
Fixed a bug when there are no specific_assets in the current bunch of assets

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1407734. The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/948

Note 1: Vitor will provide a QA test exercising this case in the near future
Note 2: the management of the specific assets should be moved to the base risk calculators class, and that will be done in the future, when the much needed refactoring of the risk calculators will be started


ptormene:
LGTM

PR #1637, opened 2015-01-08
Removed compute_hazard_maps which is now in commonlib

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1408319. The tests are running here:
https://ci.openquake.org/job/zdevel_oq-engine/955/


ptormene:
If tests are green, LGTM

PR #1643, opened 2015-01-26
Removed hypocenter field from the transfer to icebox

micheles:
This was causing an error when running the scenario damage demo


daniviga:
LGTM

PR #1645, opened 2015-01-27
Don't trap too much errors in the engine server

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1414044. Also, I removed the SES table from the database transfer, since it cannot be visualized anyway.
The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/973


daniviga:
LGTM

PR #1649, opened 2015-02-04
Supporting both Ubuntu 12.04 and Ubuntu 14.04

micheles:
It is possible to support Ubuntu 12.04 by simply upgrading django to the same version in Ubuntu 14.04, i.e. 1.6.1. Celery, rabbitmq, psutil, numpy and all the related dependencies can stay the same.
Here is a full run with Ubuntu 12.04 + Django 1.6.1: https://ci.openquake.org/job/zdevel_oq-engine/1006.
See https://bugs.launchpad.net/oq-engine/+bug/1339604.


daniviga:
A test of this branch on Ubuntu 12.04 (with Django 1.6.1 installed manually) is running here: https://ci.openquake.org/job/zdevel_oq-engine/986/console

daniviga:
LGTM

micheles:
I thought there was a problem with dbsettings?

daniviga:
No, I did a reset so I lost all the database settings apart from the default `localhost` that was causing the error. Putting the right settings fixed the issue.
So is a LGTM, we need just to be careful with the `DJANGO_SETTINGS_MODULE` if it has been already set in the system environment.

PR #1656, opened 2015-02-12
Change the 'default' Django db to use cfg settings.

daniviga:
In this way we always use only one method to connect to PostgreSQL (via TCP) instead of mixing TCP and sockets.

This conforms better to Ubuntu where connections to `localhost` are already set with `md5` authentication in the `pg_hba.conf`.

Tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1014/


micheles:
LGTM if the tests are green

PR #1659, opened 2015-02-16
Export the GMFs in csv format

micheles:
Closes https://bugs.launchpad.net/oq-engine/+bug/1409620. Notice that the file names have been changed to avoid collisions in the case of oversampling. Moreover, now the TRT numbering starts from zero for consistency with oq-lite.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1029


openquake/engine/export/hazard.py
```diff
@@ -100,55 +101,31 @@ def _get_result_export_dest(calc_id, target, result, file_ext='xml'):
             ltr = result.lt_realization
```
ptormene: This block looks exactly the same as the case of `output_type == 'gmf'`

ptormene:
Please fix the 5 broken tests

micheles:
I have fixed the tests

ptormene:
LGTM

PR #1661, opened 2015-02-17
Fixed a bug found by Chris Schneider

micheles:
https://ci.openquake.org/job/zdevel_oq-engine/1028

### DIFF
--- a/openquake/engine/calculators/risk/event_based_risk/core.py
+++ b/openquake/engine/calculators/risk/event_based_risk/core.py
@@ -86,7 +86,7 @@ def event_based(workflow, getter, outputdict, params, monitor):
             if specific_assets:
                 loss_matrix, assets = _filter_loss_matrix_assets(
                     out.output.loss_matrix, out.output.assets, specific_assets)
-                if assets:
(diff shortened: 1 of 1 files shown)


openquake/engine/calculators/risk/event_based_risk/core.py
```diff
@@ -86,7 +86,7 @@ def event_based(workflow, getter, outputdict, params, monitor):
             if specific_assets:
                 loss_matrix, assets = _filter_loss_matrix_assets(
                     out.output.loss_matrix, out.output.assets, specific_assets)
-                if assets:
+                if len(assets):
```
daniviga: This means that `assets` cannot be `None`?

larsbutler: Yep; if `assets` is None, `len(assets)` will raise a `TypeError`.

daniviga:
LGTM

micheles:
The problem was that assets is a numpy array, so the syntax `if assets` gives an error. In the companion PR I have changed the risk demo so that the bug is exercised.

PR #1673, opened 2015-02-27
Remove stale code from debian/postinst

daniviga:
This `for` is a bit strange. It wants a folder which is unused (`/var/lib/openquake`) and tries to use a group `openquake` which does not exist yet.


micheles:
LGTM

PR #1675, opened 2015-03-02
Save only non zero losses in the table event_loss_asset

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1427226
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1074
Notice the companion https://github.com/gem/oq-risklib/pull/167


raoanirudh:
LGTM

PR #1683, opened 2015-03-10
Move the log into the database

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1430206.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1089

### DIFF
--- a/openquake/engine/utils/tasks.py
+++ b/openquake/engine/utils/tasks.py
@@ -157,9 +157,9 @@ def oqtask(task_func):
         # the revoke functionality can work
         EnginePerformanceMonitor.store_task_id(job_id, tsk)
 
-        with EnginePerformanceMonitor(
-                'total ' + task_func.__name__, job_id, tsk, flush=True
-        ), logs.handle(job):
+        with (EnginePerformanceMonitor(
+                'total ' + task_func.__name__, job_id, tsk, flush=True),
+              logs.handle(job)):
             try:
                 # log a warning if too much memory is used
                 check_mem_usage(SOFT_MEM_LIMIT, HARD_MEM_LIMIT)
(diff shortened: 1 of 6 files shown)


openquake/engine/utils/tasks.py
```diff
@@ -157,10 +157,14 @@ def wrapped(*args):
         # the revoke functionality can work
         EnginePerformanceMonitor.store_task_id(job_id, tsk)
 
+        # if there is no db handler already, add one
+        if not any(isinstance(handler, logs.LogDatabaseHandler)
+                   for handler in logs.LOG.handlers):
+            logs.LOG.addHandler(logs.LogDatabaseHandler(job))
+            logs.set_level(job.log_level)
+
```
nastasi-oq: What happens when more than one job concur on the same worker ?
Add a test to check this behaviour could be a good choice.

openquake/engine/utils/tasks.py
```diff
@@ -158,9 +158,8 @@ def wrapped(*args):
         EnginePerformanceMonitor.store_task_id(job_id, tsk)
 
         with EnginePerformanceMonitor(
-                'total ' + task_func.__name__, job_id, tsk, flush=True):
-            # tasks write on the celery log file
-            logs.set_level(job.log_level)
+                'total ' + task_func.__name__, job_id, tsk, flush=True
+        ), logs.handle(job):
             try:
```
nastasi-oq: Use a couple of () to group the 2 with statements seems to me more correct.

nastasi-oq:
LGTM

openquake/engine/utils/tasks.py
micheles: I have change the approach to use a with statement which is cleaner and safer

openquake/engine/utils/tasks.py
micheles: Done

PR #1685, opened 2015-03-10
Not using Django when deleting calculations

micheles:
This solves the out of memory issue which was reported by our users (https://bugs.launchpad.net/oq-engine/+bug/1430360).
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1092


daniviga:
LGTM, and makes me happy (if tests are green)

PR #1686, opened 2015-03-11
New API for listing calculations

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1430395.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1096

### DIFF
--- a/openquake/server/tests/views_test.py
+++ b/openquake/server/tests/views_test.py
@@ -60,31 +60,34 @@ class UtilsTestCase(BaseViewTestCase):
 class CalcHazardTestCase(BaseViewTestCase):
 
     def setUp(self):
-        self.request = self.factory.get('/v1/calc/hazard/')
+        self.request = self.factory.get('/v1/calc/list?job_type=hazard')
         self.request.META['HTTP_HOST'] = 'www.openquake.org'
 
     def test_get(self):
         expected_content = [
             {u'url': u'http://www.openquake.org/v1/calc/1',
              u'status': u'executing',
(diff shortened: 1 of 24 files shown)


openquake/server/tests/views_test.py
```diff
@@ -103,26 +106,29 @@ def test_get(self):
...
+                (3, 'complete', 'risk', 'description e'),
             ]
-            request = self.factory.get('/v1/calc/risk/')
+            request = self.factory.get('/v1/calc/list?job_type=risk')
             request.META['HTTP_HOST'] = 'www.openquake.org'
-            response = views.calc(request, 'risk')
+            response = views.calc(request)
 
             self.assertEqual(200, response.status_code)
             self.assertEqual(expected_content, json.loads(response.content))
```
nastasi-oq: No tests for relevant attribute ?

micheles: The problem is that current tests are mocking completely the calc method, so they do not check nothing of it, including the `relevant` attribute. The current tests are dangerous more than useless and I plan to remove them ( https://bugs.launchpad.net/oq-engine/+bug/1393733). This will be done in a separated pull request, though.

daniviga:
I've tested this code with the current status of Icebox and it does not break the compatibility, so LGTM on that side.

nastasi-oq:
LGTM

daniviga:
I'm going to test the two latest commits against current platform

micheles:
They have no impact on the platform, they are related to the API which is not used by the platform.

PR #1701, opened 2015-03-25
Make it impossible to run classical_risk from classical_tiling

micheles:
Unfortunately, the risk does not work with an ugly error. So let's forbid the user to do that in the first place.
The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/1123


ptormene:
LGTM

PR #1704, opened 2015-03-27
Added is_running information to the calculation list

micheles:
This is a small extension requested by Matteo. The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1134

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,6 @@
   [Michele Simionato]
+  * Now the URL /v1/calc/list returns a list of dictionaries with a boolean
+    key is_running
   * Replaced the mock tests for the engine server with real functional tests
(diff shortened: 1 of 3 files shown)


debian/changelog
```diff
@@ -1,4 +1,6 @@
   [Michele Simionato]
+  * Now the URL /v1/calc/list returns a list of dictionaries with a boolean
+    key is_running
```
nastasi-oq: Maybe is better something like: a new key ```is_running``` is added to the list of dictionaries returned by ...
to emphasize the new part in a more clear way.

nastasi-oq:
Except the comment LGTM

debian/changelog
micheles: Done

PR #1709, opened 2015-03-31
Skipped the broken scenario_damage test

micheles:
While implementing the oq-lite tests I discovered a bug in the engine tests, so they are skipped for the moment. This is required for merging https://github.com/gem/oq-risklib/pull/194.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1141


micheles:
I found a way to fix the tests.

PR #1714, opened 2015-04-02
Set engineserver DEBUG = True as default

daniviga:
All test pass now.

Tests: https://ci.openquake.org/job/zdevel_oq-engine/1152/


micheles:
LGTM

PR #1715, opened 2015-04-02
is_running=True is now the default

micheles:
This should solve https://bugs.launchpad.net/oq-engine/+bug/1438112.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1153/


daniviga:
Tested and solved the issue with multiple job queued. @nastasi-oq does it look good to you?

nastasi-oq:
The case where for some reason the machine crashes and than the job seems to be running forever is covered (at least when a new engine instance is runned) ?

micheles:
Yes, when restarting the engine server is_running is set to False automatically.

daniviga:
LGTM

nastasi-oq:
LGTM

PR #1716, opened 2015-04-02
Missing logictree file

micheles:
In this case the engine server was failing to return a proper traceback. This now fixed and tested, see https://ci.openquake.org/job/zdevel_oq-engine/1154


nastasi-oq:
LGTM

PR #1717, opened 2015-04-02
Engineweb HTML improvements

daniviga:
Added header, footer and license. Lots of cleanup.
screenshot from 2015-04-03 15 17 06
screenshot from 2015-04-03 15 17 28
screenshot from 2015-04-03 15 17 45

Best viewed with: https://github.com/gem/oq-engine/pull/1717/files?w=1

PR #1722, opened 2015-04-08
engineweb: add new static content to setup.py

daniviga:
New static files need to be added after https://github.com/gem/oq-engine/pull/1717


nastasi-oq:
lil OT: could we add the favico.png file to avoid missing download from the server ?

nastasi-oq:
LGTM

PR #1724, opened 2015-04-08
webui: add authentication support for the engine server

daniviga:
This is meant to be used mainly with the Engine Web UI. Authentication is not (yet?) supported by `icebox`, but the changes are backward compatible unless `LOCKDOWN = True`.

Features:
- Can be enabled setting `LOCKDOWN = True` in `settings.py` or `local_settings.py`
- It uses a separate SQLite DB for storing auth and sessions instead of the Engine PostgreSQL DB
- PAM can be used as auth backend (depends on `django-pam`)
- It supports `AUTH_EXEMPT_URLS` to disable auth on a list of urls
- Jobs are submitted with the authenticated user, otherwise "platform" is used[1]
- Calculation list is filtered upon authenticated user, otherwise all calculation are showed (this applies also to the API)
- Outputs list is also filtered
- Application is 100% backward compatible if `LOCKDOWN = False`

Todo:
- Show all calculation if user is `superuser`
- Get only outputs owned by the authenticated user: providing an output id manually to the API gives other users outputs[2]

Before using authentication the SQLite database must be created with:
```bash
$ python manage.py syncdb  --database=auth_db
```
If the authentication is not enabled this step is not needed.

This allows the UI to be installed on a server/cluster with multi-user support.

---
[1] this is to allow backward compatibility with `icebox`
[2] this is how oq-engine works, it is not really an issue of the UI


daniviga:
Description updated

micheles:
Remember the changelog!

daniviga:
Done, thanks @micheles

daniviga:
This PR can be tested here: https://oats3-ui.openquake.org/

openquake/server/views.py
```diff
@@ -180,7 +179,12 @@ def calc(request, id=None):
     """
     base_url = _get_base_url(request)
 
-    calc_data = _get_calcs(request.GET, id=id)
+    user_name = "platform"
+    if hasattr(request, 'user'):
+        if request.user.is_authenticated():
+            user_name = request.user.username
```
micheles: It can be done in one line:
```user_name = request.user.username if hasattr(request, 'user') and request.user.is_authenticated() else 'platform'```

File: openquake/server/views.py [hunk 2]
```diff
@@ -368,10 +381,16 @@ def calc_results(request, calc_id):
...
+
     # If the specified calculation doesn't exist OR is not yet complete,
     # throw back a 404.
     try:
-        oqjob = oqe_models.OqJob.objects.get(id=calc_id)
+        oqjob = oqe_models.OqJob.objects.get(id=calc_id,
+                                             user_name=user_name)
```
micheles: What happens if the user is looking at a calculation of another user? `oqe_models.OqJob.objects.get` gives an ugly DoesNotExist error. We should probably do

```
oqjob = oqe_models.OqJob.objects.get(id=calc_id)
if oqjob.username != username:
      return ...
```

openquake/server/views.py
micheles: Or better define a helper function get_username(request).

micheles:
Apart the minor suggestion, LGTM

openquake/server/views.py
daniviga: Fixed

PR #1727, opened 2015-04-09
Fixed a bug affecting celery 2 only when nose is used to run the tests

micheles:
This seems to work. The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1178/

### DIFF
--- a/openquake.cfg
+++ b/openquake.cfg
@@ -18,7 +18,7 @@ terminate_workers_on_revoke = true
 # this is good for a single user situation, but turn this off on a cluster
 # otherwise a CTRL-C will kill the computations of other users
 
-terminate_job_when_celery_is_down = true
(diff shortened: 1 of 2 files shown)


openquake.cfg
```diff
@@ -18,7 +18,7 @@ terminate_workers_on_revoke = true
 # this is good for a single user situation, but turn this off on a cluster
 # otherwise a CTRL-C will kill the computations of other users
 
-terminate_job_when_celery_is_down = true
+terminate_job_when_celery_is_down = false
```
daniviga: Why the default is now `false`?

micheles: Argh, this was accidentally committed

daniviga:
LGTM

PR #1732, opened 2015-04-12
Fixed zero-filled curves

micheles:
Even if hazard_curves_from_gmfs was false, hazard curves were generated, filled with zeros.
Tests are green: https://ci.openquake.org/job/zdevel_oq-engine/1181/


daniviga:
LGTM

PR #1733, opened 2015-04-12
webui: make css more bootstrap compliant

daniviga:
This will 
- make more easy to reuse the new custom css code
- allow better buttons alignment
- remove color change on hover on the second level `<td>`

from
screenshot from 2015-04-12 12 30 53

to
screenshot from 2015-04-12 12 22 54


nastasi-oq:
LGTM

PR #1746, opened 2015-05-04
Special case when saving GSIMs into the database

micheles:
This is needed since I change the gsim parameter from a string into a GSIM instance in https://github.com/gem/oq-risklib/pull/227/
Tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1206


daniviga:
LGTM

PR #1747, opened 2015-05-05
Fixed the --load_gmf command

micheles:
The maximum distance is mandatory, so it must be set, even if to a fake value.
In a subsequent pull request I will add a test for the feature, but I have already checked manually that the fix works. See https://bugs.launchpad.net/oq-engine/+bug/1451853


nastasi-oq:
LGTM

PR #1750, opened 2015-05-06
debian/control: fix dependencies order

daniviga:
Fixes an issue with dependencies order and segfault of `librabbitmq`.
This error can be reproduced only on "non-fast" LXC in CI.

Bad: https://ci.openquake.org/job/zdevel_oq-engine_nofast/1/
Good: https://ci.openquake.org/job/zdevel_oq-engine_nofast/2/


micheles:
LTGM

PR #1752, opened 2015-05-07
Pkgdev rename

nastasi-oq:
Dependencies version updated.
Different development package name suffix composition.
More flexible mangling of control dependencies versions in development packaging (broken before).


daniviga:
LGTM

PR #1754, opened 2015-05-11
server: default user is now stored in settings

daniviga:
I need this to use the "openquake" users on the VMs.

I also don't like to have a username in the code...


micheles:
LGTM

micheles:
LGTM

PR #1757, opened 2015-05-13
webui: show all calculations if user is superuser

daniviga:
This PR allows a superuser to see all the calculations and outputs from all the users. If user is superuser a new column is added with the calculation owner.

### DIFF
--- a/openquake/server/utils.py
+++ b/openquake/server/utils.py
@@ -4,11 +4,12 @@ from django.conf import settings
 from openquake.engine import __version__ as oqversion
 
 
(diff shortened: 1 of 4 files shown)


openquake/server/utils.py
```diff
@@ -11,11 +11,17 @@ def getusername(request):
     compatibility.
     """
```
micheles: I would change the name from get_username to get_user since now it returns more than a name

micheles:
Apart from the minor comment. LGTM

PR #1758, opened 2015-05-13
Ci interpkg consist2

nastasi-oq:
More sophisticated behavior for package dependencies resolution:
```
in pkgtest phase, FOR EACH dependency:
    IF exists a dependency branch with the same name of the running job:
         IF the package associated with the dependency commit used during devtest isn't yet created:
             for one hour, every 10 secs, try to find it
    ELSE:
        try to use the latest master version of dependency
```

### DIFF
--- a/packager.sh
+++ b/packager.sh
@@ -241,7 +241,7 @@ _pkgbuild_innervm_run () {
 }
 
 #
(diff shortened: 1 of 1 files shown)


daniviga:
LGTM

PR #1760, opened 2015-05-15
Specified concurrent_tasks explicitly

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1455342. A refactoring of few weeks ago made OqTaskManager to inherits the default value for concurrent_tasks from oq-lite (i.e. the number of cores): so, instead of the concurrent_tasks defined in openquake.cfg the hazard calculators were using the number of cores of the controller machine. The solution is to not use the default.

You can see that the fix works by looking at the demos logs here: https://ci.openquake.org/job/zdevel_oq-engine/1228/console


daniviga:
LGTM

PR #1761, opened 2015-05-26
Datastore integration

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1448577.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1238
The feature is still temporary and this is why there are no automatic tests for it. It can be tried by
running `$ oq-engine --lite`.


daniviga:
LGTM

PR #1762, opened 2015-05-28
Improve the python packaging

daniviga:
Making RPM packages I discovered some issues with the Python packaging and the `setup.py`. Apart from outdated information, we were producing packages (even Ubuntu ones) with missing pieces to run the tests: some time ago we moved the tests inside the `openquake.engine` python package, but by default only `.py` files are kept in the package. The `MANIFEST.in` allows us to include all the data needed to run tests (csv, xml...) located in specific folders.
A similar issue was already solved with the engine server static files, but we did it in the bad way: instead of adding them in a `MANIFEST.in` we forced them to be part in the `openquake.server` package using the `package_data=` option in `setup.py`. This should be used only to include in a package files that are out of the scope (aka out of the `openquake` folder).

I also moved the `celeryconfig.py` from `/usr/openquake/engine` to `/usr/share/openquake/engine` to be more compliant with the UNIX FHS.

Full suite of tests for both precise and trusty: https://ci.openquake.org/job/zdevel_oq-engine/1245/

_Please note:_ https://github.com/gem/oq-engine/commit/d72ab988e041eafd72464cd586bab1a4fc62306f must be kept, otherwise demos will not run when binary package is used

See also: https://github.com/gem/oq-hazardlib/pull/339 and https://github.com/gem/oq-risklib/pull/262

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -87,6 +87,7 @@ setup(
     package_data={"openquake.engine": [
         "openquake.cfg", "openquake_worker.cfg",
         "README.md", "LICENSE", "CONTRIBUTORS.txt"]},
+    namespace_packages=['openquake'],
     scripts=["openquake/engine/bin/oq_create_db",
              "openquake/engine/bin/openquake"],
     zip_safe=False,
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ setup(
     maintainer_email='devops@openquake.org',
     description=("Computes hazard, risk and socio-economic impact of "
                  "earthquakes."),
-    license="GNU AGPL v3",
+    license="AGPL3",
     keywords="earthquake seismic hazard risk",
     url=url,
     long_description=README,
--- a/setup.py
+++ b/setup.py
@@ -76,16 +76,12 @@ setup(
                                     "openquake.engine.bin",
                                     "openquake.engine.bin.*"]),
     py_modules=PY_MODULES,
-
     include_package_data=True,
     scripts=["openquake/engine/bin/oq_create_db",
              "openquake/engine/bin/openquake"],
-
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -20,18 +35,18 @@ def get_version():
...
 OpenQuake is an open source application that allows users to
 compute seismic hazard and seismic risk of earthquakes on a global scale.
-
-Please note: the /usr/bin/openquake script requires a celeryconfig.py
+Please note: the /usr/bin/oq-engine script requires a celeryconfig.py
 file in the PYTHONPATH.  Please make sure this is the case and that your
 celeryconfig.py file works with your python-celery setup.
-
-Feel free to copy /usr/openquake/engine/celeryconfig.py and revise it
+Feel free to copy /usr/share/openquake/engine/celeryconfig.py and revise it
```
micheles: I would put this comment also in the

```python
try:
     import celeryconfig
    ...
```
block in openquake_cli.py

daniviga:
Manual testing on CentOS 7: PASSED

micheles:
What about those 3 scenario damage tests that were broken yesterday?

daniviga:
I'm still investigating, they are broken. But this PR is not RHEL 7 specific. It's needed also by Ubuntu packages. Currently we are producing packages with a number of missing files; if you want to run tests using the binary distribution you can't because csv, xml, ini ecc files in the tests folder are missing.

daniviga:
See https://github.com/gem/oq-engine/pull/1765

micheles:
LGTM

daniviga:
I want to merge together with the companions PR in risklib and hazardlib: gem/oq-hazardlib#339 and gem/oq-risklib#262

PR #1774, opened 2015-06-14
Fixed the datastore export

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1463859.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1270
The remaining problem is that the "view" button should not appear if the export is a zipfile.
I am opening a different bug for that: https://bugs.launchpad.net/oq-engine/+bug/1464947

### DIFF
--- a/openquake/engine/export/core.py
+++ b/openquake/engine/export/core.py
@@ -39,13 +39,11 @@ def zipfiles(fnames, archive):
 
     :param fnames: list of path names
     :param archive: path of the archive
-    :returns: the given archive path
     """
     z = zipfile.ZipFile(archive, 'w')
     for f in fnames:
         z.write(f)
     z.close()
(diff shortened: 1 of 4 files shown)


openquake/engine/export/core.py
```diff
@@ -28,11 +29,47 @@
...
+
+    :param fnames: list of path names
+    :param archive: path of the archive
+    :returns: the given archive path
+    """
+    z = zipfile.ZipFile(archive, 'w')
+    for f in fnames:
+        z.write(f)
+    z.close()
+    return archive
```
ptormene: If you want, you can also return None

micheles: Done

ptormene:
LGTM

PR #1781, opened 2015-06-25
Add configuration options to switch the broker/backend to Redis

daniviga:
but keep them commented.

To use Redis you have also to install `redis-server` and `python-redis`.

test (using `amqp`): https://ci.openquake.org/job/zdevel_oq-engine/1297/


openquake/engine/utils/tasks.py
```diff
@@ -67,6 +67,7 @@ def aggregate_result_set(self, agg, acc):
         if not self.results:
             return acc
         backend = current_app().backend
+        amqp_backend = backend.__class__.__name__.startswith('AMQP')
```
daniviga: There isn't a more clear way to check this?

micheles: I prefer to do this, since the AMQP class has changed location between celery 2 and 3, so we would need a conditional import. This is a temporary hack anyway, we can improve on that when we drop the support for celery 2.

daniviga: Ok, LGTM

PR #1784, opened 2015-06-28
The engine does not eat errors anymore

micheles:
Solves https://bugs.launchpad.net/oq-engine/+bug/1469124.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1301

### DIFF
--- a/openquake/engine/engine.py
+++ b/openquake/engine/engine.py
@@ -109,23 +109,34 @@ def job_stats(job):
     js = job.jobstats
     try:
         yield
-    except:
-        conn = django_db.connections['job_init']
-        if conn.is_dirty():
-            conn.rollback()
-        raise
     finally:
+        tb = traceback.format_exc()  # get the traceback of the error, if any
         job.is_running = False
-        job.save()
+        if tb != 'None\n':
+            # rollback the transactions; unfortunately, for mysterious reasons,
+            # this is not enough and an OperationError may still show up in the
+            # finalization phase when forks are involved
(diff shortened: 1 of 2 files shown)


openquake/engine/engine.py
```diff
@@ -109,23 +109,34 @@ def job_stats(job):
...
         job.is_running = False
-        job.save()
-
-        # save job stats
-        curs.execute("select pg_database_size(%s)", (dbname,))
-        new_dbsize = curs.fetchall()[0][0]
-        js.disk_space = new_dbsize - dbsize
-        js.stop_time = datetime.utcnow()
-        js.save()
+        if tb != 'None\n':
```
daniviga: Just a curiosity, why newline is needed at the end?

openquake/engine/engine.py
```diff
@@ -109,23 +109,34 @@ def job_stats(job):
...
+            # rollback the transactions; unfortunately, for mysterious reasons,
+            # this is not enough and an OperationError may still show up in the
+            # finalization phase when forks are involved
+            for conn in django_db.connections.all():
+                conn.rollback()
```
daniviga: Before `conn` was taken from `django_db.connections['job_init']`; now this has been removed. Where `conn` is from?

daniviga:
LGTM

openquake/engine/engine.py
micheles: I saw that `traceback.format_exc()` returns `None\n` when there is no traceback.

PR #1786, opened 2015-06-29
Fixes classical.test_case_15

micheles:
This is due to the change of wildchar character in the file names from "*" to "@". Solves the breakage in Jenkins: https://ci.openquake.org/job/master_oq-engine/1837/console


micheles:
LGTM

PR #1790, opened 2015-07-01
Fixed error message in check_limits

micheles:
This was reported by an OATS user. See https://bugs.launchpad.net/oq-engine/+bug/1470416


daniviga:
LGTM

PR #1796, opened 2015-07-13
lon and lat are now rounded to 5 digits

micheles:
See comment in https://bugs.launchpad.net/oq-engine/+bug/1473051.
This is the companion of https://github.com/gem/oq-risklib/pull/326.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1349


raoanirudh:
LGTM

PR #1799, opened 2015-07-28
Initial support for RPM packages

daniviga:
This PR provides a complete spec file to generate an RPM package for EL >= 7 and Fedora >= 21,

The `rpm/makerpm.sh` generates a SRPM file from a spec template and the current repo `HEAD` using a `mock` chroot. The generated SRPM can be built locally (using `makerpm.sh -l`) or using Fedora COPR.

No tests are needed on Ci since only new files have been added.

See: https://copr.fedoraproject.org/coprs/gem/openquake/build/106005/


rpm/makerpm.sh
```diff
@@ -0,0 +1,70 @@
...
+            shift
+            ;;
+    esac
+done
+
+cd $BASE
+mkdir -p build-rpm/{RPMS,SOURCES,SPECS,SRPMS}
+
+LIB=$(cut -d "-" -f 2 <<< $REPO)
+SHA=$(git rev-parse --short HEAD)
```
daniviga: This should be ```$BRANCH``` instead of ```HEAD```

nastasi-oq:
Take a look at: gem/oq-hazardlib#362

nastasi-oq:
LGTM

PR #1800, opened 2015-07-28
Fixed a broken export test

micheles:
The problem is that now the engine exports the GMFs in NRML 0.5, due to the change in https://github.com/gem/oq-risklib/pull/331. This was breaking Jenkins: https://ci.openquake.org/job/master_oq-engine/1891/


micheles:
LGTM

PR #1812, opened 2015-08-31
Fixed the --load-curve command

raoanirudh:
When loading a hazard curve using the `--load-curve` command, the `maximum_distance` parameter is not set, even though it is a mandatory parameter. This fix sets the parameter `maximum_distance` to a default value of 1000 km, which is the same value used when loading a gmf using the `--load-gmf` command.


micheles:
LGTM

PR #1813, opened 2015-08-31
packager.sh: checks on the debian/rules file did not work on dependencies

daniviga:
The check on the `debian/rules` dependencies did not work correctly for the GIT dependencies. We never found this error because the GIT deps (hazardlib, risklib) do not use that file to select packages dynamically, but the error was discovered with the `oq-risk-tests` where also the engine is treated as a GIT dependency.

Tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1406/ (is red due a bug in the XUnit reports, but the fix worked)


nastasi-oq:
LGTM

PR #1814, opened 2015-08-31
packager.sh: use the proper celery command based on the celery version

daniviga:
Looking at the Trusty CI outputs I discovered that the `celeryd_wait` did not work with Celery 3 because the `celeryctl` command has been replaced by `celery`. Luckily celery has been always fast enough to start before it was actually used (because we run some tests that are not using it before running any computation)... until now: with `oq-risk-tests` celery is needed by the very first test, thus we need to wait it to be ready. 

Tests: https://ci.openquake.org/job/zdevel_oq-engine/1407/

### DIFF
--- a/packager.sh
+++ b/packager.sh
@@ -335,7 +335,7 @@ celeryd_wait() {
     if command -v celeryctl &> /dev/null; then
         # celery 2.4
         celery=celeryctl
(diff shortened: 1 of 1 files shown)


nastasi-oq:
LGTM

PR #1821, opened 2015-09-10
Increase the engine server warmup time in the tests

daniviga:
To avoid broken builds like \#1970 - \#1972 (https://ci.openquake.org/job/master_oq-engine/)


micheles:
LGTM

PR #1822, opened 2015-09-11
Improved view_mean_avg_losses

micheles:
We also print the total losses if there is more than one assets, in accordance with what it is done in the Julia calculation (https://github.com/gem/oq-risk-tests/pull/7). This requires https://github.com/gem/oq-risklib/pull/364/
The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/1426/


ptormene:
LGTM

PR #1830, opened 2015-09-17
oq-engine --lite does not work with two files

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1496390.


raoanirudh:
LGTM

PR #1831, opened 2015-09-17
update for near-fault PSHA calculation

vup1120:
The pull request adding the near-fault PSHA calculation part into the release note 1.5


micheles:
I would change this to

```
Added near-fault directivity probabilistic seismic hazard
analysis for classical PSHA calculations with simple fault
sources. We implemented the most recent NGA-WEST2
directivity model and the associated GMPE which is, up to now,
the only GMPE model explicitly including the effect.  More
details can be found in the manual.
```

vup1120:
Thanks Michele! It has been updated.

micheles:
LGTM

PR #1834, opened 2015-09-21
Fixed save_sites in the case of multiple assets on the same point

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1497898.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1435/


raoanirudh:
LGTM

PR #1835, opened 2015-09-21
Simplified the GMF export test

micheles:
Producing less GMFs, so that the expected file is smaller and the debugging is easier.
This branch requires the companion https://github.com/gem/oq-risklib/pull/387.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1438/


ptormene:
LGTM

PR #1837, opened 2015-09-21
Fix broken link to risklib changelog

raoanirudh:


doc/engine/What's-new.md
```diff
@@ -137,7 +137,7 @@ the functionality only works for hazard calculations
 but it is expected to grow in the future.
 
 13. Several other improvements have been made to oq-lite, too many to list
-them all here; please see the [changelog](https://raw.githubusercontent.com/gem/oq-risklib/engine-1.5/debian/changelog) for the complete list.
```
daniviga: I would prefer to keep this file linked to the engine-1.5 branch since it's the 1.5 "What's new" it should refer the 1.5 related risklib. @micheles what do you think?

micheles: Yes, the link is to the engine-1.5 branch on purpose, let's keep it.

PR #1849, opened 2015-10-01
doc: fix demos and celeryconfig paths

daniviga:
In the stable documentation we still have path referring to the old celeryconfig and demos locations.

This PR is against engine-1.5 and should then be merged into master. I don't think we'll need a new package release (at least, not just for this).


micheles:
LGTM

PR #1857, opened 2015-10-08
Turned off the self-termination "feature"

micheles:
Now Jenkins should not commit suicide as it happens frequently with an error

```python
openquake.engine.celery_node_monitor.MasterKilled: The openquake master process was killed by the CeleryNodeMonitor because some node failed
```


daniviga:
+1, LGTM

Issue #1862, opened 2015-10-10
The region_constraint in the association queries sometimes is not respected


micheles:
We discovered the problem in the reduced risk model for Italy, with ~80,000 assets on ~8,000 sites and a region constraint around Tuscany. The graph below has a red spot on the top border, in the center:
that it a location that PostGIS thinks is inside the region constrain, while in reality it is not, it is slightly
outside. oq-lite does the right thing and as a consequence it produces results for less assets than
the engine.

![reduced-italy](https://cloud.githubusercontent.com/assets/2463856/10409550/898eaf3a-6f25-11e5-96ed-0f6117ee83ca.png)

This looks like a bug in PostGIS. Fortunately, by rewriting the query it is possible to fix the problem.

PR #1863, opened 2015-10-10
Fixed the region_constraint management in the association queries

micheles:
Fix https://github.com/gem/oq-engine/issues/1862. The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/1471/

NB: I am also fixing the export order in the loss maps csv representation, to match the one in oq-lite
(lexicographic on the asset names).

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,6 @@
   [Michele Simionato]
+  * Fixed a very subtle bug in the association queries: some sites outside
+    of the region constraint were not discarded in some situations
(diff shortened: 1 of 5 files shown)


raoanirudh:
The individual asset losses exported from the engine now match those produced by oq-lite, and the assets are also listed in the same order for both.

The aggregate loss exported from oq-lite is incorrect; the aggregate loss exported from the engine is correct. Since this appears to be an issue unrelated to this PR, LGTM.

PR #1865, opened 2015-10-12
Fixed a wrong test

micheles:
The build https://ci.openquake.org/job/master_oq-engine/2037/ is broken because the ClassicalRiskExportTestCase is broken: there is a site which is outside of the region_constraint. Because of the long standing bug fixed in https://github.com/gem/oq-engine/pull/1863 that site was included, even if in reality it was outside the region! To get back the test to work, I am moving the site inside the region constraint.

The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1472


daniviga:
LGTM

PR #1867, opened 2015-10-16
server: fix celeryconfig.py path when server is installed by packages

daniviga:
The engine server, when installed by packages, is unable to find the `celeryconfig.py` giving this error when a calculation is launched:

```python
  File "/usr/lib/python2.7/dist-packages/openquake/engine/engine.py", line 218, in run_calc
    _do_run_calc(calculator, exports)
  File "/usr/lib/python2.7/dist-packages/openquake/engine/engine.py", line 257, in _do_run_calc
    result = calc.execute()
  File "/usr/lib/python2.7/dist-packages/openquake/calculators/scenario.py", line 101, in execute
    concurrent_tasks=self.oqparam.concurrent_tasks)
  File "/usr/lib/python2.7/dist-packages/openquake/commonlib/parallel.py", line 294, in apply_reduce
    return self.reduce(agg, acc)
  File "/usr/lib/python2.7/dist-packages/openquake/commonlib/parallel.py", line 377, in reduce
    agg_result = self.aggregate_result_set(agg_and_percent, acc)
  File "/usr/lib/python2.7/dist-packages/openquake/engine/utils/tasks.py", line 73, in aggregate_result_set
    for task_id, result_dict in rset.iter_native():
  File "/usr/lib/python2.7/dist-packages/celery/result.py", line 572, in iter_native
    return results[0].backend.get_many(
AttributeError: 'DisabledBackend' object has no attribute 'get_many'
```
that's a bit strange because a `celeryconfig.py` is present in `/usr/lib/python2.7/dist-packages/openquake/engine` which is a location well known to python. Probably we have an issue (again) with the "openquake" python package prefix.

I added to the server's `manage.py` the same code we are using in the `openquake_cli.py`: this snippet will add `/usr/share/openquake/engine` to the application `PYTHONPATH` (see the comment for more info)

```python
try:
     import celeryconfig
except ImportError:
     sys.path.append('/usr/share/openquake/engine')
```

To fix previous installations (without making a new stable release) this simple workaround solves the issue:

```bash
$ sudo ln -s /usr/share/openquake/engine/celeryconfig.py /usr/lib/python2.7/dist-packages/openquake/server/celeryconfig.py
```

PR #1869, opened 2015-10-16
server: move celeryconfig.py import in a better place

daniviga:
With the fix done by #1867 the Engine server in development mode (using sources) does not start anymore with the error

```python
CommandError: You must set settings.ALLOWED_HOSTS if DEBUG is False.
```
because the import of `celeryconfig` is messing up the `DJANGO_SETTINGS_MODULE` causing the Engine settings to be loaded instead of the Server ones (even if the import is after `os.environ.setdefault("DJANGO_SETTINGS_MODULE", "openquake.server.settings")`).

Moving the import to `tasks.py` solved both the issues.

Tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1478/


micheles:
LGTM if the tests pass

Issue #1870, opened 2015-10-18
Datastore does not work well with WebUI


daniviga:
New calculators which are using the 'datastore' feature does not work well with the WebUI in a server/multiuser installation.

Since in a multiuser installation the WebUI process is not started by the end user that is actually running the computation, the datastore is written in the webserver process owner's homedir instead of the calculation user's home. In any case the datastore cannot be created on the users' home because lack of privileges.

When an end user tries to run a risk computation on top of a previously calculated hazard it will fails because it tries to find the datastore in the user's home, instead the data is stored in the WebUI process owner's home.

Calculation run through the WebUI should check for the datastore in the process owner's home and then in the user's home, even if this could break cross-compatibility with the command line: previously it was possible to run the hazard from the command line and then the risk from the WebUI and vv. This will be an issue also for exporting data: calculations done with the command line cannot be exported by the webui (if the home read permissions are owner-only) and vv (if the user is not the webserver's owner group). At this point probably we'll need to hide calculations done with different techniques (loosing very useful feature).

This has been marked as critical since the WebUI on Wilson (run by the 'openquake' user) will be used next week in the Risk workshop to run hazard + risk.

daniviga:
The issue is slight different as supposed: the problem is not related with the ownership of the datastore (unless cli and webui are mixed), but on the gunicorn multi-processing.

If the WebUI is run with a single gunicorn process everything is working as expected. If more than one worker is used the risk computation fails with

```python
Traceback (most recent call last):
  File "/usr/local/openquake/oq-engine/openquake/engine/engine.py", line 218, in run_calc
    _do_run_calc(calculator, exports)
  File "/usr/local/openquake/oq-engine/openquake/engine/engine.py", line 254, in _do_run_calc
    calc.pre_execute()
  File "/usr/local/openquake/oq-risklib/openquake/calculators/scenario_risk.py", line 105, in pre_execute
    base.RiskCalculator.pre_execute(self)
  File "/usr/local/openquake/oq-risklib/openquake/calculators/base.py", line 451, in pre_execute
    HazardCalculator.pre_execute(self)
  File "/usr/local/openquake/oq-risklib/openquake/calculators/base.py", line 272, in pre_execute
    self.read_exposure_sitecol()
  File "/usr/local/openquake/oq-risklib/openquake/calculators/base.py", line 290, in read_exposure_sitecol
    self.gmfs = get_gmfs(self)
[...]

PR #1872, opened 2015-10-19
Now the engine server closes the datastore

micheles:
See https://github.com/gem/oq-engine/issues/1870. The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1479
Companion of https://github.com/gem/oq-risklib/pull/448


daniviga:
Since the errors in CI are related to an another issue, LGTM

PR #1876, opened 2015-10-20
The ProbabilisticEventBased calculator is no more special

micheles:
Now the file is called job.ini. The rule to determine if a job is of kind hazard or risk has changed and it is now based (more correctly) on the calculation_mode. This is needed to run single file risk jobs on the Web UI. This requires the companion https://github.com/gem/oq-risklib/pull/453.

The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1484/

### DIFF
--- a/openquake/engine/db/models.py
+++ b/openquake/engine/db/models.py
@@ -300,8 +300,10 @@ class OqJob(djm.Model):
         """
         'hazard' or 'risk'
         """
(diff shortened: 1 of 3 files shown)


daniviga:
LGTM

daniviga:
LGTM if tests are green

PR #1877, opened 2015-10-20
webui: change the run calculation button label

daniviga:
Now it's possible to run a single calculation that runs both hazard and risk at the same time, thus the old "Run a Hazard Calculation" button label is now misleading.


micheles:
LGTM

PR #1883, opened 2015-10-24
Removed the epsilon_sampling parameter

micheles:
This parameter was used by the old calculators, but they are no more. Now it is a parameter of the job.ini files, with default 1000. The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/1496/

PS: I am also removing a lot of code which was of support to the old calculators.


nastasi-oq:
LGTM

PR #1887, opened 2015-10-26
Extended del_calc to remove the .hdf5 file

micheles:
Also, I am removing the `exception.message` attribute, which is deprecated. The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1499


openquake/engine/bin/openquake_cli.py
```diff
@@ -405,7 +405,7 @@ def export(output_id, target_dir, export_type):
         print 'File Exported:'
         print the_file
     except NotImplementedError, err:
```
daniviga: Please use `as` instead of the comma

micheles: Done

daniviga:
LGTM

Issue #1826, opened 2015-09-14
Improve the engine server UI for datastore outputs


micheles:
For instance, there should not be a "view" button if the output is a zip file.

micheles:
I propose to remove the "view" button altogether. After all, it is of very limited utility. Instead it would be useful to implement a plotter for geospatial outputs, but this is for another ticket.

PR #1888, opened 2015-10-28
Remove the view button

micheles:
A proposal to close https://github.com/gem/oq-engine/issues/1826 with a minimal effort so that we do not block the release 1.6. The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/1501/


daniviga:
I would keep the "view" mode in the API (and just remove the button), having the "download" mode as default: if you want the stream you pass a parameter. This view/stream mode in the future it could be useful to ingest data in GeoServer.

daniviga:
LGTM

PR #1889, opened 2015-10-28
Do not distinguish hazard and risk calculations in the performance report

micheles:
Since it is misleading with the new calculators. Here are the reports produced by https://ci.openquake.org/job/zdevel_oq-engine/1502: http://artifacts.openquake.org/demos/oq-engine/zdevel/precise/1502/jobs-2015-10-28.html#tabs-15


daniviga:
LGTM

PR #1891, opened 2015-10-31
Ignore headers in the classical_risk tests

micheles:
The header of some expected files has changed in oq-lite, so the engine tests should not be sensitive to it. Fixes: https://ci.openquake.org/job/master_oq-engine/2075/console


micheles:
LGTM

PR #1898, opened 2015-11-08
Fixed the monitoring of children

micheles:
Just reusing the parent `PerformanceMonitor.__call__` takes care of the children properly. You can see the difference by looking at http://artifacts.openquake.org/demos/oq-engine/master/precise/2089/jobs-2015-11-07.html#tabs-15 (without monitoring of the children, "getting hazard" and "computing individual risk" were missing) contrasted with http://artifacts.openquake.org/demos/oq-engine/zdevel/precise/1512/jobs-2015-11-08.html#tabs-15 (with monitoring).


daniviga:
LGTM

PR #1901, opened 2015-11-15
Engine 1.6 release notes

micheles:
Here they are. They are best read by looking at https://github.com/gem/oq-engine/blob/engine-1.6/doc/engine/What's-new.md


doc/engine/What's-new.md
```diff
@@ -1,190 +1,235 @@
...
+will get a clear error message. Moreover, each risk model has a
+`lossCategory` attribute which must be set consistently with the name
+of the key in the job.ini file (see the [manual]
+(http://www.globalquakemodel.org/openquake/support/documentation/engine/)
+for the details).
+
+6. Some work has been going on hazardlib, as usual, and you can
+have a look at the [changelog](https://github.com/gem/oq-hazardlib/blob/engine-1.6/debian/changelog). The most prominent feature is the introduction of
+epistemic uncertainties on the fault geometry into the Source Model Logic Tree.
+
```
g-weatherill: I think we can add more detail here given the size of the change and the significant increase in flexibity to which the users are now exposed. I suggest something like:

6. Addition of support for new epistemic uncertaities in the source model logic tree. Users are now able to represent alternative geometries for the fault source typologies using the following uncertainty model definitions:

i. `simpleFaultGeometryAbsolute` - Replaces the simple fault geometry with the specified simple fault surface.
ii. `complexFaultGeometryAbsolute` - Replaces the complex fault geometry with the specified complex fault surface.
iii. `characteristicFaultGeometryAbsolute` - Replaces the characteristic fault geometry with the specified surface.
iv. `simpleFaultDipRelative` - Increases or decreases the fault dip with the specified value.
v. `simpleFaultDipAbsolute` - Replaces the simple fault dip with the specified value.

Also added is the ability to replace a given evenly discretised magnitude frequency distribution of a source in the logic tree using the `incrementalMFDAbsolute` option.

Documentation on the usage of these features will be added in the forthcoming version of the manual.

doc/engine/What's-new.md
```diff
@@ -1,190 +1,235 @@
...
+7. We added an optional attribute `discretization` to the
+area source geometry XML description: this means that it is possible to
+specify a source-specific discretization step. This is useful in site specific
+analysis: area sources with little impact on the site of interest can use
+a large discretization step whereas the important area sources can use a
+finer discretization step.
+
+10. Two new GMPEs have been added to hazardlib: *Montalva et al. (2015)*
+and *Tusa and Langer (2015)*.
```
g-weatherill: If this PR (https://github.com/gem/oq-hazardlib/pull/394) gets merged in time then you can add *Allen, Wald and Worden (2012) (Intensity Prediction Equation)* to that list

doc/engine/What's-new.md
```diff
@@ -1,190 +1,252 @@
...
-We have detailed instructions for installing the engine on CentOS 7
-and Fedora and in general on [Red Hat Enterprise Linux clones]
+OpenQuake 1.6 supports Ubuntu from release 12.04 up to 15.10. 
+We provide official packages for the long term releases 12.04 and 14.04.
+We were able to extend the support to
+Ubuntu 12.04 by backporting the package `python-h5py` from Ubuntu 14.04.
+So *Ubuntu 12.04 is still supported, even if it is deprecated*.
+
+We have official packages also for CentOS 7
+and Fedora and in general for [Red Hat Enterprise Linux clones].
```
daniviga: No official packages for Fedora. Our CentOS package will work on Fedora 19-22 but not on Fedora 22 (Django 1.8 issues)

daniviga:
It's a go for me.

PR #1903, opened 2015-11-16
WebUI doc

### LINKED ISSUE #1878: Update the documentation and the manual with new demos
The documentation and the manual (@raoanirudh) must be updated with the news demos and their new path (hazard and risk dirs).

We should also check that the documentation is consistent with the current path of the demos in both Ubuntu and RedHat.

Instruction about migration from NRMl 0.4 to 0.5 are missing.

daniviga:
Not perfect, but enough for 1.6.0

Final result: https://github.com/gem/oq-engine/blob/webui-doc/doc/server/OpenQuake-engine-webui.md
Closes #1878


micheles:
I would change "It can be possible to watch the log" -> "It is possible to watch the log"
"Output formats (xml, csv, geojson) availability depends on the type of artifacts generated by the calculation." -> " Output availability depends on the type of artifacts generated by the calculation. The possible formats are  xml, csv, geojson, zip"

I would add a small note saying that the "Remove calculation" button actually hides the computation from the Web UI, but does not remove it from the database nor from the datastore.

doc/server/OpenQuake-engine-webui.md
```diff
@@ -1,33 +1,84 @@
...
+![WebUI remove](../img/WebUI-remove.gif)
+
+### Calculation status
+
+A calculation can be in the following statuses `pre_executing`, `executing`, `complete`, `failed`
+
+![WebUI calculations status](../img/WebUI-status.png)
+
+The reason behind a `failed` calculation can be inspected using the `Traceback` button.
+
```
micheles: "by using"

doc/server/OpenQuake-engine-webui.md
```diff
@@ -1,33 +1,84 @@
...
+### Remove a calculation
+
+To remove a calculation from the list press the `Remove` button
+
+![WebUI remove](../img/WebUI-remove.gif)
+
+### Calculation status
+
+A calculation can be in the following statuses `pre_executing`, `executing`, `complete`, `failed`
+
```
micheles: Please add a colon after statuses:

doc/server/OpenQuake-engine-webui.md
```diff
@@ -1,33 +1,84 @@
...
 
-### Add a new local superuser
+#### Add a new local superuser
 ```bash
 DJANGO_SETTINGS_MODULE="openquake.server.settings" python manage.py createsuperuser --database=auth_db
 ```
 
-### PAM
+#### PAM
 Process owner must be member of the `shadow` group.
```
micheles: I would one line explain what PAM is.

micheles:
LGTM

PR #1907, opened 2015-12-01
Open the DataStore in read mode when exporting

micheles:
This should solve the problem that it is impossible to export the results of a calculation ran by another user on Wilson.


daniviga:
LGTM

PR #1909, opened 2015-12-06
doc: make celeryconfig.py path more explicit

daniviga:
It seems that users do not read text that is not in command blocks, so moving the `cd` to  `celeryconfig.py` on the same line as the `celery` command.


micheles:
LGTM

PR #1912, opened 2015-12-15
Removed two obsolete tests

micheles:
We forgot to remove them since they are of kind 'slow' and zdevel did not run them, so master is broken: https://ci.openquake.org/job/master_oq-engine/2153/


ptormene:
LGTM

PR #1913, opened 2015-12-16
Fixed the ordering of the assets in the loss map XML exporter

micheles:
In this way the exported XML files are easier to compare with the ones produced by oq-lite.

NB: there is the same problem also in the loss curve exporter.


raoanirudh:
LGTM

PR #1917, opened 2015-12-18
Enabled Zip64 extensions

micheles:
Let's do that experimentally. We can always go back if there are issues and the unzipped files are always available. Now at least the users of 64 bit machines will be happy.


daniviga:
LGTM

PR #1918, opened 2015-12-18
Reflected change in oq-risklib

micheles:
Needed for https://github.com/gem/oq-risklib/issues/597, companion of https://github.com/gem/oq-risklib/issues/598


daniviga:
LGTM

PR #1929, opened 2016-01-22
Moved openquake/engine/db -> openquake/server/db

micheles:
Preliminary work to remove the dependency from Django from the engine and move it purely to the engine server. The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/1584/

PS: I have also removed some obsolete modules and scripts.


daniviga:
LGTM

PR #1930, opened 2016-01-22
Use a celery-aware default for concurrent_tasks

micheles:
This feature has been requested by Daniele for years: the default number of concurrent_tasks should be decided dynamically by the application, by asking celery how many cores are available, and by multiplying it by two, thus following an ancient tradition of the engine.
Notice that:

1. is `use_celery` is false in `openquake.cfg` this feature is not used
2. if the user passes an explicit `concurrent_tasks` in the `job.ini` file the default is set but not used

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,6 @@
   [Michele Simionato]
+  * If `use_celery` is true, use celery to determine a good default for
+    the parameter `concurrent_tasks`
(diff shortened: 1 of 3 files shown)


daniviga:
Thank you! LGTM!

daniviga:
Thank you! LGTM!

PR #1932, opened 2016-01-23
Removed dependency on django.contrib.gis

micheles:
We do not have tables using geospatial extensions anymore, so the Django models do not need to depend on django.contrib.gis.

Everything is green: https://ci.openquake.org/job/zdevel_oq-engine/1587


daniviga:
I now we are going to remove Postgres entirely, but probably we can already remove PostGIS as a dependency since django.gis gone away

micheles:
I don't want to change the installation/upgradedb procedure that requires postgis. I prefer to wait and to drop everything at once.

daniviga:
LGTM

PR #1933, opened 2016-01-25
Removed obsolete code

micheles:
I am removing some obsolete oq-engine switches and celery-related code in the packager.

Everything is green: https://ci.openquake.org/job/zdevel_oq-engine/1589/

### DIFF
--- a/packager.sh
+++ b/packager.sh
@@ -91,7 +91,6 @@ sig_hand () {
     if [ "$lxc_name" != "" ]; then
         set +e
         scp "${lxc_ip}:/var/tmp/openquake-db-installation" "out_${BUILD_UBUVER}/openquake-db-installation"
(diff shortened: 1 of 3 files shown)


packager.sh
```diff
@@ -548,13 +509,6 @@ _pkgtest_innervm_run () {
     # XXX: should the --upgrade-db command go in the postint script?
     ssh $lxc_ip "set -e; oq-engine --upgrade-db --yes"
 
-    # run celeryd daemon
-    ssh $lxc_ip "cd /usr/share/openquake/engine ; celeryd >/tmp/celeryd.log 2>&1 3>&1 &"
```
daniviga: I would keep this part as option, controlled by an ENV var so celery can be used, on demand, to run the demos.

A sed to change the openquake.cfg must be add to. I will do.

daniviga:
LGTM

PR #1942, opened 2016-02-09
Fixed some command-line switches and removed others

micheles:
After discussion with Anirudh and inspection of the current switches. See https://ci.openquake.org/job/zdevel_oq-engine/1606/

Closes https://github.com/gem/oq-engine/issues/1943


raoanirudh:
LGTM

PR #1951, opened 2016-02-17
Updated the Whats new page

kipkwony:


micheles:
This should not go in master. The documentation for the engine-1.8 and its fixes must stay in the branch 1.8.

PR #1954, opened 2016-02-17
Avoided calling get_oqparam

micheles:
This should be fixing https://ci.openquake.org/job/master_oq-engine/2254/
The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/1625/


daniviga:
LGTM

PR #1962, opened 2016-02-24
Fixed the ordering of the logs in the Web UI

micheles:
I forgot to set the ordering at the Django level in the migration to the new tables.


daniviga:
LGTM

PR #1963, opened 2016-02-24
Changed the default job status to 'executing'

micheles:
So that it looks better in the Web UI (now it is 'pre_executing' which is unclear).


daniviga:
LGTM

PR #1964, opened 2016-02-26
Fixed the double log bug

micheles:
You can see the bug by running

```bash
$ oq-engine --run demos/risk/ClassicalDamage/job_hazard.ini 
```

in the current master. The lines of log are duplicated. This PR fixes the issue, that happened because an additional stream handler was automatically generated by the line `readinput.get_oqparam(params)`
that needed to log the fact that in this demo the export_dir is missing.

Tests and demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/1637/

PS: while at that, I added storing the total runtime of the calculation in the performance table.

### DIFF
--- a/openquake/engine/engine.py
+++ b/openquake/engine/engine.py
@@ -390,14 +390,19 @@ def job_from_file(cfg_file, username, log_level='info', exports='',
     # read calculation params and create the calculation profile
     params = readinput.get_params([cfg_file])
     params.update(extras)
-    # build and validate an OqParam object
-    oq = readinput.get_oqparam(params)
-    # create the current job
+
+    class oq:  # fake OqParam object
+        calculation_mode = params['calculation_mode']
+        description = params['description']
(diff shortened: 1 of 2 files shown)


openquake/engine/engine.py
```diff
@@ -390,14 +391,20 @@ def job_from_file(cfg_file, username, log_level='info', exports='',
     # read calculation params and create the calculation profile
     params = readinput.get_params([cfg_file])
     params.update(extras)
-    # build and validate an OqParam object
-    oq = readinput.get_oqparam(params)
-    # create the current job
+
+    class oq:  # fake OqParam object
```
daniviga: There isn't any other way to get only these params instead of mocking the class?

micheles: Yes, I can just instantiate an OqParam object. Done

daniviga:
LGTM

Issue #1928, opened 2016-01-21
Remove PostgreSQL from the engine


micheles:
It should be replaced with SQLite. Probably it is best to remove the table job_stats and merge it with the table oq_job. Also the log table can probably be refactored. Moreover we can merge the authentication database in the main database. Finally we should merge `openquake.engine.settings` inside `openquake.server.settings`.

PR #1966, opened 2016-03-04
Remove Django routes and use a single database

micheles:
Another step towards removing postgres https://github.com/gem/oq-engine/issues/1928.
Django routes are just not needed when going in the direction of a single sqlite database which has no users. Also, there is no point in having a separated authentication database.
The tests run: https://ci.openquake.org/job/zdevel_oq-engine/1647

TODO: we should manually check that the authentication works when LOCKDOWN=True.


daniviga:
LGTM. It works well also with `LOCKDOWN = True`

PR #1967, opened 2016-03-07
Fixed bug in the Web UI when running a risk calculation starting from a hazard calculation

micheles:
This was discovered by Vitor and is critical.


daniviga:
Please open it against branch engine-1.9

PR #1970, opened 2016-03-07
Fixed the job description in the Web UI

micheles:
Before it was temporarily set to "A job".
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1656


daniviga:
LGTM

Issue #1944, opened 2016-02-10
Make Celery an optional dependency


daniviga:
Celery workers are not needed anymore when `use_celery = False` but it's still required as python module since the code contains several celery imports.

We should also check the same for other "multi-node only" libraries (i.e. Django).

Probably we should start by removing the CeleryNodeMonitor (it was introduced for noncluster users, and they are not using celery anymore).

PR #1971, opened 2016-03-08
Make the celery dependence optional

micheles:
Closes https://github.com/gem/oq-engine/issues/1944. Tests and demos are green: https://ci.openquake.org/job/zdevel_oq-engine/1660/

I have manually uninstalled celery on my machine and everything works.


openquake/engine/celery_node_monitor.py
```diff
@@ -1,158 +0,0 @@
...
-        Register the signal handlers for SIGTERM and SIGABRT
-        if they were not registered before.
-        """
-        if not cls.registered_handlers:  # called only once
-            signal.signal(signal.SIGTERM, cls.handle_signal)
-            signal.signal(signal.SIGABRT, cls.handle_signal)
-            cls.registered_handlers = True
-
-
-class CeleryNodeMonitor(object):
```
daniviga: An issue removing the celery monitor is that you don't get any feedback anymore if you start a calculation and celery is not started/not running. It seems that the calculation it's processing data in the pre execute phase, instead it's just waiting for a non existent celery consumer.

micheles: You are right, I will restore the check.

daniviga:
LGTM

PR #1972, opened 2016-03-09
Reflected changes in the Monitor

micheles:
`PerformanceMonitor` has been renamed to `Monitor` and the function `check_mem_usage` now depends on the monitor to send back warnings as desired in https://github.com/gem/oq-risklib/issues/732. This the companion of https://github.com/gem/oq-risklib/pull/733.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1663


daniviga:
LGTM

PR #1973, opened 2016-03-09
Removed openquake.engine.settings

micheles:
Finally removed a long standing duplication and confusion. Now we have only `openquake.server.settings`. Tests and demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/1665/


daniviga:
LGTM

PR #1974, opened 2016-03-09
deb: always replace the old celeryconfig.py

daniviga:
Since the user should not change the `celeryconfig.py` is fine if the binary package always replaces it. This will simplify the upgrade and will avoid strange errors when using celery with an old config file.

Tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1666/


micheles:
LGTM

PR #1976, opened 2016-03-10
deb: make the postinst working on usupported distros

daniviga:
This PR makes the `postinst` (and the `.deb` package) working for Ubuntu derivates like Mint as an unsupported system.

A warning is printed is the OS is unsupported.

Test for both 12.04 and 14.04 are running: https://ci.openquake.org/job/zdevel_oq-engine/1670/console
Manual testing on Mint 17 passed.


nastasi-oq:
LGTM

PR #1980, opened 2016-03-11
deb: use the postgresql meta package

daniviga:
While we are awaiting for the postgresql removal, let's use the meta package instead of a specific version. We are not even using postgis anymore, just base functions of postgresql. This makes the Trusty deb packages working also on Debian Jessie and Ubuntu 16.04 (even if we do not officially support this).

Tests: https://ci.openquake.org/job/zdevel_oq-engine/1683/


nastasi-oq:
LGTM

PR #1978, opened 2016-03-11
Confine all database interaction in the module openquake.server.db.actions

micheles:
This is a refactoring in preparation for https://github.com/gem/oq-engine/issues/1975. By moving all the database-related interaction in a dedicated module I am effectively decoupling the database from the application. I am also introducing a `dbcmd` object mediating the interaction. In a future pull request `dbcmd` will be a client sending commands to the database server.
The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/1678

PR #1981, opened 2016-03-12
Fix output exports (--eos)

daniviga:
With https://github.com/gem/oq-engine/pull/1978 we introduced a bug and now isn't possible to export outputs with the command `--eos ###`. The error is:

```python
  File "/usr/local/openquake/oq-engine/bin/oq-engine", line 359, in <module>
    main()
  File "/usr/local/openquake/oq-engine/bin/oq-engine", line 348, in main
    hc_id = dbcmd('get_hc_id', args.list_outputs)
  File "/usr/local/openquake/oq-engine/openquake/engine/logs.py", line 55, in dbcmd
    return getattr(actions, action)(*args)
  File "/usr/local/openquake/oq-engine/openquake/server/db/actions.py", line 91, in get_hc_id
    hc_id = int(hc_id)
TypeError: int() argument must be a string or a number, not 'NoneType'
```

This PR fixes the issue (I don't know if it's the best way, but it works) and also adds a test in `packager.sh` that tries to export outputs from calculation 1. The build was green because as now we do the export together with the calculation, not afterwards.

Thanks to @vup1120 for reporting the bug.
Tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1686/

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,6 @@
+  [Daniele Viganò, Michele Simionato]
+  * Fixed a bug in `oq-engine --export-outputs`
+
   [Daniele Viganò, Matteo Nastasi]
   * Allow installation of the binary package on Ubuntu derivatives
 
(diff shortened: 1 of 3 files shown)


micheles:
LGTM

PR #1986, opened 2016-03-17
Made more robust the killing of processes by patching concurrent.futures

micheles:
We are already using this approach in oq-lite, let's use it even in the engine.
The demos are running  here: https://ci.openquake.org/job/zdevel_oq-engine/1699


daniviga:
LGTM

Issue #1987, opened 2016-03-22
celery must be used when use_celery is true


micheles:
Unfortunately, this is not necessarily the case, because the support for celery is implemented in terms of a HORRIBLE HACK, i.e. the decorator `parallel.litetask` is monkey-patched with the `oqtask` decorator at runtime: depending on the order of the imports, the decorator can be patched too late. For instance, an innocuous refactoring like the one in https://github.com/gem/oq-engine/pull/1978, making the engine.logs module dependent indirectly on commonlib.parallel, has the net effect that celery is no more used, because the task functions are imported before the patch. The solution of course is to remove the horrible hack. This is now possible. Also, a test of celery usage must be added.

PR #1988, opened 2016-03-22
Celery not used bug

micheles:
Solves https://github.com/gem/oq-engine/issues/1987. A second set of pull requests will solve the issue in a robust way and will add a regression test. You can check that nothing is broken: https://ci.openquake.org/job/zdevel_oq-engine/1716


daniviga:
LGTM

PR #1990, opened 2016-03-22
Refactored the concurrency mechanisms management in the engine

micheles:
Closes https://github.com/gem/oq-engine/issues/1987 in a good enough way. There is still a tricky dependence on when the imports are done, but at least now if the import is done at the wrong moment the calculation fails when an error, which is better than running to completion by using the wrong distribution mechanism. This requires the companion PR https://github.com/gem/oq-risklib/pull/751.

The tests are green: https://ci.openquake.org/job/zdevel_oq-engine-celery/16


daniviga:
LGTM

Issue #1989, opened 2016-03-22
Another double log


micheles:
There is yet another double log when running classical/case_3/job.ini with the engine. We must check the handlers.

PR #1996, opened 2016-03-25
Fixed in a more robust way the duplicated log bug

micheles:
Closes https://github.com/gem/oq-engine/issues/1989. The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1727


daniviga:
LGTM

PR #2002, opened 2016-04-19
packager.sh: enhanced deps_list function

nastasi-oq:
enhanced deps_list function to manage dependencies with specific version constraints


daniviga:
LGTM

PR #2013, opened 2016-04-29
Remove the openquake_worker.cfg

### LINKED ISSUE #2010: Remove the openquake_worker.cfg
Is now useless. Just put the master server IP in the host and it will work for all the nodes even in an exported NFS configuration.

daniviga:
Closes #2010 

Packaging tests: https://ci.openquake.org/job/zdevel_oq-engine/1775/


debian/changelog
```diff
@@ -1,3 +1,6 @@
+  [Daniele Viganò]
+  * Removed the openquake_worker.cfg file because isn't used anymore
```
micheles: *isn't* -> *it is not used*

micheles:
LGTM

PR #2014, opened 2016-04-29
Implemented single-user mode

micheles:
If the flag `multi_user` is set to False, the engine automatically starts the DbServer using a database in ~/db.sqlite3. This is very useful for somebody running the engine from the sources (I am thinking of all of our scientists using Mac OS). The tricky bit is: should the default for `multi_user` be True or False? For a source installation on a laptop it should be False, for an installation from packages it should be True.

The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1782/


openquake/engine/bin/openquake_cli.py
```diff
@@ -290,9 +290,14 @@ def main():
...
     if err:
-        sys.exit('Please start the DbServer: '
-                 'see the documentation for details')
-
+        multi_user = valid.boolean(config.get('dbserver', 'multi_user'))
+        if multi_user:
+            sys.exit('Please start the DbServer: '
+                     'see the documentation for details')
+        # otherwise start the DbServer automatically
+        dblog = os.path.expanduser('~/oq-dbserver.log')
```
daniviga: Please move this setting to the configuration file

daniviga:
LGTM

Issue #2008, opened 2016-04-26
Add real export tests for --eos, --eo and for --exports


micheles:
By specifying --exports or not. There should be a default export_type.

micheles:
Such commands are called in the packager. This is good enough for the moment.

PR #2015, opened 2016-04-29
Fixed a small export bug

micheles:
The problem is that an output can be exported in more than one format it is now exported in all formats. We must revert to the previous situation, when only the format with highest priority is used. 
The issue https://github.com/gem/oq-engine/issues/2008 is more and more relevant.
The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/1777


daniviga:
LGTM

Issue #2004, opened 2016-04-20
webui: remove the option "Download txt" from gmfs outputs


CatalinaYepes:
The button for `Download txt` in the gmfs outputs, appearing after running a scenario calculation, should be removed.

The `xml` and `csv` options are sufficient.

![screen shot 2016-04-20 at 13 54 31](https://cloud.githubusercontent.com/assets/6975363/14673766/277769cc-0700-11e6-82ff-4bd9887a6efe.png)

Issue #2018, opened 2016-05-04
WebUI exports are broken


daniviga:
Trying to download any of the exports generated by the WebUI returns an error:

```python
Environment:


Request Method: GET
Request URL: http://192.168.124.38:8000/v1/calc/4/result/4?export_type=csv

Django Version: 1.6.1
Python Version: 2.7.6
Installed Applications:
('openquake.server.db', 'django.contrib.staticfiles', 'openquake.server')
Installed Middleware:
('django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware')


Traceback:
File "/usr/lib/python2.7/dist-packages/django/core/handlers/base.py" in get_response
  112.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/lib/python2.7/dist-packages/openquake/server/views.py" in wrap
  76.             response = func(request, *args, **kwargs)
File "/usr/lib/python2.7/dist-packages/django/views/decorators/http.py" in inner
  41.             return func(request, *args, **kwargs)
File "/usr/lib/python2.7/dist-packages/openquake/server/views.py" in get_result
  472.         exported = core.export(result_id, tmpdir, export_type=export_type)
File "/usr/lib/python2.7/dist-packages/openquake/baselib/general.py" in __call__
  457.         return self[key](obj, *args, **kw)

Exception Type: TypeError at /v1/calc/4/result/4
Exception Value: dispatch_on_colon() got an unexpected keyword argument 'export_type'
```

PR #2020, opened 2016-05-04
Webui exports

micheles:
Fixes https://github.com/gem/oq-engine/issues/2004 and https://github.com/gem/oq-engine/issues/2018. A test must be added.


daniviga:
LGTM

PR #2029, opened 2016-05-16
The export must be run by the current user

micheles:
Before, it was run by the dbserver, i.e. by the openquake user and there were errors when exporting, like the `.` directory being expanded to the current directory for the dbserver, not for the user.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1824


daniviga:
LGTM

PR #2034, opened 2016-05-19
webui: if no path is requested, go to /engine

daniviga:
Avoid the `Page not found` if any sub-url isn't requested (i.e. `http://localhost:8000`). If no sub-url is requested the public frontend (`/engine`) is provided via a redirect.

This was already done on production systems where nginx is used. Now the same behavior is applied using the embedded django server.


micheles:
This is useful, LGTM.

PR #2036, opened 2016-05-20
webui: open files in binary mode

daniviga:
To avoid corruption of zip files when running the WebUI on Windows. This does not affect Linux.

See https://docs.python.org/2/library/functions.html#open
@ptormene had a similar issues with the QGIS plugin...


pslh:
Tested on Windows 10, all working correctly (previously ZIPs were corrupt).

LGTM.

daniviga:
Tested also on OATS (Linux) and works as expected.

PR #2045, opened 2016-05-26
Support django 1.5

micheles:
This is needed to support the platform (our version of geonode uses Django 1.5.5). See the runs for Ubuntu 12.04 and 14.04: https://ci.openquake.org/job/zdevel_oq-engine/1854

When using Django 1.3.1 the dbserver and the engine work, but the WebUI does not (this is why its tests have been skipped): it is the user responsability to upgrade to a newer version of Django. Django 1.5 is enough to run the WebUI, 1.4 has not been tried.


nastasi-oq:
LGTM

PR #2046, opened 2016-05-27
Fixed the dependency tower of the OpenQuake libraries

micheles:
This is the dependency tower:
```
level 8            commands
level 7            server
level 6            engine
level 5            calculators
level 4            commonlib
level 3            risklib
level 2            hazardlib
level 1            baselib
```
I have moved around modules and introduced tests to make sure that the dependencies will not get mixed up by future refactorings, i.e. the lower levels cannot depend from the higher levels. I have also regenerated the API documentation.

The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/1858/


daniviga:
LGTM

Issue #2025, opened 2016-05-09
Make the dbserver log path configurable


daniviga:
... or at least move it in the `oqdata` folder. See https://github.com/gem/oq-engine/pull/2014#discussion_r62024506

In the future it would be great to have the dbserver able to write a log file (without having to manually redirect stderr/out). The log file should be set or via command line or via a config in the `openquake.cfg`

PR #2047, opened 2016-05-27
Added a default location for the dbserver log

micheles:
Closes https://github.com/gem/oq-engine/issues/2025. The demos are running: https://ci.openquake.org/job/zdevel_oq-engine/1893


openquake.cfg
```diff
@@ -41,7 +41,8 @@ celery_queue = celery
 [dbserver]
 # enable multi_user if you have a multiple user installation
 multi_user = false
-file = ~/db.sqlite3
+file = ~/oqdata/db.sqlite3
+log = ~/oqdata/dbserver.log
```
daniviga: Moving these files into `oqdata` will led to an error if oqdata does not exist yet:

```IOError: [Errno 2] No such file or directory: '/home/dvigano/oqdata/dbserver.log'```

daniviga:
This https://github.com/gem/oq-engine/blob/dbserverlog/packager.sh#L564 must be updated too.

openquake.cfg
micheles: Now both oq-engine and upgrade_manager are creating the datastore.DATADIR if not already there; the tests were broken because I was not passing the log file to the dbserver.

daniviga:
LGTM

PR #2056, opened 2016-05-31
Improved the parsing of logic tree files

micheles:
I have removed the attribute gsim_lt from SourceModel instances. It means that the gsim_logic_tree file is now parse only once and not once per source model.

The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1881


ptormene:
LGTM

PR #2057, opened 2016-05-31
Early check on the valid periods for spectral accelerations

micheles:
Closes https://github.com/gem/oq-risklib/issues/636. The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1890/


openquake/commonlib/logictree.py
```diff
@@ -1285,6 +1287,24 @@ def __init__(self, fname, tectonic_region_types=['*'], ltnode=None):
...
+                for attr in dir(gsim):
+                    coeffs = getattr(gsim, attr)
+                    if not isinstance(coeffs, CoeffsTable):
+                        continue
+                    for imt in imts:
+                        if imt.startswith('SA'):
+                            try:
+                                coeffs[from_string(imt)]
+                            except KeyError:
+                                raise ValueError('%s is not defined for %s' %
```
g-weatherill: It could be more useful to the user for the message to read "%s is outside of the period range supported by %s", as it tells the user explicitly why it was not possible to run the GMPE in this case

g-weatherill:
LGTM

PR #2061, opened 2016-06-06
Update oq_*_db scripts

daniviga:
In `oq_create_db` and `oq_reset_db` take the DB path from the `openquake.cfg` (looking first in `/etc` and then in the relative path to the scripts based on the repo structure) and do not guess that by default the location of the db is always the openquake user's home.

This makes the packages work also on "strange" setup like the VM.

Tests: https://ci.openquake.org/job/zdevel_oq-engine/1902/

### DIFF
--- a/openquake/engine/bin/oq_create_db
+++ b/openquake/engine/bin/oq_create_db
@@ -27,11 +27,11 @@ for c in ${OQ_CFG_FILES[@]}; do
     fi
 done
 
-# Fallback: create the database in the user's home folder
-if [ -z "$OQ_DB" ]; then
-    OQ_DB=$(eval echo "~${OQ_USER}/oqdata/db.sqlite3")
-else
+if [ -n "$OQ_DB" ]; then
     OQ_DB=${OQ_DB#*=}
+else
+    # Fallback: create the database in the user's home folder
+    OQ_DB=$(eval echo "~${OQ_USER}/oqdata/db.sqlite3")
 fi
 
 while (( "$#" )); do
--- a/openquake/engine/bin/oq_create_db
+++ b/openquake/engine/bin/oq_create_db
@@ -7,14 +7,27 @@ if [ $GEM_SET_DEBUG ]; then
 fi
 set -e
 
-OQ_USER='openquake'
-OQ_DB_SCRIPT='-m openquake.server.db.upgrade_manager'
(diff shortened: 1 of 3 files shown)


micheles:
This logic could be implemented in Python (the ConfigParser module already works this way). However I understand there is some urgency and it is not so bad as it is, so LGTM.

daniviga:
So probably, just running `oq-engine --upgrade-db` is enough and it will take the right conf

micheles:
Yes, it should do the right thing from oq-engine. The problem is when upgrade_manager.py is called directly, then the configuration file will not be read.

daniviga:
See https://github.com/gem/oq-engine/issues/2065

daniviga:
For this purpose I don't think upgrade_manager.py should be called manually:
- On packages/multi user installation it's fine to run --upgrade-db as soon as the DbServer is started first
- On single user/development we already use --upgrade-db since it starts the dbserver if it is not running

There's any other case?

micheles:
There is a direct call in packager.sh:

```564 :         sudo -u openquake python -m openquake.server.db.upgrade_manager ~openquake/db.sqlite3
```

openquake/engine/bin/oq_create_db
```diff
@@ -7,15 +7,33 @@ if [ $GEM_SET_DEBUG ]; then
...
     echo "Syntax: $(basename "$0") [-u user] [-s /custom/python/script.py] [-d /custom/db/path/to/db.sqlite3]"
     exit 0
 }
 
+OQ_BASE=$(dirname $(readlink -e "$0"))
+OQ_USER='openquake'
+OQ_DB_SCRIPT='-m openquake.server.db.upgrade_manager'
+OQ_CFG_FILES=('/etc/openquake/openquake.cfg' '../etc/openquake.cfg' '../openquake.cfg' '../../../openquake.cfg')
+
+for c in ${OQ_CFG_FILES[@]}; do
```
nastasi-oq: List instead of simple space separated list for a particular reason ?

daniviga:
Yes, in the case of the packager we know it will use the default configuration which is the openquake's home

openquake/engine/bin/oq_create_db
daniviga: This code will disappear soon, see comments above

nastasi-oq:
LGTM

PR #2064, opened 2016-06-07
Fixed the expected files for the slow tests

micheles:
The precision has been lowered to 5 decimal digits and the format has changed. The change broke master (https://ci.openquake.org/job/master_oq-engine/2471/) because the slow tests are not run on zdevel.


micheles:
LGTM

PR #2067, opened 2016-06-07
Added a CurveReader for hazard curves in CSV format

micheles:
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1905/
This is the companion of https://github.com/gem/oq-hazardlib/pull/463.

Here is an example of how to plot a set of curves:
```python
import matplotlib.pyplot as plt
from openquake.commonlib.readers import CurveReader

def plot_curve(fname, number=0):
    creader = CurveReader(fname)
    fig = plt.figure()
    ax = fig.add_subplot(111)
    ax.grid(True)
    curve = creader.read()[number]
    for imt, imls in creader.imtls.items():
        ax.plot(imls, curve[imt], label=imt)
    plt.legend()
    plt.show()

if __name__ == '__main__':
    import sys
    plot_curve(sys.argv[1])
```


openquake/commonlib/readers.py
```diff
@@ -0,0 +1,62 @@
...
+        with open(fname) as f:
+            header = next(f).split(',')
+        self.imtls, self.fields = self._parse_header(header)
+
+    def _parse_header(self, header):
+        fields = []
+        cols = []
+        for col in header:
+            if '-' in col:
+                cols.append(col.split('-', 1))
```
ptormene: You might add a couple of examples of possible formats of `col`

openquake/commonlib/readers.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    def __init__(self, fname):
+        self.fname = fname
+        with open(fname) as f:
+            header = next(f).split(',')
+        self.imtls, self.fields = self._parse_header(header)
+
+    def _parse_header(self, header):
+        fields = []
+        cols = []
```
ptormene: Please add some comments about `fields` and `cols`

ptormene:
Apart from minor comments, LGTM

PR #2070, opened 2016-06-08
Simplified the source manager

micheles:
This is a small refactoring. Now the SourceManager is decoupled from the task function. Moreover the posthook has been removed from the TaskManager.reduce method (it is unneeded now).

The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1907/


ptormene:
LGTM

PR #2073, opened 2016-06-08
webui: show all calculation by default when LOCKDOWN is false

### LINKED ISSUE #2066: Useless WebUI in packages
In the current configuration the WebUI installed by packages is almost useless, because it will not show calculation made by the users (the daemon is run by 'openquake') and any calculation run via the WebUI will be seen only through it (for the same reason). I see three solutions:

1. Do not start the WebUI automatically. If you want to use it in a proper way you have to follow the documentation ad configure django-pam or the django authentication
2. Start the WebUI and show all users calculation (no security). Anyway, any calculation run on the webui will be owned by openquake
3. Provide a WebUI configured with django auth enabled (with an admin account configured). The admin can create new users from the admin. Optionally the admin can add django-pam

I do NOT want to provide also django-pam: to make it work some changes must be done that could heavily affect security on the system (allow django to access `/etc/shadow`). I don't want to take these actions on behalf of other sysadmins/users.

Comments?

daniviga:
Add support for more generic ACLs (instead of just `is_super`) and set a new default, where everyone can see all the calculations if `LOCKDOWN = False` and `ACL_ON = False`

With `ACL_ON = False`
screenshot from 2016-06-08 15-44-21

With `ACL_ON = True`
screenshot from 2016-06-08 15-44-02

https://ci.openquake.org/job/zdevel_oq-engine/1910

Closes #2066


openquake/server/db/actions.py
```diff
@@ -391,15 +391,15 @@ def calc_info(calc_id):
     return response_data
 
 
-def get_calcs(request_get_dict, user_name, user_is_super=False, id=None):
+def get_calcs(request_get_dict, user_name, user_acl_on=False, id=None):
     """
```
micheles: Please change  `user_acl_on` ->  `acl_on`

openquake/server/db/actions.py
```diff
@@ -391,15 +391,15 @@ def calc_info(calc_id):
...
     :returns:
         list of tuples (job_id, user_name, job_status, job_type,
                         job_is_running, job_description)
     """
     # helper to get job+calculation data from the oq-engine database
     jobs = models.OqJob.objects.filter()
-    if not user_is_super:
+    if user_acl_on:
         jobs = jobs.filter(user_name=user_name)
```
micheles: Idem

micheles: Please add a comment here, like

user_acl_on is true if the configuration acl_on is True or if the user is a Django super user.

micheles:

PR #2074, opened 2016-06-08
Fix rup_data export for different TRTs

micheles:
This was discovered by Catarina. In general rup_data files coming from different tectonic region types cannot be concatenated, so we must produce a bunch of files.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1911


ptormene:
LGTM

PR #2079, opened 2016-06-09
Systemd improvements

daniviga:
This PR provides some improvements for systemd/rpm during the upgrade phase:

- Automatically stops the services when OQ is uninstalled
- Avoid the `Warning: Unit file of ***.service changed on disk, 'systemctl daemon-reload' recommended.` warning after an upgrade
- Restart the services, if running, after an upgrade


micheles:
LGTM

PR #2080, opened 2016-06-10
The DbServer automatically create and upgrade the database

micheles:
This was requested by Daniele and it is a good idea now that the migrations are fast and safe.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1927/


daniviga:
LGTM

PR #2082, opened 2016-06-10
Updated the documentation after the introduction of "oq"

micheles:
... and deprecated --rh and --rr in favor of --run, there are too many ways of doing the same thing.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1931/


daniviga:
LGTM;M

PR #2084, opened 2016-06-11
Removed the sitemesh from the datastore

micheles:
It was redundant since it can be extracted from the site collection. The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/1935


ptormene:
LGTM

Issue #2086, opened 2016-06-13
oqdata is not created on Windows and the db server is in the wrong dir


daniviga:
Latest master on Windows does not create the oqdata dir when the dbserver is started and the DB is saved in the user home instead of home/oqdata, even if the settings are

```
multi_user = false
file = ~/oqdata/db.sqlite3
log = ~/oqdata/dbserver.log
```

PR #2087, opened 2016-06-13
The dbserver now performs a makedirs call if needed

micheles:
Should close https://github.com/gem/oq-engine/issues/2086.


daniviga:
Please move the snippet before line 155: https://github.com/gem/oq-engine/blob/cf82b0d3a9ba6179ff07a772d7af520f0a2d695f/openquake/server/dbserver.py#L115

daniviga:
LGTM

PR #2089, opened 2016-06-13
Moved the setting of OQ_DISTRIBUTE in oq

micheles:
Otherwise the variable was set too late. Everything is green: https://ci.openquake.org/job/zdevel_oq-engine/1938


daniviga:
LGTM

PR #2091, opened 2016-06-14
Removed version module

micheles:
This was unused (I did not even know it existed). Also, I turned some docstrings in the tests into comments to have a more readable output when running nosetests. There was an issue with the exports: now the default is 'xml,csv' as in the past, and not ''.

The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/1940/


daniviga:
LGTM

Issue #2075, opened 2016-06-08
Graeme wants the hazard curves CSV files to be split per IMT


micheles:
This should be the default behavior when exporting from the engine. oq-lite can continue as it is now.

PR #2096, opened 2016-06-15
The engine exports the hazard curves one file per IMT

micheles:
Closes https://github.com/gem/oq-engine/issues/2075 with a hack: `oq export` continues to export all the IMTs in the same file, so no test breaks, but `oq engine` exports multiple files.

See https://ci.openquake.org/job/zdevel_oq-engine/1946/


ptormene:
LGTM

PR #2097, opened 2016-06-15
Added __codename__

micheles:
Now it will appear in the calculation logs, see https://ci.openquake.org/job/zdevel_oq-engine/1947

### DIFF
--- a/openquake/risklib/__init__.py
+++ b/openquake/risklib/__init__.py
@@ -33,3 +33,5 @@ if path:
     sys.exit('Found an obsolete version of commonlib; '
              'please remove %s and/or fix your PYTHONPATH'
              % os.path.dirname(path))
+
+__codename__ = 'Keiiti Aki'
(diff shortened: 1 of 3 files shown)


openquake/risklib/__init__.py
```diff
@@ -33,3 +33,5 @@
     sys.exit('Found an obsolete version of commonlib; '
              'please remove %s and/or fix your PYTHONPATH'
              % os.path.dirname(path))
+
+__codename__ = 'Keiiti Aki'
```
daniviga: We should keep the `__codename__` empty and populate it only during the release process.

micheles: Agreed

daniviga:
LGTM

micheles:
After discussion with Daniele we decided to not put the `__codename__` in the code for the moment, just in the README and in the communications.

PR #2100, opened 2016-06-16
Bind the DbServer

micheles:
There is a small bug for `oq engine --version-db`: if the Django connection is not bound to the database, it produces an error, first discovered by GFZ:

```python
  File "/usr/lib/python2.7/dist-packages/openquake/server/db/actions.py", line 369, in version_db
    return upgrade_manager.version_db(conn)
  File "/usr/lib/python2.7/dist-packages/openquake/server/db/upgrade_manager.py", line 359, in version_db
    upgrader = UpgradeManager.instance(conn, pkg_name)
  File "/usr/lib/python2.7/dist-packages/openquake/server/db/upgrade_manager.py", line 316, in instance
    curs = conn.cursor()
AttributeError: 'NoneType' object has no attribute 'cursor'
```

I am also adding a command `oq dbserver` which is useful to stop/start the DbServer in single user mode. Tests and demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/1948


daniviga:
Do I need to change the init scripts? I like the new way of starting the dbserver. I will open a separate PR.

LGTM

micheles:
No, don't change the scripts. This way is only meant for single-user mode.

daniviga:
LGTM

PR #2107, opened 2016-06-17
Changed the names of the exported risk files

micheles:
Closes https://github.com/gem/oq-risklib/issues/724. The tests are running here:  https://ci.openquake.org/job/zdevel_oq-engine/1959/


CatalinaYepes:
LGTM

Note: For `Scenario Risk` the output name for `losses_by_asset` have the realization number (e.g. `losses_by_asset-rlz-000.csv`) instead of the GSIM name (`losses_by_asset-ChiouYoungs2008().csv`) because the engine is using the same exporter as the Event Based.

PR #2111, opened 2016-06-21
Quoted the taxonomies in the CSV export

micheles:
This is needed since the taxonomies may contains commas. The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1963

PS: while at that, I also removed the redundant method `CompositeRiskModel.loss_type_dt`.


ptormene:
LGTM

PR #2112, opened 2016-06-21
Removed some dbserver start/stop in the packager

micheles:
This should solve the problem in https://ci.openquake.org/job/master_oq-engine/2532/console. The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/1964


daniviga:
LGMT

PR #2114, opened 2016-06-21
Webui command

daniviga:
Add a command to start the WebUI using `oq`. The django development server is started in foreground (that's a feature, not a bug) on localhost:8800. Meant to be used with `multi_user = false` but works also if true.

I want this to simplify the start of the WebUI in the 'opt' packages and also in development, it's shorter than `python -m openquake.server.manage ecc...` and it also checks that DbServer is running. If the DbServer is not started and `manage.py runserver` is fired no errors appear on the screen and the django server remains locked until `CTRL-C` is issued.

This feature is currently not tested by CI.

Closes https://github.com/gem/oq-installers/issues/20

### DIFF
--- a/openquake/commands/webui.py
+++ b/openquake/commands/webui.py
@@ -24,9 +24,9 @@ from openquake.server.dbserver import get_status
 from openquake.commands.dbserver import runserver
 
 
-def rundjango(subcmd='runserver', hostport='127.0.0.1:8800'):
+def rundjango(subcmd, hostport='127.0.0.1:8800'):
     args = [sys.executable, '-m', 'openquake.server.manage', subcmd]
-    if host:
+    if hostport:
         args.append(hostport)
     subprocess.call(args)
 
@@ -44,10 +44,11 @@ def webui(cmd):
         runserver()
 
     if cmd == 'start':
-        rundjango()
+        rundjango('runserver')
     elif cmd == 'syncdb':
-        rundjango(cmd)
+        rundjango('syncdb')
 
 parser = sap.Parser(webui)
 parser.arg('cmd', 'webui command',
            choices='start syncdb'.split())
+parser.arg('hostport', 'a string of the form <hostname:port>')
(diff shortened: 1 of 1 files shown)


openquake/commands/webui.py
```diff
@@ -0,0 +1,53 @@
...
+import subprocess
+from openquake.risklib import valid
+from openquake.commonlib import sap
+from openquake.engine import config
+from openquake.server.dbserver import get_status
+from openquake.commands.dbserver import runserver
+
+
+def rundjango(subcmd, host=''):
+    args = [sys.executable, '-m', 'openquake.server.manage', subcmd]
```
micheles: I would change this to

hostport='127.0.0.1:8800'

daniviga:
Ping @micheles

openquake/commands/webui.py
```diff
@@ -0,0 +1,53 @@
...
+from openquake.server.dbserver import get_status
+from openquake.commands.dbserver import runserver
+
+
+def rundjango(subcmd='runserver', hostport='127.0.0.1:8800'):
+    args = [sys.executable, '-m', 'openquake.server.manage', subcmd]
+    if host:
+        args.append(hostport)
+    subprocess.call(args)
+
```
micheles: I would replace `subcmd='runserver' -> subcmd`; also `if host -> if hostname`

openquake/commands/webui.py
```diff
@@ -0,0 +1,53 @@
...
+    if dbstatus == 'not-running':
+        if valid.boolean(config.get('dbserver', 'multi_user')):
+            sys.exit('Please start the DbServer: '
+                     'see the documentation for details')
+        runserver()
+
+    if cmd == 'start':
+        rundjango()
+    elif cmd == 'syncdb':
+        rundjango(cmd)
```
micheles: I would write `rundjango('runserver')` and then `rundjango('syncdb')`

openquake/commands/webui.py
```diff
@@ -0,0 +1,53 @@
...
+
+parser = sap.Parser(webui)
+parser.arg('cmd', 'webui command',
+           choices='start syncdb'.split())
```
micheles: Please add a line
`parser.arg('hostport', 'a string of the form <hostname:port>')`

micheles:
LGTM

PR #2119, opened 2016-06-24
Updated the logic tree parser to use the new ValidatingXmlParser

micheles:
Another 170 lines saved :-) The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/1978/

PS: as a side-effect, it also closes https://github.com/gem/oq-risklib/issues/722


ptormene:
LGTM

Issue #2102, opened 2016-06-16
Reusing the exposure model does not work anymore


micheles:
Probability we should remove the documentation in the manual and then remove the feature.

raoanirudh:
Removed from manual: https://github.com/GEMScienceTools/oq-engine-docs/commit/44461221ffbdb8d7c6560cf6f717c844b510b0ce

PR #2120, opened 2016-06-27
Removed reuse exposure feature

micheles:
It was buggy; for instance in the classical_damage tests below there is an asset `a6` that should have been discarded because outside of the region constraint but it was not. The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1985/

Closes https://github.com/gem/oq-engine/issues/2102.

NB: there is no need to reuse to exposure, since we can parse a million of assets in a few minutes now, parsing the exposure is never a bottleneck.


ptormene:
LGTM

PR #2123, opened 2016-06-28
Taxonomies with chars not in the range a-z0-9 are now accepted

micheles:
The bug was discovered by Catalina and Desmond for the Burundi calculation. The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/1993


ptormene:
LGTM

PR #2126, opened 2016-06-28
Fix the MANIFEST, files used by tests were missing

daniviga:
This is needed to run tests from an installation (using setuptools or whatever)


micheles:
LGTM

PR #2132, opened 2016-07-02
Python3 travis

daniviga:
Tests will run on travis using our python 3.5 distribution: see https://travis-ci.org/daniviga/oq-engine/builds/141671970

Marked as WIP since there are a couple of tests failing (import of commonlib, engine from baselib) because nose get mess when tries to import packages from the same namespace (openquake) and the namespace is located in two different locations (ie mixing sources and installed).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,5 @@ install:
 # the full 'openquake' namespace is overwritten and then hazardlib and baselib are not found
 script:
   # FIXME --with-doctest does not work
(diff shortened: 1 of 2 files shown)


micheles:
Does the situation improve if we do not install hazardlib, just unzip the code and put it in the PYTHONPATH?

daniviga:
Now it should work: https://travis-ci.org/daniviga/oq-engine/builds/142152827#L169

micheles:
LGTM

Issue #2129, opened 2016-07-01
When exporting from the Web UI an old calculation, a clear message must be given


micheles:
One should change openquake.engine.export.core; in case of an export error, the version in the HDF5 file should be compared with the version of the exporter, and if they are different the issue should be pointed out.

PR #2137, opened 2016-07-05
Added a check on the version in case of export errors

micheles:
Closes https://github.com/gem/oq-engine/issues/2129

### DIFF
--- a/openquake/engine/export/core.py
+++ b/openquake/engine/export/core.py
@@ -26,7 +26,7 @@ import zipfile
 
 from openquake.commonlib.export import export
 from openquake.commonlib import datastore
-from openquake.engine import logs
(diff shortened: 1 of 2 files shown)


openquake/engine/export/core.py
```diff
@@ -61,8 +75,9 @@ def export_from_db(output_key, calc_id, datadir, target):
         try:
             exported = export(output_key, dstore)
         except KeyError:
```
ptormene: Please check if this is needed also in case of other kinds of errors

ptormene:
Besides the comment, LGTM

openquake/engine/export/core.py
micheles: Done

PR #2144, opened 2016-07-07
Better error if rupture_mesh_spacing is missing

micheles:
The error must be visible immediately, even before starting the calculation.


daniviga:
LGTM (merge conflicts first)

PR #2146, opened 2016-07-08
Not using job.delete()

micheles:
Django 1.3.1 was the reason for this failure in Ubuntu 12.04: https://ci.openquake.org/job/master_oq-engine/2581/console
I have solved by deleting with a raw query. In the future I would like to remove Django completely,
Green run: https://ci.openquake.org/job/zdevel_oq-engine/2034/consoleFull


daniviga:
LGTM

Issue #2140, opened 2016-07-06
Add pyshp 1.2.3 to the opt distribution


micheles:
Then you can remove the line
```
  - pip install pyshp==1.2.3
```

in .travis.yml.

PR #2148, opened 2016-07-08
Use our binary wheels with Travis

daniviga:
This should speedup the build and save our bandwith. Closes https://github.com/gem/oq-engine/issues/2140


micheles:
LGTM

Issue #2145, opened 2016-07-07
Earlier check on the taxonomies


micheles:
If `oqparam.job_type == 'risk'`, the check on the taxonomies (i.e the `RuntimeError('The exposure contains the taxonomies %s which are not in the risk model' % missing)`) must go before the computation of the hazard, assuming there is a single job.ini. There should be already a test that should be updated.

PR #2149, opened 2016-07-08
Restored the early check on missing taxonomies

micheles:
The check was there, but has been broken by a refactoring in the engine 2.0 branch. I have restored it and added a test. Closes https://github.com/gem/oq-engine/issues/2145.


daniviga:
LGTM

PR #2151, opened 2016-07-11
Small refactoring of the argument parsing framework

micheles:
Now functions which receive their arguments from the command line are marked with the Script decorator, since I feel it is clearer. The demos are running here:https://ci.openquake.org/job/zdevel_oq-engine/2035/


openquake/commonlib/sap.py
```diff
@@ -81,13 +81,15 @@ def str_choices(choices):
...
 
-class Parser(object):
+class Script(object):
     """
     A simple way to define command processors based on argparse.
     Each parser is associated to a function and parsers can be
     composed together, by dispatching on a given name (if not given,
     the function name is used).
     """
+    registry = {}  # dotname -> parser
```
daniviga: This comment isn't very clear to me

daniviga:
Apart from minor comment LGTM

openquake/commonlib/sap.py
micheles: I have improved the comment

PR #2154, opened 2016-07-12
Fixed bug in the UCERF calculator

micheles:
The SourceGroup .id was always zeros, whereas it should have been equal to the branch ordinal.


g-weatherill:
LGTM

PR #2157, opened 2016-07-12
Revision to UCERF calculator

g-weatherill:
The UCERF SES calculator is modified to reflect changes in out understanding of the model. Background grid models are different for each branch, and are hence defined explicitly per branch in the files. The background source generator, and the hdf5 file are adapted to reflect this.


micheles:
LGTM

PR #2158, opened 2016-07-13
Do not save filtered away probability maps

micheles:
Otherwise one gets the error

```python
  File "/usr/lib/python2.7/dist-packages/openquake/calculators/classical.py", line 287, in post_execute
    gsims = self.rlzs_assoc.gsims_by_trt_id[trt_id]
KeyError: 0
```

signaled here: https://groups.google.com/d/msg/openquake-users/KgzyA15ideo/sO5nUUW-BQAJ

Here the problem was that the tectonic region type number 0 was empty: all the relative sources were filtered away and the datastore contained an empty dataset.


ptormene:
LGTM

Issue #2099, opened 2016-06-16
NRML 0.5 format for the source model


micheles:
Here is one proposal, which is mimicking what we did for the risk files in the migration NRML 0.4 -> 0.5:

- [x] we will keep backward compatibility forever, i.e. the engine will be able to read NRML 0.4 files and to upgrade them on-the-fly automatically and no warning will be printed
- [x] there will be a command `oq upgrade_nrml <dir>` to bulk-upgrade all the source models in the given directory
- [x] the new format will be the same as before, except that now the source model will be a sequence of SourceGroups containing sources instead of a sequence of sources
- [x] attributes of the SourceGroups will be transmitted to the underlying sources
- [x] the SourceGroup must have a mandatory attribute tectonicRegion
- [x] the number of srcs_weights is equal to the number of sources in the group and the sum of the
weights is one, if present.

Here is an example

```xml
<?xml version='1.0' encoding='utf-8'?>
<nrml xmlns:gml="http://www.opengis.net/gml"
      xmlns="http://openquake.org/xmlns/nrml/0.5">
  <sourceModel name="Test source model" grp_interdep='indep'>
    <sourceGroup name="group1"  tectonicRegion="Stable Continental Crust"
                 src_interdep="indep" rup_interdep="mutex"
                 srcs_weights="0.1 0.2 0.7">
      <pointSource id="1" name="point" >
        ...
      </pointSource>
      <pointSource id="2" name="point" >
        ...
      </pointSource>
      <pointSource id="3" name="point" >
        ...
      </pointSource>
    </sourceGroup>
    <sourceGroup name="group2"  tectonicRegion="Stable Continental Crust"
                 src_interdep="indep" rup_interdep="indep"
                 srcs_weights="0.1 0.2 0.7">
      <pointSource id="4" name="point">
        ...
      </pointSource>
      <pointSource id="5" name="point">
        ...
      </pointSource>
      <pointSource id="6" name="point">
        ...
      </pointSource>
    </sourceGroup>
        ...
  </sourceModel>
</nrml>
```

An alternative proposal is to make the parser smarter, so that it can read both NRML 0.4 and NRML 0.5, and also mixed files containing both sources and SourceGroups: in this case there will be no deprecation of the old format and no suggestion of upgrading it.

mmpagani:
+ I don't have a particular preference for the strategy to be used for maintaining backward compatibility (i.e. smarter parser Vs. upgrading tool). 
[...]

PR #2164, opened 2016-07-25
Transmit attrs

micheles:
Closes the last remaining points in https://github.com/gem/oq-engine/issues/2099: now group attributes are transmitted to the sources and the srcs_weights are recognized. Moreover, now `nrml.parse(fname, converter)` parses the given file with the given source converter. The demos are running: https://ci.openquake.org/job/zdevel_oq-engine/2073

### DIFF
--- a/openquake/commonlib/tests/nrml_test.py
+++ b/openquake/commonlib/tests/nrml_test.py
@@ -123,7 +123,7 @@ xmlns:gml="http://www.opengis.net/gml"
         self.assertIn('Could not convert imt->intensity_measure_type: '
                       "Invalid IMT: 'SA', line 8", str(ctx.exception))
 
-    def test_invalid_srcs_weights(self):
+    def test_invalid_srcs_weights_length(self):
         fname = writetmp('''\
 <?xml version="1.0" encoding="utf-8"?>
 <nrml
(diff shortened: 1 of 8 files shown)


openquake/risklib/valid.py
```diff
@@ -835,6 +835,18 @@ def positiveints(value):
     return ints
 
 
+def weights(value):
+    """
+    >>> weights('0.4 0.3')
```
ptormene: The doctest is already self-explanatory. Anyway, it might be good to specify that `value` has to be a string of space-separated numbers.
We might also consider the possibility of accepting numbers such as '1 4', meaning that weights are in the proportion of 1 to 4. These would be transformed automatically, in order to make them sum up to 1, i.e. [0.2, 0.8]

openquake/commonlib/tests/nrml_test.py
```diff
@@ -121,3 +122,65 @@ def test_invalid(self):
             read(fname)
         self.assertIn('Could not convert imt->intensity_measure_type: '
                       "Invalid IMT: 'SA', line 8", str(ctx.exception))
+
+    def test_invalid_srcs_weights(self):
```
ptormene: I think it would be more intuitive to call this test something like `test_invalid_srcs_weights_len`, because weights correctly sum up to 1, but the number of sources is wrong.

ptormene:
Apart from minor comments, LGTM

openquake/risklib/valid.py
micheles: We are already using the conventions that the weights are numbers from 0.0 to 1.0 in the logic tree files, so I want to keep compatibility with the past.

openquake/commonlib/tests/nrml_test.py
micheles: I am changing this.

PR #2168, opened 2016-07-27
Reduce the memory consumption when computing statistics for the hazard curves

micheles:
With this change it is possible to compute the stats for the SHARE computation with ~10 GB of RAM, before it was running out of memory on the cluster.

PS: I have also removed the `siteidx` concept that was a leftover of the past and not needed anymore.

### DIFF
--- a/openquake/commonlib/source.py
+++ b/openquake/commonlib/source.py
@@ -304,15 +304,15 @@ class RlzsAssoc(collections.Mapping):
         return assoc
 
     # used in classical and event_based calculators
-    def combine_curves(self, results):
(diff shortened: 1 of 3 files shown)


openquake/commonlib/source.py
```diff
@@ -304,15 +304,15 @@ def extract(self, rlz_indices, csm_info):
         return assoc
 
     # used in classical and event_based calculators
-    def combine_curves(self, results):
+    def combine_curves(self, rlz, results):
```
ptormene: Please add `rlz` also to the docstring

ptormene:
LGTM

PR #2169, opened 2016-07-28
Used a fixed MAXWEIGHT when there is more than one tile

micheles:
Computations with lots of sites (i.e. tiling calculations) and lots of realizations have the problem of slow tasks. This PR solves the issue by using a small MAXWEIGHT for the tasks, hard coded to 200, which turns out to be a good number (tested on SHARE and South America computations). With the new approach a big computation produces a lot of tasks - before the number of tasks was governed by
concurrent_tasks - but the computational size of a task is constant, so the tasks cannot become arbitrarily long as before. Also, there is no problem with rabbitmq going out of memory because the tasks are spawned slowly: due to tiling, the submission phase can take hours, contrarily to early attempts of following this strategy years ago, before tiling.

If `num_sites < sites_per_tile`, i.e. there is a single tile, nothing changes, the number of tasks is still bounded by concurrent_tasks. Since there are few sites, the problem of slow tasks is less serious. 
Still, we are splitting a bit more, the maxweight is half as before.

NB: several expected files have been regenerated. It is well known that changing the maxweight and thus the splitting algorithm changes (slightly) the numbers This is due to numerical precision issues and has nothing to do with the seeds (there are no random seeds in classical hazard).

The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2079.


openquake/commonlib/source.py
```diff
@@ -757,9 +759,11 @@ def __init__(self, csm, maximum_distance,
...
                 self.src_serial[src.id] = rup_serial[start:start + nr]
                 start += nr
-        # decrease the weight with the number of tiles, to increase
-        # the number of generated tasks; this is an heuristic trick
-        self.maxweight = self.csm.maxweight * math.sqrt(num_tiles) / 2.
+        if num_tiles > 1:
+            self.maxweight = MAXWEIGHT  # use the default
+        else:
+            self.maxweight = math.ceil(
+                csm.weight / self.concurrent_tasks) / 2.
```
ptormene: Please add a comment about `2`

ptormene:
LGTM

PR #2171, opened 2016-07-29
Fixed an horrible bug with hdf5.array_of_vstr

micheles:
For some reason on Wilson the loss_type turns out to be a numpy array instead of a string, therefore a cast is needed. TODO: add a test reduced from Alehandro's files, since the issue never happens with the current tests.
The demos run: https://ci.openquake.org/job/zdevel_oq-engine/2083/

NB: the problem affects only Ubuntu 14.04:
```
> /home/michele/py2/oq-engine/openquake/risklib/riskinput.py(80)__getitem__()
-> values = {lt: a[lt] for lt in self.loss_types}
(Pdb)  self.loss_types[0]
structural
(Pdb)  type(self.loss_types[0])
<type 'numpy.object_'>
```

there is no issue in Ubuntu 16.04, where `self.loss_types[0]` is just a string.

### DIFF
--- a/openquake/risklib/riskinput.py
+++ b/openquake/risklib/riskinput.py
@@ -76,7 +76,7 @@ class AssetCollection(object):
     def __getitem__(self, indices):
         if isinstance(indices, int):  # single asset
             a = self.array[indices]
-            values = {lt: a[lt] for lt in self.loss_types}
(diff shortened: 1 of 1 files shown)


ptormene:
If demos are green, LGTM

PR #2173, opened 2016-07-29
Fixed a bug with the initial accumulator when computing hazard curves

micheles:
The initial accumulator has to be a dictionary `grp_id -> ProbabilityMap()`; before it was working by chance. Also, I have renamed the expected files in test_case_17 according to the current conventions.


ptormene:
LGTM

PR #2178, opened 2016-08-16
Refined the number of generated tasks

micheles:
Recently we changed the distribution logic in order to generate more tasks (see https://github.com/gem/oq-engine/pull/2169).  However in some circumstances the change produced too many small tasks; also, it removed the ability to constrain the maximum number of generated tasks. I am solving both issues by introducing the following formula for the maxweight:

```python
       self.maxweight = max(
           MAXWEIGHT,  math.ceil(csm.weight / (self.concurrent_tasks * 2 * num_tiles)))
```
with MAXWEIGHT hard-coded to 200 (an heuristic number that is good for several computations). If the number of tiles is large and/or the number of concurrent tasks is large, the maxweight will be bounded by MAXWEIGHT, i.e. we will not generate a lot of ultra-small tasks. On the other hand, in the normal situation, the maxweight will by given by the usual formula and the number of tasks generated will be around `concurrent_tasks * 2 * num_tiles`.

The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/2092

NB: the numbers change as always because the source splitting procedure depends on `self.maxweight`.


ptormene:
If tests are green, LGTM

PR #2179, opened 2016-08-17
Generate the hazard maps at export time

micheles:
The UHS curves are not a real output of the engine, i.e. they are not stored in the datastore, just generated dynamically at export time. With this pull request the same approach is used also for the hazard maps. This is possible since the calculation of the hazard maps from the hazard curves is very fast.

The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/2093/


ptormene:
LGTM

PR #2181, opened 2016-08-17
Revolutioned the distribution framework

micheles:
This pull requests implements HUGE improvements to the distribution framework.

1. The `litetask` decorator is gone. The reason for its existence was to support celery: celery tasks must be decorated. However after 4 years I realized that it is enough to have a single celery task launcher that has to be decorated: all the other task functions can be left undecorated. This means that every time a new task function is added there is no need to touch celeryconfig.py. Celery is now on equal footing with all the other concurrency frameworks supported by the engine.

2. `parallel.apply_reduce(func, args, agg=agg, acc=acc)` is gone. Now you should use the syntax `parallel.apply(func, args).reduce(agg, acc)`.

3. the Starmap class and subclasses have been fixed: now they create a pool of processes only if they are instantiated.

4. The engine supports `ipyparallel` i.e. the distribution mechanism of IPython.

The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/2096


openquake/commonlib/parallel.py
```diff
@@ -558,8 +524,9 @@ class Threadmap(Starmap):
     >>> Threadmap(Counter, [('hello',), ('world',)]).reduce(acc=Counter())
     Counter({'l': 3, 'o': 2, 'e': 1, 'd': 1, 'h': 1, 'r': 1, 'w': 1})
     """
-    # following the same convention of the standard library, num_proc * 5
-    pool = multiprocessing.dummy.Pool(executor._max_workers * 5)
+    poolfactory = staticmethod(
+        lambda: multiprocessing.dummy.Pool(executor._max_workers * 5))
```
ptormene: Please re-add the comment

ptormene:
LGTM

PR #2182, opened 2016-08-18
Moved the SourceManager logic in the execute phase

micheles:
Before it was in the `pre_execute` method of the base calculator, but since it is needed only in the `execute` method of the `psha` calculator it is best to move it there. The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2099/


ptormene:
LGTM

PR #2183, opened 2016-08-18
Unicode

micheles:
The report for the demo AreaSource is broken with an error
```
Could not generate report: 'ascii' codec can't decode byte 0xe2 in position 15: ordinal not in range(128)
```

see http://artifacts.openquake.org/demos/oq-engine/master/trusty/2659/jobs-2016-08-18.html.

This PR fixes several encoding issues., both in Python 2 and Python 3. Here is the report after the fix: 
http://artifacts.openquake.org/demos/oq-engine/zdevel/trusty/2107/jobs-2016-08-19.html

The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/2107

NB: while working at this pull request I discovered that the maxweight formula was not what I wanted, so I fixed it and I had to change the expected numbers in a few tests.


openquake/engine/tools/make_html_report.py
```diff
@@ -183,9 +183,10 @@ def make_report(isodate='today'):
         (job_id, user, start_time, stop_time, status) = stats[1]
         try:
             ds = read(job_id, datadir=os.path.dirname(ds_calc))
-            txt = view_fullreport('fullreport', ds).decode('utf-8')
+            txt = view_fullreport('fullreport', ds)
             report = html_parts(txt)
         except Exception as exc:
+            raise
```
ptormene: Please remove this

ptormene:
Besides the comment, LGTM

openquake/engine/tools/make_html_report.py
micheles: Thanks, I added it for debugging purposes and I forgot to remove it.

Issue #2017, opened 2016-05-04
We need an event based risk calculator starting from the GMFs


micheles:
In the past most (virtually all) risk calculations running in a cluster where dominated by the risk part, thus it made sense to recompute the GMFs on-the-fly every time. That had the advantage of reducing the data transfer significantly. For instance for Colombia the GMFs are 100 GB, the ruptures, 0.5 GB, so the data transfer is reduced by 200 times by computing the GMFs on the fly.

However things are different now. We introduced recently the `minimum_intensity` feature and that changes the rules of the game. It is now possible to filter away the GMFs below the minimum intensity, and that means that typically 99% of the GMFs can be removed (in the case of Colombia with minimum_intensity PGA=0.1 the 100 GB becomes 0.9 GB). This kind of reduction is typical for our calculations. So it makes sense to compute the GMFs once and to store them since after the filtering they are few and easy to transfer.

Also, now the GMFs are stored in the HDF5 format in an ultra-efficient way (and they will be stored in an even more efficient way in the future, using a new feature of h5py ) so it is not an issue to store even terabytes of them. A storing speed of 500+ MB/s has been measured on Wilson. Transferring hundreds of the gigabytes is achievable too, if the transfer is done carefully, i.e. if you don't try to transfer everything at once.

Due to those advances, saving the GMFs has become reasonable now. Having the GMFs on disk allow different parallelization strategies for the risk that will result in a big speedup. Moreover computing the GMFs in the risk part has a big memory cost, whereas computing them in the hazard part has a much smaller memory cost due to the differences in the algorithm used.

Having a GMF-based calculator may also be a necessary preliminary step for the UCERF calculation, since it is too large to be done with the current calculator (it will likely run out of memory).

Another important use case is the case of a workstation/laptop: when there are few cores, the hazard part is typically dominating the risk calculation, so it makes sense to cache the GMFs to support non-cluster situations.

NB: the current calculator based on the rupture has to be maintained because there are still computations dominated by the risk part and where it makes sense to recompute the GMFs on the fly.
The user will be able to select the old calculator or the new one by specifying the calculation_mode in the job.ini file. `event_based_risk` or `ebr_from_gmf`.

PS: having a calculator starting from the GMFs is needed for two other reasons, apart from performance:
1. it makes it easier to debug and test the calculator
2. a third party can give us the GMFs and we can start the risk from there.

micheles:
[...]

PR #2184, opened 2016-08-21
Simplified the RiskInputs

micheles:
Instead of having a RiskInput object for each IMT, now a RiskInput object contains multiple IMTs. This is a first step towards https://github.com/gem/oq-engine/issues/2017, where it is natural to read the GMFs for all IMTs at the same time. It also allows to reduce the complexity and the number of line of code for all calculators, since the method .get_imt_taxonomies becomes now unneeded.
The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2108


ptormene:
LGTM

PR #2185, opened 2016-08-22
Reduced the data transfer for the RlzsAssoc object

micheles:
We are sending to the workers the full RlzsAssoc object containg all the realizations; however for a task containing sources of a given source group it is enough to send only the realizations associated to that group, which are a lot less. For instance for our demo `hazard/LogicTreeCase2ClassicalPSHA` there are 324 realizations, but only 4 realizations per each source group (there are 81 source groups). With this change the data transfer goes down as follows:

```
   classical_sent.rlzs_assoc       11,840,256 bytes
   classical_sent.rlzs_by_gsim     177,390 bytes
```

Also, before we were sending the full `oqparam` object in the monitor, now we are sending only
the needed attributes, so the data transfer goes down::

```
classical_sent.monitor          589,680 
classical_sent.monitor          162,324 
```

For the complete data transfer in this example we have

```
master
Sent 12.16 MB of data in 162 task(s)
Received 770.69 KB of data, maximum per task 4.76 KB

this branch
Sent 645.31 KB of data in 162 task(s)
Received 361.07 KB of data, maximum per task 2.23 KB
```
The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/2116/


openquake/risklib/riskinput.py
```diff
@@ -585,7 +587,7 @@ def __repr__(self):
             self.__class__.__name__, self.imts, self.weight)
 
 
-def create(GmfColl, eb_ruptures, sitecol, imts, rlzs_assoc,
+def create(GmfColl, eb_ruptures, sitecol, imts, rlzs_by_gsim,
```
ptormene: Please update the docstring accordingly

ptormene:
Apart from a minor comment, LGTM

openquake/risklib/riskinput.py

PR #2190, opened 2016-08-23
Regenerated the CSV reports

micheles:
After the improvement in https://github.com/gem/oq-engine/pull/2185 you can see that the data transfer is a lot less than before. Also, I have fixed the report generation mechanism that had been broken recently.

The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2135


ptormene:
LGTM

PR #2191, opened 2016-08-24
Better monitor

micheles:
A lot of memory is used when unpickling the arguments sent to a task, however until now we were not measuring it. This pull request rectifies the situation. Also the unpickling time is taken into account.
You can see the generated reports here:

http://artifacts.openquake.org/demos/oq-engine/master/trusty/2666/jobs-2016-08-22.html#tabs-15 (master)
http://artifacts.openquake.org/demos/oq-engine/zdevel/trusty/2137/jobs-2016-08-24.html#tabs-15 (this branch)

Focusing on "total event_based_risk", which is unpickling a lot of ruptures, one sees that the maximum allocated memory per task is 31 MB whereas master was giving an incorrect value of only 18 MB.

### DIFF
--- a/openquake/commonlib/parallel.py
+++ b/openquake/commonlib/parallel.py
@@ -103,16 +103,19 @@ def safely_call(func, args, pickle=False):
         if set, the input arguments are unpickled and the return value
         is pickled; otherwise they are left unchanged
     """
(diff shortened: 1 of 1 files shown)


ptormene:
LGTM

Issue #2175, opened 2016-08-01
WebUI is unusable with gunicorn/uwsgi


daniviga:
Somewhere (found: a0bdcf2b9266c04beb898ad70c910e26d7679580) between engine-2.0 and master the WebUI became extremely slow/unresponsive when it's run through gunicorn/uwsgi. This does not happen with the django embedded `runserver` (but it does not scale up).

I've been able to reproduce this issue on different machines (wilson and oats3).

Also the memory usage of the gunicorn workers is much higher with the current master (memory used is (up >8GB vs 800MB-1GB on oats3) and too many processes are spawned:

- with engine-2.0 and 10 gunicorn workers, total number of running web processes are 12 (10 + parent + supervisor) and that's correct.
- with master, 10 workers, total is 252, which is wrong

daniviga:
Some more information:

with engine-2.0 and 10 gunicorn workers, total number of running web processes are 12 (10 + parent + supervisor) and that's correct.
with master, 10 workers total is 252, which is wrong

daniviga:
The culprit is a0bdcf2b9266c04beb898ad70c910e26d7679580, from this point the WebUI is unusable with a wsgi.

micheles:
This is really strange Daniele, but you can try to revert that merge and see if it is really the culprit. We are not using that part anywhere at the moment.

daniviga:
```
dvigano@oats3:/usr/local/openquake/oq-engine# git status
On branch master
Your branch is up-to-date with 'origin/master'.
dvigano@oats3:/usr/local/openquake/oq-engine# ps axu | grep gunic|wc
    252    3778   32217
```

```
dvigano@oats3:/usr/local/openquake/oq-engine# git revert -m 2 a0bdcf2b9266c04beb898ad70c910e26d7679580
## webui restart
dvigano@oats3:/usr/local/openquake/oq-engine# ps axu | grep gunic|wc
     12     178    1497
```

daniviga:
Ok, after more bisecting I discovered that the exact commit is the latest one, not the PR merge: ebffdc5bdc75499ee358a08527285acd7bb2f7e2 (I'm sorry, revert reverted also this one unexpectedly).

daniviga:
With current master I get the following behavior:

- Gunicorn starts normally: if (for example) 4 workers are set, four processes are spawned and the WebUI is well responsive
```bash
dvigano@oats3:~$ ps aux | pgrep gunic | wc -l
5
```
[...]

PR #2194, opened 2016-08-24
Restart the ProcessPoolExecutor at the end of each job

micheles:
This should close https://github.com/gem/oq-engine/issues/2175


daniviga:
Good catch! LGTM

Issue #2195, opened 2016-08-25
Yet another segmentation fault of h5py


micheles:
When running really large classical computations (i.e. SHARE) sometimes the `.save_hcurves` method fails with a segfault. This never happens in regular computations, nor in the tests and in the demos. The reason probably is that such method is writing on an extensible dataset which is very large, thus exciting a bug in h5py (out of memory in the heap??). Even if the computation runs (since the bug is random) it is still slow. Probably we should not use an extensible dataset and go back to regular fixed-size datasets, which are safer and faster. The price to pay is that we will write some zero curves, but it is a small price.

PR #2189, opened 2016-08-22
Compute the statistics in parallel

micheles:
This the companion of https://github.com/gem/oq-hazardlib/pull/482 and it is needed to close https://github.com/gem/oq-engine/issues/2180.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/2139

PR #2197, opened 2016-08-25
Save the hazard curves without resizing the datasets

micheles:
After the monster change https://github.com/gem/oq-engine/pull/2189 some bug was introduced without noticing. This change closes https://github.com/gem/oq-engine/issues/2195. The saving time goes down significantly:
 
```
   saving hcurves and stats       1,891 seconds (master)
   saving hcurves and stats      153 seconds (this branch)
```

The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/2147


daniviga:
LGTM

PR #2198, opened 2016-08-29
Ucerf classical 1

micheles:
The calculator was written by Graeme and I did some minor bug fixes. It is still missing a test but at least it does not break anything.


micheles:
Merging since there is a (preliminary) tests now.

PR #2200, opened 2016-08-29
Made the TaskManager lazy

micheles:
Before `TaskManager(func, arglist)` was submitting the tasks at instantiation time. Now the tasks are submitted only when you start iterating on the instance. This is consistent with the way multiprocessing works and has the big advantage that it is possible to transfer back a TaskManager instance from a task. This will make it possible to implement hierarchies of tasks and subtasks (in the future).
The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2149/


ptormene:
LGTM

PR #2202, opened 2016-08-30
Moved the contextmanager GroundShakingIntensityModel to hazardlib

micheles:
So that the parallelization infrastructure become independent from hazardlib. The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/2152


ptormene:
LGTM

Issue #2150, opened 2016-07-11
CSV formats for event based risk outputs


raoanirudh:
#### Stochastic event sets

The current CSV exporter for stochastic event sets provides very limited information regarding the ruptures. Instead, the stochastic event set parsers from [nrml_converters](https://github.com/GEMScienceTools/nrml_converters) should be used to generate the CSV outputs in a format similar to that listed below:

| event_id | event_set | tectonic_region | source_id | rupture_id | magnitude | centroid_lon | centroid_lat | centroid_depth | strike | dip | rake | boundary |
| :-: | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1 | 0003 | Subduction IntraSlab | Inslab_west_coast_4_1997_1997_0 | 251609-01 | 7.3 | -75.82 | 6.09 | 101.22 | 11.68 | 53.83 | -90 | MULTIPOLYGON(((-75.964360294 5.87265701971,-75.8674308742 6.33919340539,-75.6827673987 6.30122649939,-75.7798502417 5.83472306571,-75.964360294 5.87265701971))) |
| 2 | 0006 | Subduction IntraSlab | Inslab_west_coast_4_2123_2123_0 | 252497-01 | 7.3 | -75.74 | 6.18 | 103.34 | 11.72 | 60.48 | -90 | MULTIPOLYGON(((-75.8641816798 5.95834116722,-75.766942107 6.42481703028,-75.6127751264 6.39302933666,-75.7101446147 5.92658146426,-75.8641816798 5.95834116722))) |
| 3 | 0008 | Subduction IntraSlab | Inslab_west_coast_4_3246_3246_0 | 260998-01 | 7.8 | -75.27 | 5.96 | 159.35 | 11.53 | 80.82 | -90 | MULTIPOLYGON(((-75.3936883327 5.58000060929,-75.2339970802 6.35914892305,-75.1506570513 6.34224559907,-75.3104616445 5.56312127024,-75.3936883327 5.58000060929))) |
| 4 | 0013 | Subduction IntraSlab | Inslab_west_coast_4_3997_3997_1 | 266777-01 | 7.3 | -76.59 | 6.11 | 102.37 | 11.45 | 40.12 | -90 | MULTIPOLYGON(((-76.752728967 5.90030145191,-76.6577578255 6.36723670654,-76.4182802189 6.31902689747,-76.5134516092 5.85213367763,-76.752728967 5.90030145191))) |
[...]

PR #2203, opened 2016-08-30
Changed the event loss table exporter

micheles:
First step towards https://github.com/gem/oq-engine/issues/2150. The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/2161/


raoanirudh:
We should remove the `:float32` portion from the loss category strings in the csv file headers if possible.

micheles:
I was able to remove it from everything except  the field `occupants`, which is rather special. I would live that case for later since in the future it would be nice to have occupants become more similar to a loss type.

raoanirudh:
LGTM

PR #2152, opened 2016-07-11
Removed the django ORM

micheles:
The goal is to make the DbServer Django-independent. The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/2101
Here is the generated documentation for the new dbapi module:
http://artifacts.openquake.org/zdevel_oq-engine/workspace_2101/doc/sphinx/build/html/openquake.server.html#module-openquake.server.dbapi

PR #2209, opened 2016-09-01
Fix WebUI logs

daniviga:
Fixes a regression introduced in #2152 with the change of `is_running = false|true` with `is_running = 0|1`.
It also add a local copy of a broken js dep. URL.

We still need to make the status "executing" when running the calculation.


daniviga:
LGTM on the `set_status`

micheles:
LGTM if the tests are green

PR #2210, opened 2016-09-01
Avoid recomputing the background sources multiple times and code cleanup

micheles:
I am updating the ucerf_classical calculator to be more similar to the others, reusing some code and removing some unneeded/duplicated stuff. Also, the computation of the background sources was inside the `rupset_idx` loop, probably due to an indentation error. Now it is done externally to the task, and it is done in parallel.


daniviga:
LGTM on trust (but at least tests are now green on my machine too)

Issue #1965, opened 2016-03-02
'MultiPolygon' object has no attribute 'exterior' error


micheles:
This happens when filtering some sources for large enough integration distances. I am attaching a case exhibiting the error.

[indonesia.zip](https://github.com/gem/oq-engine/files/1630405/indonesia.zip)

```python
  File "/home/michele/oq-engine/openquake/commonlib/source.py", line 771, in filter
    for src, _sites in src_filter(src_group.sources):
  File "/home/michele/oq-engine/openquake/hazardlib/calc/filters.py", line 401, in __call__
    box = self.get_affected_box(src)
  File "/home/michele/oq-engine/openquake/hazardlib/calc/filters.py", line 358, in get_affected_box
    return src.get_bounding_box(maxdist)
  File "/home/michele/oq-engine/openquake/hazardlib/source/base.py", line 141, in get_bounding_box
    return self.get_rupture_enclosing_polygon(dilation).get_bbox()
  File "/home/michele/oq-engine/openquake/hazardlib/source/simple_fault.py", line 150, in get_rupture_enclosing_polygon
    return polygon.dilate(dilation)
  File "/home/michele/oq-engine/openquake/hazardlib/geo/polygon.py", line 195, in dilate
    return type(self)._from_2d(new_2d_polygon, self._projection)
  File "/home/michele/oq-engine/openquake/hazardlib/geo/polygon.py", line 133, in _from_2d
    xx, yy = numpy.transpose(polygon2d.exterior.coords)
AttributeError: 'MultiPolygon' object has no attribute 'exterior'
```

micheles:
This happens because there is SimpleFaultSource with a dip angle of 90 degrees giving a degenerate polygon which is actually a line.

micheles:
The bug has been solved because `get_rupture_enclosing_polygon` does not exist anymore in engine 3.1 ;-)

PR #2211, opened 2016-09-01
Restored nice error message when the filtering fails

micheles:
It is important to print the source_id. This feature (introduced in https://github.com/gem/oq-engine/issues/1965) was temporarily removed in the recent refactoring of the source manager.


openquake/commonlib/source.py
```diff
@@ -840,13 +863,13 @@ def _gen_args(self, src_data, tiles):
     def _gen_src_light(self, tiles):
         filter_mon = self.monitor('filtering sources')
         tiles = [Tile(tile, self.maximum_distance) for tile in tiles]
-        if self.filter_sources and self.num_tiles == 1:
+        if self.num_tiles == 1:
             logging.info('Filtering light sources')
```
micheles: Don't remove filter_sources!

ptormene:
Please check if this is tested (otherwise add a test), then LGTM

micheles:
Test added

PR #2215, opened 2016-09-05
Reduce the data transfer in realizations

micheles:
Closes https://github.com/gem/oq-engine/issues/2214. The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2174.
In the case of `demos/hazard/LogicTreeCase2ClassicalPSHA`, with 324 realizations, the data transfer in rlzs_by_gsim goes down by a factor of 7, from 177,390 bytes to  25,434  bytes. In more realistic cases you can gain several orders of magnitude. Here is a real case with an improvement of over 10,000 times:

```
   (master)
   count_eff_ruptures_num_tasks             56
   count_eff_ruptures_sent.monitor          52,248
   count_eff_ruptures_sent.rlzs_by_gsim     850,059,977
   count_eff_ruptures_sent.sitecol          27,608
   count_eff_ruptures_sent.sources          13,651,404

   (this branch)
   count_eff_ruptures_num_tasks             56
   count_eff_ruptures_sent.monitor          53,536
   count_eff_ruptures_sent.gsims            79,958
   count_eff_ruptures_sent.sitecol          27,608
   count_eff_ruptures_sent.sources          13,651,404
```

This is a computation for Switzerland by Laurentiu with 82,944 realizations which was totally dominated by the rlzs_by_gsim data transfer (811 MB vs 13 MB of sources) which goes down to 7 k. There are computations with much larger logic trees where the effect is even bigger.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,5 @@
   [Michele Simionato]
+  * Reduced the data transfer due to the `rlzs_by_gsim` parameter
   * Added an HDF5 export for scenario GMFs
   * If `filter_sources` if false, the light sources are not filtered, but the
     heavy sources are always filtered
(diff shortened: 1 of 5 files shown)


ptormene:
LGTM

PR #2216, opened 2016-09-05
Store the ruptures in the aggregation phase

micheles:
At the moment the ruptures are kept in memory and stored at the end of the calculation in the EventBasedRuptureCalculator. The saving time is a lot and a lot of memory is used. With this pull request the saving time is spread in the aggregation phase while waiting for the tasks to complete. This also makes it possible further optimizations that will be implemented in a following pull request. This is another step towards https://github.com/gem/oq-engine/issues/2017. The demos are green: https://ci.openquake.org/job/zdevel_oq-engine/2173/


openquake/calculators/event_based.py
```diff
@@ -50,6 +49,10 @@
 event_dt = numpy.dtype([('eid', U32), ('ses', U32), ('occ', U32),
                         ('sample', U32)])
 
+long_event_dt = numpy.dtype([
```
ptormene: Please change into `stored_event_dt`

micheles: Done

ptormene:
LGTM

PR #2218, opened 2016-09-05
Removed multi hazard curve xml writer

micheles:
This is not used anymore. The demos are green: https://ci.openquake.org/job/zdevel_oq-engine/2176


ptormene:
If also demos are green, LGTM

PR #2219, opened 2016-09-06
Fixed the case of OQ_DISTRIBUTE=no

micheles:
When disabling the task distribution there was an error when flushing the monitor. This has been solved here. I am also changing `logictree.get_effective_rlzs` to keep the original ordinal, which is useful when debugging the reduction of logic trees,


ptormene:
LGTM

PR #2220, opened 2016-09-06
Restored a weight of 1 for each rupture

micheles:
This is simpler and actually ensures a better task distribution when computing the GMFs. For instance, in the case of a computation for Chile on my workstation I got

```
~/sara_v4$ oq show task_info # master
================================= ==== ====== =========== === =========
measurement                       mean stddev min         max num_tasks
compute_gmfs_and_curves.time_sec  96   114    3.540       516 23       
================================= ==== ====== =========== === =========
~/sara_v4$ oq show task_info # this branch
================================= ==== ====== ========= === =========
measurement                       mean stddev min       max num_tasks
compute_gmfs_and_curves.time_sec  98   34     3.488     126 23       
================================= ==== ====== ========= === =========
```

i.e. the standard deviation in the times per task went down from 114 seconds to 34 seconds, with an average around 97 seconds. The total runtime went down from 531s to 282s, i.e. nearly half.


ptormene:
LGTM

PR #2221, opened 2016-09-06
Added weight information in task_info

micheles:
In theory, the weight of a task should be proportional to its runtime. In practice, this is a working hypothesis that has be checked. When it is wrong the weighting algorithm has to be changed, as we did in https://github.com/gem/oq-engine/pull/2220. In order to have the weight information, it has to be stored. This pull request provides such functionality as well as a way to show the information by extending the `task_info` view.

Example of usage:

```
$ oq run demos/hazard/AreaSourceClassicalPSHA/job.ini
$ oq show task_info:classical
========= ====== ====== ========
rduration taskno weight duration
========= ====== ====== ========
0.551     4      8.000  4.407   
0.532     3      11     5.848   
0.540     1      11     5.936   
0.543     2      11     5.969   
========= ====== ====== ========
```

`rduration = duration / weight` is the reduced task duration in seconds per unit of weight. It should be
more or less constant which is the case in this example.

The demos are green: https://ci.openquake.org/job/zdevel_oq-engine/2180


ptormene:
LGTM

PR #2223, opened 2016-09-06
Do not save the ruptures if the ground motion fields are saved

micheles:
When computing the GMFs a lot of time is spent in saving the ruptures and then reading them from the datastore. Since they are already in memory we can use them simply by adding a .result attribute to the precalculator. Once the ebr_from_gmfs calculator will be ready, there will be no need to store the ruptures when the GMFs are stored. This is another step towards https://github.com/gem/oq-engine/issues/2017.

The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2186/


ptormene:
LGTM

Issue #2224, opened 2016-09-07
API endpoint /v1/calc/list is broken


daniviga:
The API endpoint `/v1/calc/list` is currently broken:

```python
Environment:


Request Method: GET
Request URL: http://localhost:8800/v1/calc/list

Django Version: 1.8.7
Python Version: 2.7.12
Installed Applications:
('openquake.server.db', 'django.contrib.staticfiles', 'openquake.server')
Installed Middleware:
('django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware')


Traceback:
File "/home/daniele/py2/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  132.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/home/daniele/py2/lib/python2.7/site-packages/django/views/decorators/http.py" in inner
  45.             return func(request, *args, **kwargs)
File "/home/daniele/GIT/gem/openquake/oq-engine/openquake/server/views.py" in wrap
  82.             response = func(request, *args, **kwargs)
File "/home/daniele/GIT/gem/openquake/oq-engine/openquake/server/views.py" in calc
  233.                            user['name'], user['acl_on'], id)
File "/home/daniele/GIT/gem/openquake/oq-engine/openquake/engine/logs.py" in dbcmd
  60.         raise etype(res)

Exception Type: OperationalError at /v1/calc/list
Exception Value: 
  File "/home/daniele/GIT/gem/openquake/oq-engine/openquake/commonlib/parallel.py", line 109, in safely_call
    got = func(*args)
  File "/home/daniele/GIT/gem/openquake/oq-engine/openquake/server/db/actions.py", line 452, in get_calcs
    filterdict)
  File "/home/daniele/GIT/gem/openquake/oq-engine/openquake/server/dbapi.py", line 317, in __call__
    raise exc.__class__('%s: %s %s' % (exc, templ, args))
OperationalError: near "ORDER": syntax error: SELECT *, CASE
WHEN calculation_mode LIKE '%risk'
OR calculation_mode LIKE '%bcr'
OR calculation_mode LIKE '%damage'
THEN 'risk'
ELSE 'hazard'
END AS job_type
 FROM job WHERE  ORDER BY id DESC ()
```

I'm investigating why the WebUI is working even with this error (the endpoint should be used by its main page -calculations list-) and this endpoint is not tested.

Always reproducible.

PR #2226, opened 2016-09-07
Fixed the dbapi filtering when there is no filtering

micheles:
This closes https://github.com/gem/oq-engine/issues/2224


daniviga:
LGTM if Travis is happy

PR #2227, opened 2016-09-07
Saved engine_version and hazardlib_version

micheles:
We were saving the engine version but not the hazardlib version. If you need to export an old calculation
the solution is to call

```
$ oq show_attrs / <job_id>
engine_version 2.1.0-gitfdf3d6d
hazardlib_version 0.21.0-git89bccaf
```

and then `git checkout` on the correct commits and run the export. The demos are running here:
https://ci.openquake.org/job/zdevel_oq-engine/2187/


daniviga:
LGTM

PR #2228, opened 2016-09-07
Reduced a big log message

micheles:
When a lot of ruptures are sampled,  `logging.info('Considering %s %s', branch_id, indices)` is trying to log a lot of stuff and the DbServer becomes unresponsive, with the error

```
   RuntimeError: Cannot connect on localhost:1999
```

This was discovered by Anirudh. Now we only log the number of ruptures, not the indices.


daniviga:
LGTM

PR #2230, opened 2016-09-08
Copy deb files only if binaries are built in _pkgbuild_innervm_run

daniviga:
Fixes build issue on Xenial (see also https://github.com/gem/oq-hazardlib/pull/490)


nastasi-oq:
LGTM

PR #2236, opened 2016-09-13
Forbid the usage of oq run on our cluster

micheles:
This is to stop users like Catalina and Marco to use oq-lite on the cluster: we want to be able to see the logs on the Web UI!


daniviga:
LGTM

PR #2238, opened 2016-09-13
Run some demos with celery on Jenkins

micheles:
Before celery was untested; see https://ci.openquake.org/job/zdevel_oq-engine/2194/


daniviga:
LGTM

PR #2239, opened 2016-09-13
Some improvements to the UCERF event based calculator

micheles:
Compute ruptures and GMFs/hazard curves have been merged to save memory. The computation of hazard curves is now tested. If there are no hazard curves for a given realization (typically because the GMFs are under the minimum intensity), no file is exported. The case of missing intensity measure types is correctly flagged as an error.

The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2197

PS: here are some numbers for a calculation with 66 hazard sites and 500 years, small enough to run on Wilson. With master the total memory occupation was around 35 GB, with this branch around 15 GB, i.e. less than half. The runtime has improved too, from  2,135 seconds to 1,648 seconds. We saved 10 GB of data transfer in the ruptures. The memory allocation in the workers was less than 1/3 than before.

### DIFF
--- a/openquake/calculators/event_based.py
+++ b/openquake/calculators/event_based.py
@@ -600,7 +600,8 @@ class EventBasedCalculator(ClassicalCalculator):
             # save individual curves
             if self.oqparam.individual_curves:
                 for i in sorted(result):
-                    self.datastore['hcurves/rlz-%03d' % i] = result[i]
+                    if result[i]:
+                        self.datastore['hcurves/rlz-%03d' % i] = result[i]
             # compute and save statistics; this is done in process
             # we don't need to parallelize, since event based calculations
             # involves a "small" number of sites (<= 65,536)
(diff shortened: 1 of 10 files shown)


openquake/calculators/event_based.py
```diff
@@ -597,7 +600,8 @@ def post_execute(self, result):
             # save individual curves
             if self.oqparam.individual_curves:
                 for i in sorted(result):
-                    self.datastore['hcurves/rlz-%03d' % i] = result[i]
+                    if result[i]:
+                        self.datastore['hcurves/rlz-%03d' % i] = result[i]
```
ptormene: When nothing is saved, we could produce a warning

ptormene:
Apart from a note, LGTM

openquake/calculators/event_based.py
micheles: Warning added

PR #2241, opened 2016-09-13
Xenial needs python-celery-common to run tests using celery

daniviga:
The `celery` utility is now provided by `python-celery-common`.
We must be very careful with the order of the dependencies: if we install `common` before celery, python3 version of celery is installed.

Fixes master.

Tests: https://ci.openquake.org/job/zdevel_oq-engine/2205/


micheles:
LGTM

PR #2242, opened 2016-09-14
Filter the source model upfront when there are less sites

micheles:
Disaggregation calculations produce not enough tasks (for instance the recent on for Brasil produced 57 tasks instead of the wanted 512). The reason is that the splitting in tasks is based on the total weight of the source model, while we should considered only the filtered weight (i.e. the discard the weight of the sources outside the integration distance).
This is the companion of https://github.com/gem/oq-hazardlib/pull/491/
The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/2203/

### DIFF
--- a/openquake/commonlib/source.py
+++ b/openquake/commonlib/source.py
@@ -32,8 +32,6 @@ import numpy
 from openquake.baselib import hdf5
 from openquake.baselib.python3compat import decode, raise_
 from openquake.baselib.general import groupby, block_splitter, group_array
-from openquake.hazardlib.site import Tile
-from openquake.hazardlib.calc.filters import context
 from openquake.commonlib import logictree, sourceconverter
 from openquake.commonlib import nrml, node, parallel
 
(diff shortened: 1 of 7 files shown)


openquake/commonlib/source.py
```diff
@@ -616,7 +593,7 @@ def __init__(self, gsim_lt, source_model_lt, source_models,
         self.gsim_lt = gsim_lt
         self.source_model_lt = source_model_lt
         self.source_models = source_models
-        self.source_info = ()  # set by the SourceFilterSplitter
+        self.source_info = ()  # set by the SourceSitesFilterSplitter
```
daniviga: Check the comment

openquake/commonlib/source.py
```diff
@@ -626,6 +603,37 @@ def __init__(self, gsim_lt, source_model_lt, source_models,
             self.source_model_lt.num_samples,
             [sm.get_skeleton() for sm in self.source_models])
 
+    def filter(self, sitecol, src_sites_filter):
```
daniviga: Maybe should be good to use `ss_filter` everywhere

openquake/commonlib/source.py
```diff
@@ -767,9 +775,8 @@ def split_filter(src, sites, max_dist, random_seed):
             ss.serial = src.serial[start:start + nr]
             start += nr
         ss.id = src.id
-        with context(ss):
-            s = ss.filter_sites_by_distance_to_source(max_dist, sites)
-        if s is not None:
+        ss_sites = ss_filter.affected(ss, sites)
```
daniviga: ss -> sub_source otherwise it makes confusion

openquake/commonlib/source.py
```diff
@@ -880,20 +884,13 @@ def _gen_args(self, src_data, tiles):
...
-            for src in sources:
-                if self.filter_sources:
-                    with filter_mon, context(src):
-                        ok = src in tile
-                else:  # accept all sources
-                    ok = True
-                if ok:
-                    data.append((src, i, [], filter_mon.dt, 0))
+            for src, sites in self.ss_filter(sources, tile):
+                data.append((src, i, [], filter_mon.dt, 0))
```
daniviga: You have lost the monitoring here

daniviga:
Please update the changelog

daniviga:
LGTM

openquake/commonlib/source.py
micheles: Removed

openquake/commonlib/source.py
micheles: Done

openquake/commonlib/source.py
micheles: Used "split" for the split source

openquake/commonlib/source.py
micheles: Restored

Issue #2243, opened 2016-09-14
Improvements to the disaggregation calculator


micheles:
We need to add at least two checks:

1. the parameter poes_disagg must be mandatory
2. the parameter individual_curves must be false

Also, very few tasks are produced in the disaggregation phase: the reason is that the sources are not split. We should split them.

PR #2245, opened 2016-09-15
Improved the disaggregation calculator

micheles:
Closes https://github.com/gem/oq-engine/issues/2243. The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2209/

In the case of Valerio's simplified computation the splitting produces 18 tasks instead of 6: better, even if not spectacular. It means his sources are not much splittable.


micheles:
I should add a changelog after the other branches enter

daniviga:
Please refresh after #2247 merge

PR #2246, opened 2016-09-16
The WebUI now displays only the last 100 calculations

micheles:
This should solve the performance issues of the WebUI. It is possible to set the `start_time` manually in the URL. For instance `/v1/calc/list?start_time=2016-09-01` returns the calculations started in September. However, this is an internal detail not meant for the final user. We could add other features, like a `stop_time`, but this is left for the future.


daniviga:
Mmm this will not be very suitable for most of the uses, like OATS, when we have not so many computations across large time-windows. I would instead
- keep this API
- by default filters by an arbitrary number, like last 100 jobs

PR #2247, opened 2016-09-16
Make sure python-amqp is installed _before_ python-celery

daniviga:
Otherwise, on Trusty, `python-librabbitmq` is installed and it's known to be broken (it causes segfaults).

`python-librabbitmq` is more optimized (has C speedups) but it's broken and works with Python 2 only.

Xenial has same deps. order just to be in sync with Trusty.
See: https://ci.openquake.org/job/zdevel_oq-engine/2216

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -16,9 +16,9 @@ export DH_OPTIONS
 
 bd=openquake/bin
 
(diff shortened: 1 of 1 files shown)


micheles:
LGTM

PR #2248, opened 2016-09-16
Use custom oq-hazardlib branch in travis

daniviga:
This PR allows to build `oq-engine` against a custom `oq-hazardlib` branch if a branch with the same name exists in its repo, otherwise `master` is used.


micheles:
LGTM

PR #2249, opened 2016-09-16
Added a command "oq db"

micheles:
My use case is in helping debugging the queries sent to the DbServer by the WebUI. However, this is also useful for sending commands to the database, for instance, to assess the usage patterns. As an example, I have introduced the command

```
$ oq db get_longest_jobs
```


openquake/server/dbserver.py
```diff
@@ -94,6 +96,18 @@ def get_status(address=None):
     return 'not-running' if err else 'running'
 
 
+def ensure_on():
```
daniviga: That's useful. It can be used in `commands/webui.py` too.

micheles: Done

PR #2250, opened 2016-09-19
Improvements to the ebrisk calculator

micheles:
Now it processes one source model at the time and send the corresponding ruptures in blocks of size `ruptures_per_block` (default 1000). It parallelizes very well and the Chile model goes down from 2 hours to 30  minutes of calculation on the cluster. This is a step towards the solution of https://github.com/gem/oq-engine/issues/2251.

The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/2226


ptormene:
If all tests are green, LGTM

Issue #2252, opened 2016-09-19
The event based risk calculator is using too much memory


micheles:
Catalina is sending the Zurich cluster out of memory. The problem is that we are keeping all the IMTs and all the realizations in memory at the same time when computing the GMFs. In a SARA model with 54 realizations and 3 IMTs it means that we are using 54 x 3 = 162 times more memory than needed.
We should change the calculator to keep in memory only one IMT and realization at the time. This looks hard to implement but possible.

PR #2257, opened 2016-09-20
Reintroduced the hazard getters

micheles:
This is a big refactoring (the first of many) of the risk calculators, a preliminary step for https://github.com/gem/oq-engine/issues/2252. The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2224.
The formal difference is that now the hazard can be extracted for a given imt and rlz:

```
 hazard = hazard_getter.get(imt, rlz)
```

Right now this is just formal, because we still keep everything in the memory, but in the future (after several other pull requests) this call will keep in memory only the GMFs of the specified `imt` and `rlz`.


openquake/risklib/riskinput.py
```diff
@@ -641,16 +700,19 @@ def get_hazard(self, rlzs_assoc, monitor=Monitor()):
...
+        return gg
 
     def __repr__(self):
         return '<%s imts=%s, weight=%d>' % (
             self.__class__.__name__, self.imts, self.weight)
 
 
 def create(GmfColl, eb_ruptures, sitecol, imts, rlzs_by_gsim,
-           trunc_level, correl_model, min_iml, monitor=Monitor()):
+           trunc_level, correl_model, min_iml, monitor=Monitor(), ddic=False):
```
ptormene: Please add the description of ddic to the docstring

PR #2258, opened 2016-09-21
Use the celery backend defined in the celeryconfig.py

daniviga:
This allows us to test redis. We probably will need also to switch from `amqp://` to `rpc://` as recommended in the documentation:
http://docs.celeryproject.org/en/latest/configuration.html#amqp-backend-settings (tkt #2259)

Tests: https://ci.openquake.org/job/zdevel_oq-engine/2230/


micheles:
LGTM

PR #2260, opened 2016-09-22
Three small fixes

micheles:
1. Recently I had removed the saving of the sescollection when the GMFs are generated. This was meant to be temporary and it is the source of the current problems with the --hc functionality. I am restoring the unconditional saving.

2. I am also fixing an error with the management of exceptions when the WebUI export fails, since I discovered this bug today.

3. I am change the expected numbers for ucerf_classical due to the changes in https://github.com/gem/oq-hazardlib/pull/492.

The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/2239/


daniviga:
LGTM

PR #2268, opened 2016-09-28
Fixed the expected values in the classical_risk tests

micheles:
Due to an ordering bug in hazardlib (see https://github.com/gem/oq-hazardlib/pull/498) the classical_risk calculator is producing incorrect results. The effect is only visible when probability map has some zero values (i.e. there are sites outside of the integration distance). For instance we could have something
like this with four sites:

```
correct
site0: nonzero_hazard_curve0
site1: 0 ... 0
site2: nonzero_hazard_curve2
site3: nonzero_hazard_curve3
```

```
incorrect
site0: nonzero_hazard_curve0
site1: nonzero_hazard_curve2
site2: nonzero_hazard_curve3
site3: 0 ... 0
```


micheles:
Good news: I checked that the bug is fairly recent, since it was introduced after the release 2.0. As such it affects only the users of the nightly builds. I am closing it.

PR #2274, opened 2016-10-03
Use rtree wheels made by us

daniviga:
This wheels are experimental, but should work. These are not generated automatically because .so lib requires some patchwork using `patchelf`:

```bash
patchelf --set-rpath \$ORIGIN/.libs --force-rpath _wrapper_lib.so
```

An mock extension is used to be compatible with the `manylinux1` profile: https://github.com/gem/rtree/compare/master...gem:mock-wheel

Companion of: https://github.com/gem/oq-hazardlib/pull/502

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,9 @@ addons:
   apt:
     packages:
     - git
-    - libspatialindex-dev
 
 install:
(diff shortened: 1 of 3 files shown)


micheles:
LGTM

PR #2276, opened 2016-10-04
Enable logging in and out programmatically

ptormene:
This is useful to drive the engine server from external tools such as QGIS plugins.
Unfortunately I wasn't able to do this without disabling `django.middleware.csrf.CsrfViewMiddleware`. Please @daniviga, could you check about it?


doc/web-api.md
```diff
@@ -179,3 +184,18 @@ a JSON object, containing:
     * valid: a boolean indicating if the provided text is a valid NRML
     * error_msg: the error message, if any error was found (None otherwise)
     * error_line: line of the given XML where the error was found (None if no error was found or if it was not a validation error)
+
+
+#### POST /v1/ajax_login/
+
+Attempt to login, given the parameters `username` and `password`
+
+#### POST /v1/ajax_logout/
```
daniviga: Please add a blank line (other `####` are separated by two blank lines)

openquake/server/settings.py
```diff
@@ -89,7 +89,7 @@
 
 MIDDLEWARE_CLASSES = (
     'django.middleware.common.CommonMiddleware',
-    'django.middleware.csrf.CsrfViewMiddleware',
+    # 'django.middleware.csrf.CsrfViewMiddleware',
```
daniviga: Please add a comment. i.e. `disabled because....`
or
is it possible to keep it on and disable (or decorate) only the affected views?

daniviga:
About the issue with `django.middleware.csrf.CsrfViewMiddleware` please add a comment about why it has been disabled (see above) and we can merge like this for now. As soon as I have a working plugin I can start investigating the issue.

PR #2285, opened 2016-10-07
Storing the source IDs per each task

micheles:
This is useful to investigate the issue of slow tasks.


openquake/calculators/views.py
```diff
@@ -665,6 +665,8 @@ def view_task_info(token, dstore):
 
     data = ['operation-duration mean stddev min max num_tasks'.split()]
     for task in dstore['task_info']:
+        if task == 'source_ids':
```
daniviga: Move this out of `task_info` hdf5 group, so you can avoid the if -> continue

PR #2291, opened 2016-10-11
Remove "reset_is_running" from the wsgi module

micheles:
This should solve the problem that restarting the WebUI marks perfectly fine calculation as failed.


daniviga:
LGTM

PR #2295, opened 2016-10-13
Renamed rup_id->eid

micheles:
For consistency sake.

### DIFF
--- a/openquake/commonlib/export/hazard.py
+++ b/openquake/commonlib/export/hazard.py
@@ -727,7 +727,7 @@ def export_gmf_txt(key, dest, sitecol, imts, ruptures, rlz,
     return {key: [dest]}
 
 
-def get_rup_idx(ebrup, etag):
+def get_eidx(ebrup, etag):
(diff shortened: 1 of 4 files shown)


openquake/commonlib/export/hazard.py
```diff
@@ -727,7 +727,7 @@ def export_gmf_txt(key, dest, sitecol, imts, ruptures, rlz,
     return {key: [dest]}
 
 
-def get_rup_idx(ebrup, etag):
+def get_eidx(ebrup, etag):
```
ptormene: Please make sure you want the name to be `get_eidx` rather than `get_eid`. Furthermore, in this diff I don't see any call to `get_eidx`. Please make sure calls to `get_rup_idx` are all renamed to `get_eidx` (or `get_eid`)

micheles: Thanks Paolo, actually I did not realize that `get_eidx` was not used anymore. I removed it now.

PR #2302, opened 2016-10-17
DO NOT remove /etc/openquake/ but only /etc/openquake/openquake.cfg

daniviga:
By default `apt remove --purge` deletes the whole `/etc/openquake`: THIS IS BAD.

Since `/etc/openquake` is the entire namespace for all OQ releated apps, not just the engine, `/etc/openquake/openquake.cfg` should be removed alone.

This caused the drop of the entire OQ Platform configuration after a removal of `python-oq-engine` (see https://github.com/gem/oq-platform/issues/615)

### DIFF
--- a/debian/postrm
+++ b/debian/postrm
@@ -42,10 +42,11 @@ case "$1" in
     # This package has previously been removed and is now having
     # its configuration purged from the system.
     :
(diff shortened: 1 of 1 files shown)


nastasi-oq:
LGTM

PR #2307, opened 2016-10-19
Fixed a bug when splitting the fault sources by magnitude

micheles:
On Graeme's suggestion, now the splitting procedure produces ArbitraryMFD instances.


g-weatherill:
LGTM

PR #2310, opened 2016-10-19
Add "upgrade" documentation

daniviga:
Easter egg included.

### DIFF
--- a/doc/installing/macos.md
+++ b/doc/installing/macos.md
@@ -34,6 +34,17 @@ chmod +x openquake-setup-macos-2.1.0-1.run
 ```
 then follow the wizard on screen. By default the code is installed in `~/openquake`
 
+### Upgrade from a previous installation
+
+To upgrade from a previous installation you need to manually remove it first
+
+```bash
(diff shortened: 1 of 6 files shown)


doc/installing/macos.md
```diff
@@ -34,6 +34,17 @@ chmod +x openquake-setup-macos-2.1.0-1.run
...
+### Upgrade from a previous installation
+
+To upgrade from a previous installation you need to manually remove it first
+
+```bash
+# default is ~/openquake
+rm -Rf /path/to/previous/installation
+```
+Different versions can be installed alongside if a unique destination path is provided.
+
```
micheles: But running different versions at the same time require changing the port of the DBServer. Not sure if we should mention this, though.

micheles:
LGTM

doc/installing/macos.md
micheles: I would "hide" this information. It is not a good idea to install multiple versions anyway, let's not suggest that this is easy to do.

PR #2311, opened 2016-10-20
Skip UCERF tests with h5py < 2.6.0

daniviga:
A user reported failing tests on RedHat 7. That's partially true since the UCERF calculator tests are failing: the tests are not correctly skipped.
RHEL 7 provides `h5py` 2.3.1 which was satisfying the test requirements, but 2.3 is actually too old for UCERF. I have increased the required `h5py` version to 2.6.0 which is currently our target for UCERF.


micheles:
LGTM

PR #2312, opened 2016-10-21
Introduced a parameter iml_disagg in the job.ini

micheles:
In this way the user can specify the IML she wants to disaggregate upon. If not given for an IMT, it is inferred from the `poes_disagg` parameter, as in the past. Implemented in collaboration with M. Pagani.
This requires https://github.com/gem/oq-hazardlib/pull/520.


preinh:
Great job @micheles !

Just let me add something else here, in terms of my understanding. I have two questions/comments.

One is about this fixed value (eg. {PGA: 0.1}). As implemented, this option attend a more general case (I'll comment later) what looks great, but often this imt fixed value, is no other than 'the corresponding IML, from the 'mean-hazard-curve', to the disagg_poe'. In other words... its a value that can be computed FROM the mean hazard curve at disag_poe (eg, poe 1.e-4). And as you have noticed, when we fix this value, it is 'conflicting' with a 'disag_poe', what looks good, since we get different poes for a fixed iml.

The second comment may highlight one efficiency point. If we are fixing a iml from beginning, still make sense to compute the whole hazard curve, I mean, for many iml values, (eg. intensity_measure_types_and_levels = {IMT: iml1, iml2, .. imlN} ... iml_disag = {IMT: imlX })... if we already know a fixed IMT (from a previous hazard job), to disaggregate for a fixed value could be the same as disaggregate only for one imt level (eg. that one from the mean hazard curve at the target disagg_poe)... Reversing, by fix an imt level for disagg, only the imt levels used to interpolate the hazard curve will finally be used (in the maximum of two)... 

Seems that the understanding of what this 'target disagg poe' means, changes from that poe which be disaggregated at the hazard curves of each individual branch (at different iml) to that poe 'at the mean hazard curve' which fix the iml (taking different poes) to disaggregate each branch.

What do you think about them?! 
Whats about an option like 'compute from mean hazard = true' for this feature. Does it make sense?!

preinh:
@mmpagani and @micheles ,

I apologise if this were not the best place for share with you both that comments. 
Of course I'm more than happy with this update. It solves completely the problem faced. As I said, it solves a more general problem. I let these comments for your further consideration purpose only.

mmpagani:
Hi Marlon, no need to apologise. Thanks for your comments, which were appreciated. Once this PR and the accompanying one are merged into master give a try with these new feature and let us know how it goes.

PR #2316, opened 2016-10-25
Made it possible to export the losses generated by a specific event

micheles:
The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2289


raoanirudh:
@micheles: I think adding this functionality for the `agg_loss_table` won’t provide any benefit to the users. The `agg_loss_table` is small enough that one can always export the full table easily.

On the other hand, this functionality could be quite useful for asset loss tables, where the sheer quantity of data makes it difficult to export the full tables. In this case too, the typical use case I envisage is exporting the asset losses for a specified _set_ of events, rather than for a single event.

micheles:
You have a point, Anirudh. I have removed that functionality and instead added an exporter for the asset loss table. It only works for a single event for the moment. An extension to multiple events can be done later.

PR #2317, opened 2016-10-25
Fix for Ubuntu 14.04

micheles:
Jenkins is broken: https://ci.openquake.org/job/master_oq-engine/2866/console

Until the `opt` distribution is a reality, I suggest to change Jenkins to run Ubuntu 14.04 during the day and Ubuntu 16.04 during the night. Here you can see that the fix works: https://ci.openquake.org/job/zdevel_oq-engine/2280/


daniviga:
Jenkins has been updated

PR #2325, opened 2016-10-27
Compute the epsilons on the workers for asset_correlation=0

micheles:
This is implemented only for  the ebrisk calculator. This may be enough, if the old calculator goes away.
I measured the improvement on a computation for Chile. Here are the figures:

```
memory consumption on the master: 74 GB -> 7 GB
data transfer master->workers: 44 GB -> 2 GB
```


openquake/commonlib/oqvalidation.py
```diff
@@ -213,6 +213,10 @@ def __init__(self, **names_vals):
                     'iml_disagg=%s will not be computed from poes_disagg=%s',
                     str(self.iml_disagg), self.poes_disagg)
 
+        # checks for ebrisk
+        if self.calculation_mode == 'ebrisk' and self.asset_correlation:
+            raise ValueError('assert_correlation != 0 is not supported')
```
ptormene: Typo: `asseRt_correlation`

PR #2328, opened 2016-10-27
Solved the performance problem of the scenario calculators

micheles:
This was reported by several users running calculation from a rupture model. It turns out that all the time and memory were spent re-ordering the GMFs read from the datastore, but there was no need to do so.
The improvement is drastic (from 1 hour to 1 minute in some calculation).
The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2283


micheles:
The ruptures output was exported by accident, I did not realize it. I removed it again. I have also added a test for the case of a a FilteredSiteCollection.

PR #2334, opened 2016-11-02
Used a smaller tolerance when checking the expected files

micheles:
This is needed to have some broken tests pass on Mac OS X.


openquake/calculators/tests/classical_test.py
```diff
@@ -165,6 +165,7 @@ def test_case_14(self):
 
     @attr('qa', 'hazard', 'classical')
     def test_case_15(self):  # full enumeration
+        check_platform('xenial', 'trusty')
```
daniviga: This means that we are skipping this test/these tests on RHEL too. For now it is ok, but we should think about it

micheles:
All tests are green now: https://ci.openquake.org/job/zdevel_mac_oq-engine/28/

openquake/calculators/tests/classical_test.py
micheles: I have moved down the `check_platform` line, so now the CSV are tested on any platform. Only the .geojson and .xml tests are skipped since they rely on the exact numbers (the tolerance feature is implemented only for CSV).

daniviga:
LGTM

Issue #2338, opened 2016-11-04
Improve error message when wrong file is assigned


CatalinaYepes:
If the user indicates a wrong 'logic tree file' (for example, a 'source_model_file' instead of a 'source_model_logic_tree_file') the following error message appears:

 `NameError: No subnode named 'logicTree' found in 'nrml'`

A better message will be:
'The `source_model_logic_tree_file` provided do not have the correct format'

PR #2339, opened 2016-11-05
Improved error message

micheles:
Improve the error when the user gives a source model file instead of a source model logic tree file, as requested by Catalina. Closes https://github.com/gem/oq-engine/issues/2338.


CatalinaYepes:
LGTM

PR #2340, opened 2016-11-07
Replaced the calculator ucerf_event_based with ucerf_rupture

micheles:
The calculator was not setting the seeds correctly, so the result was dependent on the number of tasks generated. Since the calculator is currently NOT used (it was experimental) I am removing it. It will be re-added if needed. For the moment, the only request from the scientists is to have a rupture calculator able to save the seismic events and the `rup-data` information.


ptormene:
LGTM

PR #2345, opened 2016-11-08
Fixed excessive strictness when validating `consequenceFunction.id`

micheles:
Catalina reported this bug:

```
ValueError: node consequenceFunction: Could not convert id->SimpleId(100, ^[\w_\-]+$): Invalid ID 'CR+PCPS/LFM': the only accepted chars are a-zA-Z0-9_-, line 9, line 9 of /home/cyepes/colombia/consequence_model.xml
```

The solution is to relax the check on `consequenceFunction.id`.
Moreover, at Catalina's request, I am changing the default export format from 'xml,csv' to 'csv,xml' i.e. try the CSV export first, the XML export later.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,5 @@
   [Michele Simionato]
+  * Fixed excessive strictness when validating `consequenceFunction.id`
   * Added an `ucerf_rupture` calculator able to store seismic events and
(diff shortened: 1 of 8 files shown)


CatalinaYepes:
LGTM

PR #2347, opened 2016-11-09
Fixed the storage of rup_data

micheles:
I discovered an error when storing the `rup_data` in the ebrisk calculator: the data was stored multiple times in presence of multiple tectonic region types. I am fixing the bug and changing the way `rup_data` is stored. This is also part of https://github.com/gem/oq-engine/issues/2150.
The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2292


openquake/commonlib/export/hazard.py
```diff
@@ -126,6 +126,36 @@ def export_ruptures_xml(ekey, dstore):
     return [dest]
 
 
+@export.add(('rup_data', 'csv'))
```
ptormene: Is it still `rup_data`, or does it become `ses`?

micheles: Good point. There is some confusion here, since this is essentially a view over `rup_data` and `events` that becomes a `ses` output. I am not clear myself of what the name should be. Let's keep `rup_data` for the moment but it could change in the future.

PR #2348, opened 2016-11-09
The source ID were silently truncated to 30 characters

micheles:
I have raised the limit to 60 characters: if a source model contains a source ID longer than the limit a clear error message is displayed, including the number of the line in the source model where the offending ID is.
The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2288


CatalinaYepes:
LGTM

PR #2349, opened 2016-11-09
Remove some old scripts

daniviga:
- Removes a bash script used with very old versions of LXC on CI
- Removes the `oq-engine` wrapper left for backward compatibility


micheles:
LGTM

PR #2352, opened 2016-11-11
Assigned a random year to each event

micheles:
This has been requested by several users. For time independent calculations (i.e. all calculations so far) the year can be assigned randomly. For instance if `investigation_time=50` and there are 3 stochastic event sets, the year will be:

- a random number between 1 and 50 for the first SES;
- a random number between 51 and 100 for the second SES;
- a  random number between 101 and 150 for the third SES.

Extra care is taken so that calculations are reproducible: the seed is fixed and the events are sorted by rupture serial number. The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2295


openquake/calculators/event_based.py
```diff
@@ -241,35 +240,51 @@ def save_ruptures(self, ruptures_by_grp_id):
...
 
     def post_execute(self, result):
         """
         Save the SES collection
         """
+        logging.info('Setting event years')
+        inv_time = int(self.oqparam.investigation_time)
+        numpy.random.seed(self.oqparam.random_seed)
+        for sm in self.datastore['events']:
+            set_random_years(self.datastore['events/' + sm].value, inv_time)
```
raoanirudh: Seems like the years assigned in the function `set_random_years()` do not show up in the datastore on exiting the function - all of the years are zero.

micheles: Good catch. The issue is that I am setting the year only in memory and not in the datastore. I am going to fix that.

openquake/calculators/event_based.py
```diff
@@ -241,35 +240,51 @@ def save_ruptures(self, ruptures_by_grp_id):
...
+        self.datastore.set_nbytes('rup_data')
+
+
+def set_random_years(events, investigation_time):
+    """
+    Sort the `events` array and attach year labels sensitive to the
+    SES ordinal and the investigation time.
+    """
+    events.sort(order='rupserial')
+    years = numpy.random.choice(investigation_time, len(events))
```
raoanirudh: It would be better to have the first year labeled as year 1 rather than as year 0.

PR #2355, opened 2016-11-14
Removed SourceSitesFilter

micheles:
Companion of https://github.com/gem/oq-hazardlib/pull/532


openquake/calculators/disaggregation.py
```diff
@@ -227,15 +217,15 @@ def full_disaggregation(self):
                     if (sm_id, sid) in self.bin_edges:
                         bin_edges[sid] = self.bin_edges[sm_id, sid]
 
-                ss_filter = SourceSitesFilter(oq.maximum_distance)
+                ss_filter = RtreeFilter(sitecol, oq.maximum_distance, False)
```
ptormene: Please use None instead of False also here, for consistency

Issue #2346, opened 2016-11-09
The composite risk model is stored incorrectly (again)


micheles:
This happens when there are slashes in the taxonomy. We should understand why the escaping does not work and fix the test.

micheles:
The right solution is to make CompositeRiskModel serializable to HDF5 by properly escaping the taxonomies.

PR #2358, opened 2016-11-15
Escape the taxonomies when serializing the CompositeRiskModel

micheles:
There is also some refactoring. Now the CompositeRiskModel is fully serializable to and from HDF5.
Closes https://github.com/gem/oq-engine/issues/2346.


openquake/risklib/riskinput.py
```diff
@@ -209,21 +209,87 @@ def build_asset_collection(assets_by_site, time_event=None):
...
+    :returns: a :class:`CompositeRiskModel` instance
+    """
+    oqparam = dstore['oqparam']
+    crm = dstore.getitem('composite_risk_model')
+    rmdict, retrodict = {}, {}
+    for taxo, rm in crm.items():
+        rmdict[taxo] = {}
+        retrodict[taxo] = {}
+        for lt in rm:
+            lt = str(lt)
```
ptormene: Please add a comment

openquake/risklib/riskinput.py
```diff
@@ -22,7 +22,7 @@
 import numpy
 
 from openquake.baselib import hdf5
-from openquake.baselib.python3compat import zip
+from openquake.baselib.python3compat import zip, encode
```
ptormene: Please do not import encode

openquake/risklib/riskinput.py

PR #2365, opened 2016-11-17
rpm: put openquake in /opt and use python-oq-libs

daniviga:
**!! breaks backward compatibility if merged into master !!**

Work in progress.

Needs https://github.com/gem/oq-libs as dependency


nastasi-oq:
LGTM

PR #2373, opened 2016-11-23
Added a `shared_dir` parameter in openquake.cfg

micheles:
The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2302/


openquake.cfg
```diff
@@ -47,6 +47,11 @@ host = localhost
 port = 1999
 authkey = changeme
 
+[directory]
+# the base directory containing the <user>/oqdata directories
```
daniviga: I would add that the default is `<user_home>/oqdata`, otherwise it is `shared_dir/<user>/oqdata`

PR #2374, opened 2016-11-23
Replace the 'oq' entrypoint with a wrapper script

daniviga:
This is needed by the 'opt' version of openquake. For now it does not change anything, everything should work as before.

Tests https://ci.openquake.org/job/zdevel_oq-engine/2304/


micheles:
LGTM

PR #2382, opened 2016-11-28
Rewrite the auth middleware and make compatible with py3

daniviga:
The previous middleware (an adaption of the one coming from the Platform) was not working with Python 3

Tests: https://ci.openquake.org/job/zdevel_oq-engine/2307/

### DIFF
--- a/openquake/server/middleware.py
+++ b/openquake/server/middleware.py
@@ -16,13 +16,14 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with OpenQuake. If not, see <http://www.gnu.org/licenses/>.
 
+import re
+
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


openquake/server/middleware.py
```diff
@@ -17,33 +17,33 @@
 # along with OpenQuake. If not, see <http://www.gnu.org/licenses/>.
 
 from django.conf import settings
-from django.core.urlresolvers import reverse
 from django.http import HttpResponseRedirect
 from re import compile
```
micheles: ` from re import compile` shadows a Python builtin called `compile`. Better to import `re` and then use `re.compile`

micheles:
LGTM

PR #2393, opened 2016-12-02
Added a command `oq to_hdf5` to convert .npz files into .hdf5 files

micheles:
While there are libraries to read .npy files from other languages, they are not as mature as the HDF5 libraries. Fortunately it is rather trivial to convert .npz files to .hdf5 files. In particular, the conversion make it possible to explore the outputs with `hdfview`.


micheles:
For the moment this is just a convenience, I will add tests when we will have official specifications for the HDF5 outputs.

PR #2394, opened 2016-12-02
Merged the tests case_master and case_master_ebr

micheles:
Since now there is a single calculator it makes sense to have a single test. Since I have changed
the investigation_time the numbers are different. I have also committed the expected loss maps for each realization.


micheles:
LGTM since there are no code changes

PR #2397, opened 2016-12-05
Force UTF-8 encoding in supervisord

daniviga:
By default supervisord starts daemons with `LANG=C` causing troubles when a job that contains `UTF-8` characters is uploaded through the API/QGIS

Demos: https://ci.openquake.org/job/zdevel_oq-engine/2324/


micheles:
LGTM

PR #2402, opened 2016-12-12
Fixed UHS export bug

micheles:
One of our users (Céline Beauval) discovered that the engine was giving bogus numbers for the uniform hazard spectra. Marco Pagani correctly noticed that this was due to a mismatch with non-spectral acceleration IMTs. Actually we had a function to filter out such IMTs but it was called too late.
Also there was not a single test sensitive to the issue. Here I am adding the test and fixing the issue,
which affects only the exporter. The fix amounts to two lines in calc.py.


daniviga:
Does this should be backported to 2.1?

mmpagani:
LGTM

micheles:
The bug was already there in engine 2.0, perhaps even before. It is not very urgent because it only happens when there is a non-SA intensity measure type and it has an easy workaround (remove the offending IMT). Let's not make a new release only for that.

PR #2410, opened 2016-12-15
Support for mutually exclusive probabilities in the source model parser/writer

micheles:
Reflect the changes in hazardlib implemented in https://github.com/gem/oq-hazardlib/pull/500


mmpagani:
LGTM

PR #2411, opened 2016-12-15
Mark the binary dist as pure, universal python

daniviga:
See: https://wheel.readthedocs.io/en/latest/#defining-the-python-version


micheles:
LGTM

PR #2412, opened 2016-12-16
Wrong order of operations in .travis.yml

daniviga:
`python setup.py develop` is useless since we are using PYTHONPATH and we are on pure python (no extensions). This was also made before hazardlib was actually available causing the download of a wrong version from pypi (now removed)


micheles:
LGTM

PR #2435, opened 2017-01-10
Fix build timeout in Jenkins when tests fail

daniviga:
This was caused by the `oq dbserver start` command which is run in background leaving the ssh session active even if `set -e` and an exit code =1 from nose.

Remove also some old code from the `packager.sh`

With an error: https://ci.openquake.org/job/zdevel_oq-engine/2345/console
Clean test: https://ci.openquake.org/job/zdevel_oq-engine/2347/

### DIFF
--- a/packager.sh
+++ b/packager.sh
@@ -59,10 +59,10 @@ GEM_MAXLOOP=20
 GEM_ALWAYS_YES=false
 
 if [ "$GEM_EPHEM_CMD" = "" ]; then
-    GEM_EPHEM_CMD="lxc-start-ephemeral"
+    GEM_EPHEM_CMD="lxc-copy"
 fi
 if [ "$GEM_EPHEM_NAME" = "" ]; then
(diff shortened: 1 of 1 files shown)


packager.sh
```diff
@@ -329,14 +311,18 @@ _devtest_innervm_run () {
             skip_tests="!slow,"
         fi
 
-        # run tests (in this case we omit 'set -e' to be able to read all tests outputs)
+        ssh $lxc_ip "set -e
+                 export PYTHONPATH=\"\$PWD/oq-hazardlib:\$PWD/oq-engine\"
+                 echo 'Starting DbServer. Log is saved to /tmp/dbserver.log'
+                 cd oq-engine; nohup bin/oq dbserver start &>/tmp/dbserver.log < /dev/null &"
```
nastasi-oq: ``</dev/null`` instead of ``< /dev/null``

nastasi-oq:
Except a minor comment LGTM.

PR #2436, opened 2017-01-10
Fix an if in packager.sh: demos were run even if not asked to

daniviga:
A bug was introduced in `packager.sh` (when?): even if on CI `run_pkg_demos` was __off__.

See for example: https://ci.openquake.org/job/zdevel_oq-engine/2351/parameters/ the computation was stuck because CI was trying to run the demos but celery was not started because `run_pkg_demos` was not set. It shouldn't had run any demo.

Fixed run: https://ci.openquake.org/job/zdevel_oq-engine/2354

### DIFF
--- a/packager.sh
+++ b/packager.sh
@@ -547,7 +547,6 @@ celeryd_wait() {
 }
 
 celeryd_wait $GEM_MAXLOOP"
-        fi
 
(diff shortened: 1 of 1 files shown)


micheles:
LGTM

PR #2437, opened 2017-01-10
Fix errors with Django 1.9

daniviga:
Fixes master on macOS
Django 1.10 still needs more work

Tests: https://ci.openquake.org/job/macos/job/zdevel_mac_oq-engine/79/

### DIFF
--- a/doc/requirements.md
+++ b/doc/requirements.md
@@ -26,4 +26,4 @@ HDF5 | 1.8
 
 ### Python dependencies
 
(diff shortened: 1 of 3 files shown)


micheles:
I see that the range is  'django >=1.6, <1.10', while Ubuntu 16 is using Django 1.8.7.
Why do we need to go up to 1.10?
Apart from that, LGTM

daniviga:
'Cause support for 1.8 is going to be stopped (speaking about the one downloaded from pypi, the one from deb/rpm packages should has fixes backported).

PR #2441, opened 2017-01-11
Port WebUI to Django 1.9 and 1.10

daniviga:
Keeps backward compatibility with any Django >= 1.6

Some manual testing would be great.

https://ci.openquake.org/job/zdevel_oq-engine/2355/ (could be red, because of https://github.com/gem/oq-engine/pull/2435, look only at server tests)
https://ci.openquake.org/view/macOS/job/macos/job/zdevel_mac_oq-engine/80/

### DIFF
--- a/openquake/server/urls.py
+++ b/openquake/server/urls.py
@@ -17,33 +17,34 @@
 # along with OpenQuake. If not, see <http://www.gnu.org/licenses/>.
 
 from django.conf import settings
-from django.conf.urls import patterns, url, include
+from django.conf.urls import url, include
 from django.views.generic.base import RedirectView
+from django.contrib.auth.views import login, logout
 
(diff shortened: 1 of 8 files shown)


openquake/server/urls.py
```diff
@@ -17,33 +17,34 @@
...
 from django.conf import settings
-from django.conf.urls import patterns, url, include
+from django.conf.urls import url, include
 from django.views.generic.base import RedirectView
+from django.contrib.auth.views import login, logout
 
-urlpatterns = patterns(
-    '',
+from openquake.server import views as oqserver
+
```
micheles: Why we are renaming views as oqserver? Can you we just keep it as it is?

micheles:
Apart the minor comment, LGTM

PR #2442, opened 2017-01-11
Update classifiers in setup.py

### LINKED ISSUE #2438: Classifiers need to be updated
Before we push code to pypi we have to update classifiers in `setup.py`: https://github.com/gem/oq-engine/blob/master/setup.py#L94

- Mark as stable
- Add support flag for python3

daniviga:
Closes #2438


micheles:
LGTM

PR #2448, opened 2017-01-12
Add matplotlib to extras

### LINKED ISSUE #2447: Make matplotlib an optional dependency of the engine
Right now it is used only in debugging, so it is not mandatory.

daniviga:
And allow usage of Rtree==0.8.3

Closes #2447 

Now, by default, "extra" deps are installed when `pip install -r requirements-py[27,35]-linux64.txt` is used on Linux.


micheles:
LGTM

PR #2450, opened 2017-01-13
Load django auth only if needed to avoid issues during packaging

daniviga:
This avoids run 'migrate' or 'syncdb' before using the WebUI: `manage.py` requires the dbserver to be on and this could be a problem during packaging (see https://github.com/gem/oq-engine/issues/2451).

Tests: https://ci.openquake.org/job/zdevel_oq-engine/2361/


micheles:
LGTM

PR #2452, opened 2017-01-13
Call the db server in manage.py only for runserver

### LINKED ISSUE #2451: python manager.py should not rely on dbserver
`python manage.py` should require an active dbserver only for `runserver` command.

Currently it requires the dbserver to be alive for _any_ command (including `createsuperuser`). Some of these commands are useful during maintenance/installation when the dbserver cannot be started.

daniviga:
Fixes #2451


micheles:
LGTM

PR #2460, opened 2017-01-17
Cython isn't needed, get rid of it

daniviga:
In setup and requirements we had Cython, but it isn't used anymore and installed by binary packages.

This was causing, for example, compilation of cython during installation of those apps that have `openquake.engine` in their requirements and are installed on a system which uses only binary packages.

(an example: https://ci.openquake.org/job/zdevel_oq-platform-standalone/63/console)

Companion:

- https://github.com/gem/oq-hazardlib/pull/564


micheles:
Why this was causing compilation of cython? The Cython wheel is already compiled, isn't it?

daniviga:
The problem is like this:

on Trusty we have installed the OQ Ubuntu packages (no cython installed from deb), then we installed `oq-platform-standalone` which requires `openquake.engine`, which brings `cython` because the `setup.py`.
Trusty's pip is too old to use Cython wheel by default, so a `tar.gz` is downloaded and installed instead, increasing the build time.

daniviga:
We can also decide to keep Cython in our dependencies for future needs, but in these case, for coherency it must be added to `debian/control` and `rpm spec`

micheles:
Let's remove it and re-add in the future only if we will use it. LGTM

PR #2463, opened 2017-01-17
Add requirements with local wheels copy for macOS

daniviga:
As we do for linux, this is our reference.
Python 2.7 only since currently Apple does not ship Python 3 by default.

Companion on hazardlib https://github.com/gem/oq-hazardlib/pull/565


micheles:
LGTM

PR #2469, opened 2017-01-19
Used logscale in the AreaSource demo

micheles:
This should produce better graphs for the Uniform Hazard Spectra, as discussed with @mmpagani and @ptormene . Companion of https://github.com/gem/oq-hazardlib/pull/566 .


mmpagani:
Fantastic! basically it is already supported. Does this mean that we can also use a linscale?
LGTM

PR #2473, opened 2017-01-20
Reflected changes in the sourcewriter

micheles:
Companion of https://github.com/gem/oq-hazardlib/pull/571


mmpagani:
The testing we have for this part is quite limited. We will need to work on expanding it considerably. 
For the time being I think it's fine.
LGTM

PR #2479, opened 2017-01-23
Use a differnt port, with a better reputation for dbserver

daniviga:
Currently it's `1999` which is used by some Cisco network equipment and has then a bad reputation (see http://www.speedguide.net/port.php?port=1999), so firewall are locking it even on the `loopback` interface. This PR changes the default port to `1908` (the [Great Messina earthquake](http://emidius.mi.ingv.it/CPTI15-DBMI15/eq/19081228_0420_000)) which has a far better reputation (http://www.speedguide.net/port.php?port=1908).


micheles:
So that was the reason for the issues on some Windows machine! LGTM

daniviga:
Not sure, could be. Anyway `1908` is safest. I will add a comment too.

PR #2483, opened 2017-01-24
New logic for openquake.cfg file location

daniviga:
https://ci.openquake.org/job/zdevel_oq-engine/2373/
https://ci.openquake.org/job/builders/job/redhat-builder/60/


micheles:
Notice that I would not mind removing `utils_config_test.py`, since the configuration file logic is tested in the unix/windows/mac installers. Let's think if they are worth maintaining or not.

openquake/commonlib/config.py
```diff
@@ -68,17 +78,50 @@ def set(self, name, dic):
...
+
+        local_path = os.path.join(local_prefix, self.CFG_PREFIX,
+                                  self.CFG_FILE)
+
+        paths = [local_path]
+
+        # when not running in a virtualenv system config
+        # has the highest priority
+        if not hasattr(sys, 'real_prefix'):
+            paths.append(os.path.join(self.SYS_PREFIX, self.CFG_PREFIX,
```
micheles: It seems that 'real_prefix' does not work for people using `python3 -m venv` instead of virtualenv, see the discussion in http://stackoverflow.com/questions/1871549/python-determine-if-running-inside-virtualenv

PR #2487, opened 2017-01-25
Support for negative depths

micheles:
Companion of https://github.com/gem/oq-hazardlib/pull/573/
A test is missing, it should be provided by @rcgee 

The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2382/


micheles:
We need to run a large computation and compare the data transfer with master before merging it.

PR #2489, opened 2017-01-25
Put openquake.cfg in openquake/engine/ python package

### LINKED ISSUE #2478: Put openquake.cfg in a proper place when installed via pip
Currently `oq` looks for `openquake.cfg` in `/etc/openquake` and in `site-packages`. This does not play well with installations from pip because:

1. It is not part of the python package 
2. It cannot be installed in `/etc`
3. `site-packages` folder is the wrong place (out of the openquake context)

To work around this issue in various installers we are setting the `OQ_SITE_CFG_PATH` env var (see https://github.com/gem/oq-installers/blob/master/unix/env/install.sh#L75 and https://github.com/gem/oq-installers/blob/master/windows/nsis-installer/oq-console.bat#L6)

daniviga:
`openquake.cfg` has been moved to `openquake/engine/openquake.cfg`, inside the python package and is now installed by setuptools/pip.

The follow logic is used to determine which `openquake.cfg` must be used:

1. if the `OQ_CONFIG_FILE` environment variable is set, take the config from there
if not
2. use `openquake.cfg` from `/etc/openquake` when not running inside a venv (tested with `python3 -m venv` and `virtualenv`)
if not
3. use the `openquake.cfg` provided by the python package

**Please note**: `OQ_LOCAL_CFG_PATH` and `OQ_SITE_CFG_PATH` vars have been removed

Replaces #2483 
Closes #2478 

https://ci.openquake.org/job/zdevel_oq-engine/2377/
https://ci.openquake.org/job/macos/job/zdevel_mac_oq-engine/86/
https://ci.openquake.org/job/windows/job/zdevel_oq-engine/2/console


daniviga:
Now it works pretty well in almost any condition (tested using pip, venv over packages, using env var). Please do some manual testing. I still need to fix/rewrite/drop/nuke old tests: I want to spend a bit of time on them because I don't like how they look like now.

openquake/commonlib/config.py
```diff
@@ -68,16 +68,24 @@ def set(self, name, dic):
...
+
+        # path from python package
+        paths.append(os.path.join(OQ_PATH, "engine", self.CFG_FILE))
+
+        # path from system etc dir, only if a venv is not active
+        venv = 'VIRTUAL_ENV' in os.environ or hasattr(sys, 'real_prefix')
+        if not venv:
+            sys_path = self.ETC_PATH
+            if os.path.exists(sys_path):
+                paths.append(sys_path)
```
micheles: Just use self.ETC_PATH

micheles:

PR #2495, opened 2017-01-27
Add prctl to optional deps

daniviga:
And include its wheel (a `manylinux1` with embedded `libcap` made by us) in requirements for linux.

We still have the issue with orphan children when a calculation is stopped with `CTRL-C` or `SIGTERM`; adding `prctl` mitigates the issue.


micheles:
LGTM

PR #2500, opened 2017-01-30
Added command `oq reset`

micheles:
This command is meant only for single user situations.While at it, I have also fixed a bug with `raw_input` in Python 3.

### DIFF
--- a/openquake/commands/reset.py
+++ b/openquake/commands/reset.py
@@ -1,0 +1,60 @@
+# -*- coding: utf-8 -*-
+# vim: tabstop=4 shiftwidth=4 softtabstop=4
+#
+# Copyright (C) 2015-2017 GEM Foundation
+#
+# OpenQuake is free software: you can redistribute it and/or modify it
+# under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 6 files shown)


daniviga:
I tried to use `oq reset` but I've got:

```
Do you really want to destroy the database and all datastores? (y/n) y
dbserver stopped
Removed /home/daniele/oqdata
You have no permission to remove ~/oqdata/db.sqlite3
```
but actually `oqdata` has been removed

openquake/commands/reset.py
```diff
@@ -0,0 +1,58 @@
...
+    dbpath = config.get('dbserver', 'file')
+    datadir = os.path.realpath(datastore.DATADIR)
+    if os.path.exists(datadir):
+        shutil.rmtree(datadir)  # datastore of the current user
+    print('Removed %s' % datadir)
+    try:
+        os.remove(dbpath)  # database of the current user
+        print('Removed %s' % dbpath)
+    except OSError:
+        print('You have no permission to remove %s' % dbpath)
```
daniviga: I would remove first the DB and then the datadir,  otherwise when using the default configuration the DB is deleted by the commands above and this will fail

micheles:
I am removing the db before oqdata now.

daniviga:
I'm running a test on windows here: https://ci.openquake.org/job/windows/job/zdevel_oq-engine/3/

Issue #2476, opened 2017-01-23
Include output file to specify tectonic region type


CatalinaYepes:
Currently outputs from Event based calculator are dived per tectonic region type, with a corresponding `grp_id`. It is necessary to provide information about the corresponding tectonic region type for each `grp_id`

raoanirudh:
The new aggregate loss tables already list the tectonic region type. It would be useful to provide the same column also in the stochastic event set (eg: #2150).

micheles:
There is an easy way to get the tectonic region type, i.e the command `show ruptures_per_trt`,
Here is an example:

```bash
$ oq show ruptures_per_trt <calc_id>`

====================== ====== ==================== =========== ============ ============
source_model           grp_id trt                  num_sources eff_ruptures tot_ruptures
====================== ====== ==================== =========== ============ ============
as_model_emme_Iran.xml 0      Subduction Inslab    3           89           393,276     
as_model_emme_Iran.xml 1      Subduction Interface 3           97           36,774      
as_model_emme_Iran.xml 2      Deep Seismicity      16          152          687,247     
as_model_emme_Iran.xml 3      Stable Shallow Crust 30          160          963,941     
as_model_emme_Iran.xml 4      Active Shallow Crust 309         3448         3,397,401   
fs_model_emme_Iran.xml 5      Subduction Inslab    3           95           393,276     
fs_model_emme_Iran.xml 6      Deep Seismicity      16          146          687,247     
fs_model_emme_Iran.xml 7      Subduction Interface 18          300          38,786      
fs_model_emme_Iran.xml 8      Stable Shallow Crust 7,537       92           247,792     
fs_model_emme_Iran.xml 9      Active Shallow Crust 38,117      2504         1,921,509   
====================== ====== ==================== =========== ============ ============
```

From the rupture tag one can extract the source group ID and then the tectonic region type. For instance the tag `grp=09~ses=0001~src=active_546~rup=8604456-01` corresponds to the group 9 which is an Active Shallow Crust.

CatalinaYepes:
Please keep in mind that many users don't use the command line to run the engine. If the webui is used, then there is not possibility to obtain this info.

PR #2501, opened 2017-01-31
Export information about the source groups

micheles:
Exports a CSV called `sourcegroups_CALCID.csv` with fields `grp_id` (the source group ID), `trt` (the tectonic region type) and `eff_ruptures` (the effective number of ruptures contained in the source group).
Closes https://github.com/gem/oq-engine/issues/2476.


CatalinaYepes:
LGTM

PR #2502, opened 2017-01-31
Do not save boundary info into `rup_data` for ucerf event based

micheles:
As discovered by Anirudh, the  datastore files for UCERF event based calculations are a lot larger than before (for instance from ~2 GB to ~8 GB in typical calculations for California). The reason is that in engine 2.2 we added information to the event loss table exporter (i.e. magnitude and centroid for each rupture). Such information is stored in the `rup_data` table; keeping such information takes a lot of space.
Actually, the `rup_data` also contains a `boundary` string which is pretty large and it is not needed by the event loss table exporter. This PR strips that information (only for UCERF) and reduces the data transfer and data storage. Here are some numbers:

```
    [#13044 INFO] Received 5.67 GB of data # master
    /home/openquake/michele/oqdata/calc_13044.hdf5 : 8.67 GB

    [#13045 INFO] Received 2.38 GB of data  # this branch
    /home/openquake/michele/oqdata/calc_13045.hdf5 : 5.34 GB
```

We could remove `rup_data` entirely at the cost of removing information from the event loss table exporter (i.e. centroid_lon, centroid_lat, centroid_depth, magnitude would became numpy.nan fields). Not sure if we want to go that way, though.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,5 @@
   [Michele Simionato]
+  * Reduced the data transfer back and disk space for UCERF event based risk
   * Tasks meant to be used with a shared directory are now marked with a
     boolean attribute `.shared_dir_on`
   * Added a warning when running event based risk calculations with sampling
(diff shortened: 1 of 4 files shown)


micheles:
I think the major contributor is the table rup_data.

PR #2503, opened 2017-01-31
Fixed combine_pmaps for reduced logic trees

micheles:
Fixes a bug discovered by Elisabeth, happening when some source groups have probability of exceedence zero, i.e. empty probability maps.


micheles:
I should add a test

PR #2507, opened 2017-02-02
Fix RabbitMQ user and vhost creation, update openquake.cfg

### LINKED ISSUE #2505: RabbitMQ on Xenial does not allow 'guest'
RabbitMQ from Ubuntu 16.04 Xenial does not allow 'guest' user to connect from remote hosts anymore (but only from `localhost`).

This issue has not been caught by our tests since yes, we are running some demos with celery, but only on a single node configuration; our cluster, even if it runs Xenial's python stack it's still using Ubuntu 14.04 Trusty and RabbitMQ from it.

We already have support for a custom user in Ubuntu packages installation (https://github.com/gem/oq-engine/blob/master/debian/postinst#L34), but I'm not sure that it's working correctly and anyway we need to update the `openquake.cfg` to use them.

Thanks to Kyle from FMGlobal for being reporting the issue.

daniviga:
This PR fixes #2505, an issue discovered by FMGlobal on their cluster.

On Xenial, RabbitMQ does not allow `guest` connection from the network. This PR makes these changes:

1. The `pid` of RabbitMQ has a new path, so the old code for creating host and users was not working on Xenial and CentOS 7
2. The `openquake.cfg` now uses these credentials. RabbitMQ + Celery are normally used only through packages, so no manual steps (creating user/vhost) are usually needed; anyway this info has been added to the documentation
3. The vhost/user creation has been added to the RPM package too

Now any issue with user/vhost creation or with wrong settings in `openquake.cfg` are caught by Jenkins since if something is wrong the demos will fail with error `error: [Errno 104] Connection reset by peer` (even if demos are run on a single node).

Marked as critical because it actually prevents a working installation on a cluster with current LTS.

https://ci.openquake.org/job/zdevel_oq-engine/2381/

### DIFF
--- a/doc/installing/cluster.md
+++ b/doc/installing/cluster.md
@@ -32,9 +32,9 @@ On all worker nodes, the `/etc/openquake/openquake.cfg` file should be also modi
 [amqp]
 host = w.x.y.z
 port = 5672
-user = guest
-password = guest
-vhost = /
+user = openquake
+password = openquake
+vhost = /openquake
 # This is where tasks will be enqueued.
 celery_queue = celery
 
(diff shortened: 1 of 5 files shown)


doc/installing/cluster.md
```diff
@@ -32,9 +32,9 @@ On all worker nodes, the `/etc/openquake/openquake.cfg` file should be also modi
 [amqp]
 host = w.x.y.z
 port = 5672
-user = guest
-password = guest
-vhost = /
+user = openquake
+password = openquake
+vhost = /openquake
```
daniviga: Remove the `/`

nastasi-oq:
LGTM

Issue #2512, opened 2017-02-02
Put PDF manual inside packages


daniviga:
Now the OQ manual sources are stored inside this repo (#2511). 
The 'final' PDF is generated by Jenkins, currently by a standalone job. PDF generation has lot of dependencies and requires Xenial.

We need to find a proper way to include automatically the generated pdf (which cannot be included in the GIT repo due size and frequent changes) in deb, rpm and other installers.

daniviga:
Should we put the PDF generated by the builder in a structure inside jenkins

daniviga:
Now the builder job puts the pdf in `http://packages.openquake.org/pdf/oq-engine/$GIT_ID/` where the packager can grab it

daniviga:
We should fix #2636 first

daniviga:
For now we'll keep the manuals only in the standalone installers.

PR #2511, opened 2017-02-02
Add OQ manual files

daniviga:
from https://github.com/GEMScienceTools/oq-engine-docs

1. History has NOT been moved, just plain files to keep the repo size under control (6.9MB vs 600MB)
2. The cover is now converted in PDF fom SVG by Inkscape in the `build.sh` scrip
3. Cover has been reduced in size from 8MB to ~1MB reducing pictures to 300dpi and using jpg compression (high quality: 90% and a 4:4:4 matrix)
4. The builder on CI (which uses Docker: https://github.com/gem/oq-containers/blob/master/jenkins/Dockerfile.manual) has been updated.
5. I want to keep the generation separated from the master oq-engine job on CI for now

See https://ci.openquake.org/job/builders/job/doc-builder/7/ and https://ci.openquake.org/job/builders/job/doc-builder/7/artifact/oq-engine/doc/manual/oq-manual.pdf

Next step could be set automatically the version, taking it from `__init__.py`

PR #2513, opened 2017-02-02
Change what is include in deb docs

daniviga:
In the deb package we are including the sphinx and latex sources (after #2511) which are almost unreadable and useless to final users (as sources), instead we were missing the README, CONTRIBUTORS and LICENSE.

My proposal is to include just the *.md files and in the future we'll add the manual in the form of a pdf (see #2512 ) and eventually the html generated by sphinx.

https://ci.openquake.org/job/zdevel_oq-engine/2385/

### DIFF
--- a/debian/docs
+++ b/debian/docs
@@ -1,1 +1,8 @@
-doc/*
+README.md
+CONTRIBUTORS.txt
+LICENSE
+doc/*.md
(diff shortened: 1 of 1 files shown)


nastasi-oq:
LGTM

PR #2515, opened 2017-02-03
Reduced the size of the event tags

micheles:
Continuation of the work done in https://github.com/gem/oq-engine/pull/2504 and https://github.com/gem/oq-engine/pull/2502. Here is the size of the datastore for the same UCERF calculation:

    # engine 2.2
    /home/openquake/michele/oqdata/calc_13044.hdf5 : 8.67 GB

    # removing the boundaries in rup_data
    /home/openquake/michele/oqdata/calc_13045.hdf5 : 5.34 GB

    # removing rup_data fully
    /home/openquake/michele/oqdata/calc_13051.hdf5 : 4.14 GB

    # stripping the source ids in the event tags
    /home/openquake/michele/oqdata/calc_13062.hdf5 : 0.98 GB

Roughly an order of magnitude of disk space has been saved compared to engine 2.2.
Notice that I am stripping the source IDs in *all* calculators, not only in UCERF. This is safe since they were not used by the calculators, they were just extra information.

The manual was built here: https://ci.openquake.org/job/builders/job/doc-builder/8


micheles:
Yes, I deleted them on purpose. I discovered that a lot of files were not used by tests, they were a legacy from the past.

Issue #2387, opened 2016-11-29
ucerf_classical does not work with sampling


micheles:
Graeme got an error

```python
  File "/usr/local/openquake/oq-engine/openquake/calculators/ucerf_classical.py", line 389, in pre_execute
    self.rlzs_assoc = self.csm.info.get_rlzs_assoc()
  File "/usr/local/openquake/oq-engine/openquake/commonlib/source.py", line 466, in get_rlzs_assoc
    assoc._init()
  File "/usr/local/openquake/oq-engine/openquake/commonlib/source.py", line 213, in _init
    len(self.realizations), self.num_samples)
AssertionError: (1440, 256)
````

in the file `job_classical_full.ini`.

After analysis, it turns out that the calculators does not work even without sampling if there is more than one branch. A test is missing :-(

PR #2516, opened 2017-02-03
Fixed sampling in UCERF

micheles:
Closes https://github.com/gem/oq-engine/issues/2387 and a bug with the `report.rst`. A test for the case of sampling has been added. Code triplication in `build_idx_set` has been removed. Now Graeme's calculation runs but there is no guarantee that the numbers are correct. The calculator is still highly experimental.


openquake/calculators/ucerf_event_based.py
```diff
@@ -856,7 +847,7 @@ class List(list):
 def compute_losses(ssm, sitecol, assetcol, riskmodel,
                    imts, trunc_level, correl_model, min_iml, monitor):
     """
-    Compute the losses for a single source model.
+    Compute the losses for a single source model and returns the ruptures.
```
ptormene: Please check if the :returns: description is still right

micheles: I have made the description clearer

PR #2520, opened 2017-02-06
Refactoring of the UCERF calculators

micheles:
Some duplication in `ucerf_poe_map` has been removed. Ucerf sources are now closer to regular sources and have an `.iter_ruptures()` method.


g-weatherill:
In so far as I'm understanding the logic (which is not really that far at all), LGTM

PR #2531, opened 2017-02-07
Unified UCERF time-dependent and time-independent classes

micheles:
There was no need to differentiate, since the only difference is in the `start_date` attribute.


g-weatherill:
LGTM

PR #2533, opened 2017-02-07
Fixed a bug discovered by Mabe

micheles:
This bug was extremely easy to fix, but extremely to put under test. It was outside of test coverage because it only happens for scenario_risk if there are multiple assets of the same taxonomy on the same site, but there are no insurance losses. I have changed test case_5 to exercise this situation, so that now we are covered.
The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2392/


raoanirudh:
What was the bug?

micheles:
The bug was in the shapes of the array. We were trying to fit an array of shape (2, 1) inside an array of shape (2,), being 2 the number of assets on the same site.

Issue #2527, opened 2017-02-06
Remove the two-phase constraint on the event based calculators


micheles:
Right now the event based calculators have a serious architectural constraint: they are split in two phases, 1) rupture generation and 2) gmf generation. This architecture causes the following issues:

a) we may run out of memory when collecting the whole set of ruptures at the end of phase 1)
b) the calculation may become ultra-slow when saving and then reading the ruptures between phases 1) and 2)
c) rabbitmq may fail while transferring the ruptures at the beginning of phase 2)

The solution is to generate the ruptures and use them within the same task (single phase calculation). This is the way the UCERF risk calculator works. However, the present UCERF risk calculator parallelizes by source model branch: this means that when there is a single branch it will use a single core, i.e. it is practically useless.

Why we have the two-phase architecture? Because we lack a *provider of unique event IDs*. Currently in phase 1) we generate bogus event IDs which are replaced by sequential event IDs only after, when the events are stored. In order to avoid storing the events or keeping them in memory all at once, we need to set the event IDs before, in the workers. An easy solution is to have the task number inside the ID (for instance the first two bytes could be the `task_no`).

Incidentally, by having a provider of event IDs it will become possible to split the UCERF sources in ruptures and to parallelize well even in the case of a single source model branch.

micheles:
I am closing this ticket because things have changed a lot since February.

1. Nowadays ruptures as stored as numpy arrays very efficiently, so saving and reading lots of them is not a problem, i.e. issue b) has disappeared
2. Ruptures are now saved incrementally, so issue a) has disappeared
3. We might even read them directly from the workers by using the two-datastores approach, thus avoiding issue c) altogether.

The important thing was having a way to specify unique event IDs and this has been implemented. End of the story.

PR #2535, opened 2017-02-07
Generate unique event IDs inside the event based rupture calculator

micheles:
See the discussion in https://github.com/gem/oq-engine/issues/2527. One big advantage of this approach is that the IDs are reproducible once `concurrent_tasks` is set (i.e. if you run the same computation twice you get the same event IDs).

The algorithm used to build the event ID has a limit of  65,536 tasks, and each task can generate at most 65,536 ruptures. All known event based calculations are well below such limits; if in the future we will pass the limits, it will be trivial to raise them (change the `eid` from a 32 bit unsigned integer to a 64 bit unsigned integer: then the limits will be 4,294,967,296 tasks each one producing at most  4,294,967,296 ruptures).

The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2390/


openquake/calculators/event_based.py
```diff
@@ -144,9 +166,8 @@ def build_eb_ruptures(
                 num_occ_by_rup[rup].items()):
             for occ_no in range(1, num_occ + 1):
                 # NB: the eid below is a placeholder; the right eid will be
-                # set later, in EventBasedRuptureCalculator.post_execute
-                events.append((eid, ses_idx, occ_no, sampleid))
-                eid += 1
+                # set a but later, in compute_ruptures
```
daniviga: but -> bit?

openquake/commonlib/logs.py
```diff
@@ -15,17 +15,18 @@
...
 # along with OpenQuake. If not, see <http://www.gnu.org/licenses/>.
-
-
 """
 Set up some system-wide loggers
 """
 
 import os.path
 import logging
+import multiprocessing
```
daniviga: Why are you importing multiprocessing which is not directly used?

openquake/commonlib/logs.py
```diff
@@ -15,17 +15,18 @@
...
 from datetime import datetime
 from contextlib import contextmanager
 from multiprocessing.connection import Client
+import numpy
```
daniviga: idem (and in any case I would group it with 'imports' before 'froms')

openquake/commonlib/logs.py
micheles: It's a leftover from a past implementation, I am removing it

openquake/calculators/event_based.py
micheles: yes, typo

PR #2537, opened 2017-02-08
Fixed a serious bug in the UCERF time dependent calculator

micheles:
Inside `ucerf_classical_hazard_by_branch`, the `branch_id` attribute was not set and the method `.build_idx_set()` was not called, so in practice we were using the same `idx_set` for all branches: the produced hazard curves were always the same for all branches.


g-weatherill:
LGTM

Issue #2538, opened 2017-02-08
Decouple the UCERFControl and UCERFSource objects


micheles:
Right now the UCERF calculators use a UCERFControl class which is somewhat tricky: it plays the role of a source, but it also plays the role of a collection of sources, when a branch_id is not specified. Originally there is a single UCERFControl object which unspecified `branch_id`; during the calculation the `branch_id` is specified and the UCERFControl takes the role of a source. The problem is that  if somebody forgets to specify in the right place the `branch_id` and to call `.build_idx_set()`, the wrong data is taken. This is not theoretical, since it caused a real bug: https://github.com/gem/oq-engine/pull/2537. Therefore I propose to use a cleaner design.

On one side, there is a single UCERFControl object, associated to the input file, without the concept of `branch_id` and without the concept of `idx_set`, which cannot be used by the calculators. On the other side, from the UCERFControl one can generate a set of distinct UCERFSource object, one per branch, each one with its one independent `idx_set`. The design if functional: `branch_id` and `idx_set` are fixed at instantiation time in UCERFSource and never change. This is more robust and actually correct: I discovered the bug https://github.com/gem/oq-engine/pull/2537 while implementing this cleaner design.

PR #2539, opened 2017-02-08
Introduced an UcerfSource class

micheles:
Implements https://github.com/gem/oq-engine/issues/2538. I have also improved `test_ruptures`: now the ruptures are checked, before we were purely checking the total number of ruptures.


g-weatherill:
LGTM

PR #2540, opened 2017-02-09
Changed the UCERF classical calculators to run one branch at the time

micheles:
Then each branch is parallelised by ruptures. This allows a big simplification and the function `ucerf_classical_hazard_by_branch` becomes redundant.


g-weatherill:
LGTM

PR #2543, opened 2017-02-09
Correct date of Tusa and Langer GMPE

rcgee:
Companion of gem/oq-hazardlib#588

The year of this GMPE needs to be corrected. The article describing this GMPE was available online in 2015, but the official publication date was actually 2016, when the article was printed.


micheles:
LGTM

PR #2544, opened 2017-02-09
Fixed the event based calculator so that it can run UCERF ruptures

micheles:
Now computing the GMFs is possible, but one can easily run out memory for large calculations, unless an appropriate minimal_intensity is set. I have extended the event based test to compute the GMFs from the ruptures, convert to hazard curves, generate the mean hazard maps for poes=0.02 0.01 and check the result against the expected CSV.


g-weatherill:
LGTM

PR #2545, opened 2017-02-09
Add support for standalone apps in server

daniviga:
This is a preliminary work needed to be able to run our standalone apps (mainly `ipt` and `taxtweb`) using `openquake.server`.

This will simplify a lot the integration of such apps inside a single installer (which is a task of this sprint), specially on Windows where spawning to django instances in background requires some tricks.

Only few changes are required in the 'main' code everything else is hosted in the `local_settings.py` (a template is provided).

Minor changes to the template (of the apps) may be required. We'll see.


micheles:
LGTM

Issue #2522, opened 2017-02-06
The event loss table exporter is ultra-slow


micheles:
Reading the `rup_data` information is slow. For cases where that extra information is not needed probably we can speed up the exporter which now takes a lot more time than the computation itself for large datasets.

PR #2546, opened 2017-02-13
Stripped tectonic region type from the agg_loss_table exporter

micheles:
The tectonic region type can be extracted from the `sourcegroups.csv` output now, so we can avoid flooding the event loss table exporter with long strings. This saves some memory and export time, but we are still far from solving https://github.com/gem/oq-engine/issues/2522.

The manual has been updated: https://ci.openquake.org/job/builders/job/doc-builder/9/


micheles:
The problem is that the event_tag are ugly but unique (as requested by our sponsors) whereas the event IDs are not unique. Running the same computation on two different machines will give the same tags but different event IDs.

Notice that TRT has been already stripped by the ruptures.csv file, so the user has to download the sourcegroups file anyway.

PR #2548, opened 2017-02-13
Update a docstring in readinput.py

raoanirudh:
Removed a reference to a non-existent function `get_exposure_lazy`.


micheles:
LGTM

PR #2549, opened 2017-02-14
Magdist

micheles:
Companion of https://github.com/gem/oq-hazardlib/pull/591. The test case_19 has been changed to use a magnitude-dependent maximum distance: as expected, more ruptures are discarded and the computation becomes faster. We need to try this on the cluster.

While at it, I have improved the `oq info --report` command: now it reports the effective number of ruptures after rupture filtering.


openquake/qa_tests_data/classical/case_19/report.rst
```diff
@@ -70,19 +70,19 @@ Number of ruptures per tectonic region type
 ============================ ====== ================= =========== ============ ============
 source_model                 grp_id trt               num_sources eff_ruptures tot_ruptures
 ============================ ====== ================= =========== ============ ============
-simple_area_source_model.xml 4      Subduction Inslab 1           7770         7,770       
+simple_area_source_model.xml 4      Subduction Inslab 1           5680         7,770
```
ptormene: Please add the comma separator for thousands also to eff_ruptures

PR #2553, opened 2017-02-15
Simplified the event loss table CSV exporter

micheles:
By changing the way the data is stored. This also improves the performance and closes https://github.com/gem/oq-engine/issues/2522.

### DIFF
--- a/openquake/calculators/event_based_risk.py
+++ b/openquake/calculators/event_based_risk.py
@@ -36,7 +36,7 @@ F64 = numpy.float64
 getweight = operator.attrgetter('weight')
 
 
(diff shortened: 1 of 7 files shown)


openquake/calculators/event_based_risk.py
```diff
@@ -36,7 +36,7 @@
 getweight = operator.attrgetter('weight')
 
 
-def build_el_dtypes(insured_losses):
+def build_el_dtypes(loss_types, insured_losses):
     """
     :param bool insured_losses:
```
ptormene: Please add the description of `loss_types` to the docstring

PR #2556, opened 2017-02-16
SeisDepthNRML

rcgee:
This updates the classical qa test#25, now testing that a negative seismogenic depth can be used in the source_model.xml 

Companion of: https://github.com/gem/oq-hazardlib/pull/594


micheles:
LGTM

Issue #2491, opened 2017-01-26
Remove hdf5.vstr arrays from the datastore


micheles:
Variable-length strings are buggy in h5py version 2.6: if the array is short, everything works fine, but if the array is long it may fail in mysterious ways, including corrupting the datastore, saving the wrong arrays, etc etc. All that silently. We should at least remove vstr from long arrays.

PR #2558, opened 2017-02-17
Remove `hdf5.vstr` from the realizations array

micheles:
See https://github.com/gem/oq-engine/issues/2491. The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2398


micheles:
If they are not enough the strings will be silently truncated. This is not a disaster, since this a mere informational output. Notice that this is better than the present situation, were a big computation may mysteriously fail because the datastore gets corrupted (happened this night with a full SHARE).

PR #2559, opened 2017-02-17
DbServer is started twice in packager.sh

daniviga:
This could cause the stall in CI builds when a test is broken since the DbServer is locking the ssh session. Anyway, it's wrong.

https://ci.openquake.org/job/zdevel_oq-engine/2401/console


micheles:
LGTM

PR #2560, opened 2017-02-17
What's new in engine 2.3

micheles:
The usual document. Read it here: https://github.com/gem/oq-engine/blob/whatsnew/doc/whats-new.md


daniviga:
I would like to add some quick info about opt

doc/whats-new.md
```diff
@@ -1,356 +1,197 @@
...
-in the event based and scenario calculators. Moreover we have significantly
-improved our exports, by providing more information (especially for the
-event based ruptures) and/or using more efficient export formats, like the
-.npz format. As of now, the XML exports are deprecated, even if there is
-no plan to remove them at the moment.
+This release introduces several new features and improvements of the
+engine calculators. Moreover, our packaging strategy has been
+revolutionized and now we provide packages for nearly every operating
+system - including old ones like Ubuntu 12.04, previously abandoned - by
+using the exact same versions of the libraries on every supported
```
pslh: "... packages for most common operating systems - including previously abandoned systems such as Ubuntu 12.04 - by ..."

doc/whats-new.md
```diff
@@ -1,356 +1,197 @@
...
+platform (Linux, Windows, Mac).
```
pslh: It would be good if we could add here something like:  "A complete list of supported systems/packages is available from <link to page>"

doc/whats-new.md
```diff
@@ -1,356 +1,197 @@
...
+calculation time, since small magnitude ruptures will affects a lot
+less sites than before. In order to use the feature the user has to
+specify in the `job.ini` file a list of pairs (magnitude, distance) for each
+tectonic region type. In the past the user would write
+
+`maximum_distance = {'Active Shallow Crust': 200}`
+
+and the engine would use an integration distance of 200 km for all
+ruptures of the given tectonic region type, disregarding the magnitude.
+Nowadays, the user can write
```
pslh: "Now" would be more appropriate than "Nowadays".

doc/whats-new.md
```diff
@@ -1,356 +1,197 @@
...
-This release has focused on memory and performance optimizations, especially
-in the event based and scenario calculators. Moreover we have significantly
-improved our exports, by providing more information (especially for the
-event based ruptures) and/or using more efficient export formats, like the
-.npz format. As of now, the XML exports are deprecated, even if there is
-no plan to remove them at the moment.
+This release introduces several new features and improvements of the
+engine calculators. Moreover, our packaging strategy has been
+revolutionized and now we provide packages for nearly every operating
+system - including old ones like Ubuntu 12.04, previously abandoned - by
```
daniviga: `nearly every operating system` it's too optimistic IMHO

doc/whats-new.md
```diff
@@ -1,356 +1,197 @@
...
+constant maximum distance for all magnitudes, and the engine will keep
+working as in the past.
+
+A lot of work went into the UCERF calculators, but they are still officially
+marked as experimental and are left undocumented on purpose. There is a
+new time-dependent classical calculator, while the old calculators were
+substantially improved.
+
+There is an experimental command `run_tiles` to split a `sites.csv`
+file in tiles and run multiple calculations at once. This for expert
```
pslh: I suggest "This too is for use by experts only." but perhaps it would be useful to say how one can find out more or if we really don't want to talk about it, to not mention it in the what's new.  Also, should we rename run_tiles to experimental_run_tiles (or at least show a scary "This is experimental" message) ?

doc/whats-new.md
```diff
@@ -1,356 +1,197 @@
...
-====================================
+New features in the engine
+--------------------------
+
+From this release it is possible to take into account the *topography
+of the region of interest*. It is enough to write the elevation of the
+hazard sites in the sites file or in the site model file, as a
+negative depth, and the engine will compute the ground shaking
+correctly. Before there was no way to specify the elevation of the hazard sites
+and the ground shaking was computed only at the level of the sea.
```
pslh: "sea level"

doc/whats-new.md
```diff
@@ -1,356 +1,197 @@
...
+Changes
+------------
+
+There are two new limits in the event based calculators:
+
+1. it is impossible to run calculations with more than 65536 tasks
+2. it is impossible to have tasks generating more than 65536 events each
+
+The limits are there for safety purpose: users trying to run calculations
+impossibly large will get an error early, rather than trashing their machines.
```
pslh: These limits have been added to prevent users from running impossibly large calculations: an error message will be displayed early rather than causing machines to crash.

doc/whats-new.md
```diff
@@ -1,356 +1,197 @@
...
+
+The event loss table exporter has changed and it is now faster than in
```
pslh: The event loss table exporter is now faster ...

doc/whats-new.md
```diff
@@ -1,356 +1,197 @@
...
+anymore, however you can infer it from the `event_tag` field which
+contains the source group ID. The association between the source group
+ID and the tectonic region type is contained in the new export
+`sourcegroups.csv`.  The `source_id` has been stripped from the `event
+tag` since it was taking a lot of memory in the case of extra-large
+calculations (UCERF).
+
+The calculation of quantiles hazard curves used different algorithms
+in the case of logic tree full enumeration and logic tree sampling.
+Now, for consistency sake, we use always the same algorithm, the
```
pslh: For the sake of consistency we now use...

doc/whats-new.md
```diff
@@ -1,356 +1,197 @@
...
-In the future a lot more outputs are expected to be exported into
-`.npz` format.
+The "Remove" button now actually removes the calculation. Before
+it was only hiding it from the Web UI. Be careful with it, since it is
+final. If you remove a calculation, the only way to restore it is
+to repeat the computation.
+
+We fixed the names of the files downloaded from the Web UI. They were
+containing slashes, that were then replaced in a browser-dependent way
+with dashes or underscores. This is solved now.
```
pslh: The names of files downloaded from the Web UI previously contained slash ('/') characters that were then replaced with dashes or underscores by the browser.  We no longer include slash characters in file names.

doc/whats-new.md
daniviga: https://github.com/gem/oq-engine/blob/b6111109edac1caecf19715f58b64e8f06c19c84/doc/faq.md#supported-operating-systems which will become `faq.md#supported-operating-systems`

I would like do not duplicate the detailed lists of OSes (like windows 7,8,8.1,10 macos 10.10,10.11 and counting)

PR #2561, opened 2017-02-17
Generate sphinx doc in travis

daniviga:
It's fast and it does not impact on the overall build time. On the other side it's very usefull to identify broken docstrings during a PR, before merging into master.


micheles:
LGTM

PR #2572, opened 2017-02-20
Added an example with two IMTs in the classical_damage hazard curves

micheles:
This is just to exercise the function `get_mesh_csvdata`. Since the exposure does not have an asset of taxonomy "Z" the data with IMT SA(0.1) are not actually used. We should add a proper test for this situation.


micheles:
LGTM

PR #2573, opened 2017-02-20
Improve RabbitMQ start/stop in RPM spec

### LINKED ISSUE #2542: In RHEL stop rabbitmq only if was not already running
With https://github.com/gem/oq-engine/pull/2507 on RHEL we are starting RabbitMQ, installing the queues and then stopping it. We should stop RabbitMQ only if it has been started by us during `%pre`, if it was already running it should be kept running.

daniviga:
Closes #2542


daniviga:
This PR has been tested on a very slow machine (low power AMD Neo N36L) to check race conditions with rabbitmq start and vhost/vuser creation: the operations are made synchronously, so there's no race between them.

PR #2578, opened 2017-02-21
Changed the `build_rcurves` procedure to read the data directly from the workers

micheles:
This reduces enormously the data transfer via Python. The gain depends on the data stored in `all_loss_ratios`; it could go from 100 GB to zero. It requires to store the `all_loss_ratios` array in a separate HDF5 file opened in read mode: the datastore is opened in read-write mode and therefore its content is not accessible to the workers. It cannot be closed, since the risk curves are being saved while the workers work.

As an example, I ran a computation for Chile with 5,000 years of investigation time and 3.8 GB of loss ratios generated; the memory occupation in `build_rcurves` went down from 104 MB to 18 MB (there is no more unpickling). Also now the data transfer time is measured (it is the reading time inside the task). The total computation time improved by 40 seconds. Of course, the effect would be a lot bigger with a longer investigation time (i.e. 50,000 years, as wanted).

### DIFF
--- a/openquake/calculators/event_based_risk.py
+++ b/openquake/calculators/event_based_risk.py
@@ -26,6 +26,7 @@ from openquake.baselib.python3compat import zip
 from openquake.baselib.general import (
     AccumDict, humansize, block_splitter, group_array)
 from openquake.hazardlib.stats import compute_stats, compute_stats2
+from openquake.commonlib import config
 from openquake.calculators import base, event_based
 from openquake.baselib import parallel
 from openquake.risklib import riskinput, scientific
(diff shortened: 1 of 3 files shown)


openquake/calculators/event_based_risk.py
```diff
@@ -80,21 +82,22 @@ def build_agg_curve(cb_inputs, monitor):
     return result
 
 
-def build_rcurves(rlzname, data, cbs, assets, monitor):
+def build_rcurves(h5path, rlzname, cbs, assets, monitor):
     """
     :param rlzname: string of the form `rlz-\d\d\d\d`
-    :param data: `all_loss_ratios` data for the realization
     :param cbs: list of `L` CurveBuilders instances
-    :param assets: full list of assets
```
ptormene: Please fix the docstring

PR #2598, opened 2017-03-01
64 bit <-> 32 bit mismatch

micheles:
Consequence of https://github.com/gem/oq-hazardlib/pull/605. The tests are green: https://ci.openquake.org/job/zdevel_oq-engine/2429


daniviga:
In travis it fails because:

```python
numpy.max(numpy.abs(dists)) > tolerance
0.00097656 > 0.000763410137435
```
in https://github.com/gem/oq-hazardlib/blob/point3d/openquake/hazardlib/geo/surface/planar.py#L140

Travis probably is doing some different rounding

daniviga:
I've been able to reproduce the error on a Ubuntu 12.04 LXC container (using our wheels).

It's seem related to an underlying library, not strictly related to libgeos, numpy, scipy... since we are using the same wheels. It isn't also related to Python since on my tests I was using Python 2.7 (2.7.3) and Travis uses Python 3.5 (3.5.2 I guess).
It must be GLIBC. More investigation could require a huge on investigation through C calls. Maybe we should ask scientists (like @mmpagani or @vpoggi) if we can increase the tolerance.

Thanks to @nastasi-oq for the idea about manually running tests on Precise (since Travis runs on it).

daniviga:
This is the current situation:
- Ubuntu 16.04: OK
- Ubuntu 14.04: OK
- Ubuntu 12.04: KO
- Fedora 25: OK
- CentOS 7: OK
- CentOS 6: KO
- CentOS 5: KO

CentOS 5, 6 and Ubuntu 12.04 are producing the same numbers. Any distro with LIBC <= 2.15 breaks the tests.

We can then
- Skip this tests if LIBC <= 2.15
OR
- Keep numbers as 64bit
OR
- Increase the tolerance

micheles:
We will increase the tolerance. We are already doing that in UCERF, anyway. Thanks for figuring out the issue which was extremely tricky to debug (Matteo's insight is also appreciated).

PR #2602, opened 2017-03-02
 Add a doc page with useful commands for development and debugging

daniviga:
Feel free to add/suggest any missing command/utility/feature

### DIFF
--- a/doc/internals.md
+++ b/doc/internals.md
@@ -1,0 +1,37 @@
+# Internal commands
+
+This commands and internals are meant to be used for debug and development. DO NOT USE them in production. They MAY CHANGE ANYTIME, without notice.
+
+### Database creation
+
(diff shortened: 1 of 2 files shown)


doc/internals.md
```diff
@@ -0,0 +1,37 @@
...
+The OpenQuake server uses Django for the web/API application. Management can be done via:
+
+```bash
+$ python -m openquake.server.manage [help]
+```
+
+### DbServer
+
+Sometimes it's useful to run the DbServer in background for debug purposes. It can be started like this:
+
```
micheles: I think you meant "foreground" here.

micheles:
I should add a page specific to the `oq` commands, but it is too early for that. Anyway, apart from a comment, LGTM

PR #2605, opened 2017-03-06
Extend `oq to_hdf5` to convert source models into HDF5 files

micheles:
Requires https://github.com/gem/oq-hazardlib/pull/607


mmpagani:
Very interesting. Are we planning to support a "to_hdf5" feature into the OQ engine and/or the possibility of reading an .hdf5 formatted input file? 
LGTM

micheles:
Yes, we have discussed about that for a long time and I think that now we are close to reach that goal.

PR #2608, opened 2017-03-07
Fix a bug when deleting a calculation from the WebUI

micheles:
Paolo found this error:

```python
  File "/home/paolo/projects/oq-engine/openquake/server/db/actions.py", line 276, in del_calc
    'dependent from it: %s' % (job_id, [j.id for j in dependent]))
TypeError: %d format: a number is required, not unicode
ERROR Internal Server Error: /v1/calc/73/remove
```


micheles:
It should be fixed now

micheles:
A have opened an issue for the required improvement: https://github.com/gem/oq-engine/issues/2611

PR #2609, opened 2017-03-07
Use a ThreadPoolExecutor instead of a ProcessPoolExecutor in dbserver.py

daniviga:
because SQLite in macOS is not fork-safe; using the ProcessPoolExecutor a 'ProgrammingError: Cannot operate on a closed database' error is raised. Se also https://bugs.python.org/issue27126

I would like to make some extensive manual testing on other platforms before merging this code.

lin: https://ci.openquake.org/job/zdevel_oq-engine/2432/
mac: https://ci.openquake.org/job/macos/job/zdevel_mac_oq-engine/88/
win: https://ci.openquake.org/job/windows/job/zdevel_oq-engine/4/

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,6 +1,6 @@
   [Daniele Viganò]
   * Use threads instead of processes in DbServer because SQLite3
-    isn't fork-safe on macOS
+    isn't fork-safe on macOS Sierra
 
   [Michele Simionato]
   * Extended the command `oq to_hdf5` to manage source model files too
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,7 @@
+  [Daniele Viganò]
+  * Use threads instead of processes in DbServer because SQLite3
+    isn't fork-safe on macOS
+
   [Michele Simionato]
   * Extended the command `oq to_hdf5` to manage source model files too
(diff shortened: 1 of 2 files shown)


debian/changelog
```diff
@@ -1,3 +1,7 @@
+  [Daniele Viganò]
+  * Use threads instead of processes in DbServer because SQLite3
+    isn't fork-safe on macOS
+
```
micheles: I would say `is not fork-save on macOS Sierra`

daniviga: It seems to be so also on some of the latest Mac OS X 10.11

openquake/server/dbserver.py
```diff
@@ -34,7 +34,7 @@
 from openquake.server import dbapi
 from openquake.server.settings import DATABASE
 
-executor = ProcessPoolExecutor(1)  # there is a single db process
+executor = ThreadPoolExecutor(1)  # there is a single db process
```
micheles: Please replace the comment `# there is a single db process` with something like `# using a ThreadPool because fork and sqlite3 do not play well on macOS Sierra`

micheles:
LGTM

PR #2614, opened 2017-03-08
Moved the matplotlib imports to toplevel

micheles:
This is causing a very ugly error: https://ci.openquake.org/job/zdevel_oq-engine/2434/console
My guess is that matplotlib is not available on Jenkins and we are getting an error masking the real issue.

### DIFF
--- a/openquake/commands/plot.py
+++ b/openquake/commands/plot.py
@@ -17,6 +17,7 @@
 # along with OpenQuake. If not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import print_function
+import matplotlib.pyplot as plt
 from openquake.baselib import sap
 from openquake.commonlib import datastore
 from openquake.commands.show import get_hcurves_and_means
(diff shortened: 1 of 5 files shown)


daniviga:
the DbServer won't start becase `python-tk` (tkinter) is missing. On Ubuntu, by default, `python-tk` isn't installed becasue has a lot of graphical dependencies.

The only way to start the `DbServer` or run the engine without `tkinter` is passing `TEMPLATE=Agg` before `oq` is run. This could complicate a lot the things.

We could try setting, in the python code, TEMPLATE=Agg and switch it to a real template only where the plotting functions are really used, making tkinter optional.

daniviga:
A simple solution could be:

```python
import matplotlib as mpl
# [cut]

mpl.pyplot.figure()
```
 in any case, on Ubuntu and RHEL to have all the features (including plotting) we'll need to add `python-tk` and `tikinter` to `control` and `spec` respectively.

I'll check how this works on other platforms (but I'm confident that both Windows' python and Apple's one already ship with tkinter)

micheles:
I propose to close this pull request without merging it. It was a bad idea. Right now matplotlib is imported only when needed, while if this PR was merged, we will pay the cost of importing it always, even when doing `oq help`. This is absurd since matplotlib is *never* used by the engine and it should never be used by it. It must stay a debugging plotting utility external to the engine code base.

Sorry for the noise.

micheles:
Superseded by https://github.com/gem/oq-engine/pull/2615

PR #2629, opened 2017-03-14
Small fixes to the event based risk calculator

micheles:
This are needed so that Anirudh's tests can run.


micheles:
The changes are very minor and I have an informal "go ahead" from Anirudh, so I am merging.

PR #2631, opened 2017-03-14
Removed `check_platform` because we have /opt now

micheles:
Unskipped the tests that very passing only on Ubuntu 16.04: now they pass on most platforms, thanks to the opt project. See https://ci.openquake.org/job/zdevel_oq-engine/2445/
https://ci.openquake.org/job/windows/job/zdevel_oq-engine/6/
https://ci.openquake.org/job/macos/job/zdevel_mac_oq-engine/90/

### DIFF
--- a/openquake/calculators/tests/__init__.py
+++ b/openquake/calculators/tests/__init__.py
@@ -38,14 +38,6 @@ def strip_calc_id(fname):
     return re.sub('_\d+\.', '.', name)
 
 
-def check_platform(*supported):
-    """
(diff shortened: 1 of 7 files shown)


daniviga:
Why you had to update numbers in test outputs?

micheles:
The file was generate two years ago and then the test was skipped. In two years many things have changed. My guess is that here we simply changed the seed management.

daniviga:
LGTM

PR #2632, opened 2017-03-16
Simplified the management of the assets

micheles:
Right now we have two objects, the AssetCollection and the list of lists `assets_by_site` which contains the assets per each site. Since the latter can be expressed in terms of the former, this pull request removes the attribute `calculator.assets_by_site`.


openquake/calculators/base.py
```diff
@@ -352,13 +352,16 @@ def assoc_assets_sites(self, sitecol):
                 'maximum distance of %s km' % maximum_distance)
         mask = numpy.array([sid in assets_by_sid for sid in sitecol.sids])
         assets_by_site = [assets_by_sid.get(sid, []) for sid in sitecol.sids]
-        return sitecol.filter(mask), numpy.array(assets_by_site)
+        return sitecol.filter(mask), riskinput.AssetCollection(
```
ptormene: Please describe the return value accordingly in the docstring

openquake/commonlib/readinput.py
```diff
@@ -764,7 +768,10 @@ def get_sitecol_assets(oqparam, exposure):
     for lonlat in zip(sitecol.lons, sitecol.lats):
         assets = assets_by_loc[lonlat]
         assets_by_site.append(sorted(assets, key=operator.attrgetter('idx')))
-    return sitecol, numpy.array(assets_by_site)
+    return sitecol, riskinput.AssetCollection(
```
ptormene: Same as above

Issue #2529, opened 2017-02-06
Let the dbserver listening on localhost too even if IP is set


daniviga:
Currently in `openquake.cfg` we have a single config parameter called `host` for the dbserver: this is used for both the dbserver IP (where to push messages) and as the IP where to listen at. On a single machine it's left to `localhost` which is fine. On a cluster the master node IP must be set instead: on master this has the side effect that dbserver is not available on `localhost` but only on the iface associated to that IP.

My proposal is not to change the `openquake.cfg` (which would mean split `host` into `host` and `bind_address`) but, when host != localhost, is to bind dbserver on the defined IP _and_ localhost.

PR #2634, opened 2017-03-20
Differentiate listening interface and connect interface in dbserver

daniviga:
~~This PR differentiate the setting used to connect at a dbserver to the one which describes the interface where dbserver is attached to (i.e. 0.0.0.0, lo, a specific ip...)~~

After some more thought I decided to keep everything simple as it is, which works for most users, changing only the `dbserver` command line: This change allows smart user to differentiate the binding interface from the client host settings.

the `openquake.server.dbserver` command is changed from
```python
python openquake.server.dbserver /path/to/file:port
```
to
```python
python openquake.server.dbserver listeningip:port /path/to/file
```

This arguments can be now also passed via `oq dbserver`
```bash
$ oq dbserver start 0.0.0.0:1985 /tmp/dbfoo.sqlite3
```

https://ci.openquake.org/job/zdevel_oq-engine/2447/

Closes https://github.com/gem/oq-engine/issues/2529

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,7 @@
+  [Daniele Viganò]
+  * Introduced an optional argument for 'oq dbserver' command line 
+    to be able to override its default interface binding behaviour
+
   [Michele Simionato]
   * Added a check on missing vulnerability functions for some loss type
     for some taxonomy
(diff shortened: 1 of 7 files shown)


micheles:
I like this. Please update the changelog. Other than that, LGTM.

PR #2638, opened 2017-03-20
Small fixes for local_settings template to be able to append settings

daniviga:
The `local_settings.py` template was actually broken. This file isn't currently used, but it will be soon by the standalone apps.


micheles:
LGTM

PR #2641, opened 2017-03-21
Check if oq is talking to a foreign DbServer

### LINKED ISSUE #2563: Make sure you're taliking at the right dbserver
When `multi_user = false` we must make sure that `oq` is talking to the right DbServer and not to another user DbServer (which may use a different code base)

An easy way would be, when the first connection is made, comparing the `openquake.server` path of `oq` and of the DbServer.

daniviga:
See https://github.com/gem/oq-engine/issues/2563

This doesn't ensure you have the DbServer codebase aligned with the client one, but at least, it checks that you aren't talking to another user/instance DbServer.

This is useful on a server where there's a centralized installation and a user creates her own venv. The checks is done only if the client configuration has `multi_user` false. Only a warning is printed.

Closes #2563


daniviga:
This should be probably checked in the rest API server too

openquake/commands/engine.py
```diff
@@ -123,6 +123,13 @@ def engine(log_file, no_distribute, yes, config_file, make_html_report,
 
     dbserver.ensure_on()
 
+    # check if the db is the right one
```
daniviga: DB -> dbserver

debian/changelog
```diff
@@ -1,4 +1,6 @@
   [Daniele Viganò]
+  * Added a check to verify that an 'oq' client is talinking to
+    the right DbServer
```
micheles: misprint in `talinking`

micheles:
Why a warning and not an error? Using the DbServer of a different code base is dangerous. When I run calculations on Wilson in my virtualenv I use `oq run` and not `oq engine` exactly to bypass the dbserver.
An user wanting to run the full engine on a virtualenv should start the DbServer on a different port.

daniviga:
Done, now a `sys.exit` is called if the dbserver is the wrong one.

openquake/server/manage.py
```diff
@@ -53,6 +53,8 @@ def dbcmd(action, *args):
     if 'runserver' in sys.argv:
         if '--nothreading' in sys.argv:
             logs.dbcmd = dbcmd  # turn this on when debugging
+        # check if the dbserver is the right one
+        dbserver.check_foreign() # check if we are taliking to the right server
         logs.dbcmd('upgrade_db')  # make sure the DB exists
```
micheles: I would change `check_foreign` to return a string and I would use it as follows:

```python
err = dbserver.check_foreign() 
if err:
    sys.exit(err)
```

micheles:
Other than the minor comment, LGTM

openquake/commands/engine.py
```diff
@@ -122,7 +122,10 @@ def engine(log_file, no_distribute, yes, config_file, make_html_report,
         os.makedirs(datastore.DATADIR)
 
     dbserver.ensure_on()
-    dbserver.check_foreign()
+    # check if we are taliking to the right server
```
micheles: Misprint in `taliking` ;-)

openquake/server/manage.py
```diff
@@ -53,8 +53,10 @@ def dbcmd(action, *args):
     if 'runserver' in sys.argv:
         if '--nothreading' in sys.argv:
             logs.dbcmd = dbcmd  # turn this on when debugging
-        # check if the dbserver is the right one
-        dbserver.check_foreign() # check if we are taliking to the right server
+        # check if we are taliking to the right server
```
micheles: Even here ;-)

PR #2653, opened 2017-03-24
Regenerated the expected files after the fix in https://github.com/gem/oq-hazardlib/pull/617

micheles:
Catalina needs to try this branch.
Companion of https://github.com/gem/oq-hazardlib/pull/617.


micheles:
LGTM

Issue #2650, opened 2017-03-23
Add .npz exports for some risk outputs


micheles:
- [ ] avg_losses, shape (A, R, LI)
- [x] losses_by_asset, shape (A, R, LI) with `stats_dt`
- [x] loss_maps, shape (A, R) with `loss_type_poe_dt`

micheles:
For loss maps and curves it is probably better to have a limited export able to export a single asset. Then there will be a REST API calling that exporter, usable by QGIS for plotting purposes.

micheles:
We agreed with Catalina to rename the `avg_losses` outputs as `losses_by_asset` (see https://github.com/gem/oq-engine/issues/2673) for consistency with the other calculators, so the existing .npz exporter for `losses_by_asset` can be reused.

PR #2656, opened 2017-03-24
Changes to the risk exports

micheles:
1. The average losses in event based risk calculators are now stored as an array of shape (A, R, L') where A is the number of assets, R the number of realizations, L is the number of loss types and L' is L or 2*L depending if there are insured losses. Before we were using a more complicated array of arrays.

2. the output `avg_losses-stats` is now dynamic, i.e. it is not stored in the datastore but instead computed on the fly at export time, when requested, consistently with several other statistical outputs.

3. the `avg` field in the exported loss curves is now computed as in the past, with a call to  `scientific.average_loss([losses, poes])`. This is also consistent with the algorithm used in the calculation of the aggregated loss curve. We changed the approach in the past for performance reasons, but now that the loss curves are computed differently (via the .ext5 file) we can go back to the original algorithm.

This is preliminary to https://github.com/gem/oq-engine/issues/2650.


openquake/calculators/event_based_risk.py
```diff
@@ -510,13 +503,8 @@ def save_results(self, allres, num_rlzs):
         if avg_losses:
             # since we are using a composite array, we must use fillvalue=None
             # and then set the array to 0 manually (to avoid bogus numbers)
```
ptormene: Please remove the comment

openquake/engine/engine.py
```diff
@@ -84,21 +84,25 @@ def expose_outputs(dstore):
...
     if oq.uniform_hazard_spectra:
         dskeys.add('uhs')  # export them
     if oq.hazard_maps:
         dskeys.add('hmaps')  # export them
+    if 'avg_losses-rlzs' in dstore and rlzs:
+        dskeys.add('avg_losses-stats')
     if oq.conditional_loss_poes:  # expose loss_maps outputs
         if 'rcurves-rlzs' in dstore or 'loss_curves-rlzs' in dstore:
-            dskeys.add('loss_maps-rlzs')
+            if rlzs:
```
ptormene: Please change into `if len(rlzs) > 1` (also below)

PR #2661, opened 2017-03-27
Changed the Web UI button from "Run Risk" to "Continue"

micheles:
"Run Risk" is a bad name. For instance one can run a classical calculation and then click "Continue" to run another classical calculation computing hazard maps with different PoEs. The "Continue" feature has nothing to do with risk per se: it just means that the calculation is split in multiple steps. Paolo may want to do the same change in the QGIS side.


doc/running/server.md
```diff
@@ -44,7 +44,7 @@ Output availability depends on the type of artifacts generated by the calculatio
 
 ### Run a risk calculation on top of hazard
 
-To run a risk-only calculation on top of a previous hazard run you have to click the `Run Risk` button on the corresponding hazard calculation and select the zip file containing the risk `job.ini` and the related inputs
+To run a risk-only calculation on top of a previous hazard run you have to click the `Continue` button on the corresponding hazard calculation and select the zip file containing the risk `job.ini` and the related inputs
```
ptormene: The whole sentence (and the title) should be changed, explaining as in the subject of this PR.

micheles: Fixed the docs

Issue #2645, opened 2017-03-22
Store the ruptures in a better way


micheles:
Now that the project /opt has landed, we may rely on having h5py version 2.6. This means that we could store the rupture geometries are variable-length arrays of points. This should improve a lot the time to store and read the ruptures.

daniviga:
Please open an issue on oq-libs asking for inclusion of h5py 2.6

---
Daniele Viganò
http://daniele.vigano.me

On mobile, please excuse my brevity.

daniviga:
Version 2.7 is out and it seems to include several enhancements, including one variable lenght arrays: should we move to it?

micheles:
Not sure; first of all we should try to run all our tests and confirm than they are green. Then I should run some cases that were broken in the past and see if they are fixed now. We will see.

PR #2666, opened 2017-03-29
Changed the storage of the events

micheles:
Fourth step towards #2645.
Currently the events are store twice, in the `events` tables and in the ruptures. With this pull request the events are stored only once; the ruptures do not contain copies of the events, only indices to the their location in the events table.

Here is the saving in space for the Chile computation seen in https://github.com/gem/oq-engine/pull/2660:

```
master: 641.52 MB
this branch: 586.17 MB
```

The time is more or less the same.

### DIFF
--- a/openquake/commonlib/calc.py
+++ b/openquake/commonlib/calc.py
@@ -636,7 +636,7 @@ class RuptureSerializer(object):
         Populate a dictionary of site IDs tuples and save the ruptures.
 
         :param ebruptures: a list of EBRupture objects to save
-        :param offset: the last event index saved
+        :param eidx: the last event index saved
         """
         for ebr in ebruptures:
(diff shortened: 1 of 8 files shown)


openquake/commonlib/calc.py
```diff
@@ -626,11 +631,18 @@ def __init__(self, datastore):
...
         self.data = []
 
-    def save(self, ebruptures):
+    def save(self, ebruptures, eidx):
         """
-        Collect the ruptures and a set of site IDs tuples.
+        Populate a dictionary of site IDs tuples and save the ruptures.
+
+        :param ebruptures: a list of EBRupture objects to save
+        :param offset: the last event index saved
```
ptormene: Please replace `offset -> eidx` as in the signature

PR #2674, opened 2017-03-29
Ultra-fast rupture storage

micheles:
Fifth step towards https://github.com/gem/oq-engine/issues/2645. Here are the numbers for the Chile calculation described in https://github.com/gem/oq-engine/pull/2660, which has 176,946 ruptures and 251,415 events:

```
operation                        time_sec  memory_mb counts 
saving ruptures                  10        40        107    
```

i.e. we passed from 31 minutes (engine-2.3) to 10 seconds!
The disk occupation went down from 586.17 MB to 189.2 MB.


openquake/commonlib/calc.py
```diff
@@ -667,3 +660,51 @@ def close(self):
...
         self.datastore.flush()
         del self.data[:]
+
+
+def get_ruptures(dstore, grp_id):
+    """
+    Extracts the ruptures of the given grp_id
+    """
+    oq = dstore['oqparam']
+    # oq.complex_fault_mesh_spacing
```
ptormene: Please remove this line

PR #2675, opened 2017-03-29
Fix output encoding in lhc and lrc

### LINKED ISSUE #2662: lhc and lrc output is broken
it does not print strings is UTF-8
```
$ oq engine --lhc
job_id |     status |          start_time |         description
b' 12855 | successful | 2016-12-05 14:57:50| Classical PSHA \xe2\x80\x94 Area Source'
b' 13047 | successful | 2017-01-31 09:14:36| Classical PSHA with non-trivial logic tree (2 source models and 2 GMPEs per tectonic region type)'
b' 13048 |     failed | 2017-01-31 09:17:41| Classical PSHA with non-trivial logic tree (2 source models and 2 GMPEs per tectonic region type)'
b' 13755 | successful | 2017-03-05 13:30:36| Classical PSHA with non-trivial logic tree (1 source model + absolute uncertainties on G-R a and b values and maximum magnitude and 2 GMPEs per tectonic region type)'
b' 13758 | successful | 2017-03-07 10:10:16| Classical PSHA with non-trivial logic tree (1 source model + absolute uncertainties on G-R a and b values and maximum magnitude and 2 GMPEs per tectonic region type)'
```

```bash
$ env | grep LANG
LANG=en_US.utf8
```

Tested with Python 3 on both Wilson and my laptop.

daniviga:
Closes #2662 

The output on windows still isn't valid UTF-8 (as before), just because the terminal does not support for it.

Could be improved more, but at least now it works.

### DIFF
--- a/openquake/commands/engine.py
+++ b/openquake/commands/engine.py
@@ -20,6 +20,7 @@ import os
 import sys
 import getpass
 import logging
+import codecs
 from openquake.baselib import sap
 from openquake.commonlib import datastore, config, logs
 from openquake.engine import engine as eng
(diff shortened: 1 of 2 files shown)


micheles:
LGTM

PR #2679, opened 2017-03-30
Improve the WebUI command

### LINKED ISSUE #2676: oq webui start not working as expected
I just installed oq 2.3 via apt on ubuntu 16.04. I wanted to try the api so I ran

    oq webui start

which just gave me

"/usr/bin/python: No module named openquake.server"

daniviga:
This PR improves the behavior of the `oq webui` command.

Now it can be used in both single and multi user environment; in a multi-user environment it does tasks only if the command is run by the same user that is the database owner.

The `oq` entry point has been improved too. Now, when OQ is installed in `/opt` (packages), the `sys.path` is saved inside `os.environ['PYTHONPATH']` to pass the proper configuration to any subproccess call.

Tests (change affects only linux + packages): https://ci.openquake.org/job/zdevel_oq-engine/2450/

Fixes #2676. Thanks to @3rdcycle for the feedback.

### DIFF
--- a/bin/oq
+++ b/bin/oq
@@ -26,6 +26,7 @@ if os.path.realpath(__file__) == '/opt/openquake/bin/oq':
     # https://docs.python.org/2/library/sys.html#sys.path
     sys.path.insert(1, '/opt/openquake/lib/python' + str(sys.version_info[0]) +
                     '.' + str(sys.version_info[1]) + '/site-packages')
+    os.environ['PYTHONPATH'] = ":".join(sys.path)
 
 import openquake.commands.__main__ as main
 
--- a/bin/oq
+++ b/bin/oq
@@ -20,12 +20,13 @@
 import sys
 import os
 
+# this is only executed when openquake is installed via linux binary packages
 if os.path.realpath(__file__) == '/opt/openquake/bin/oq':
     # custom pythonpath is pushed in position 1 instead of 0 because
     # sys.path[0] is populated at runtime by python itself. See:
     # https://docs.python.org/2/library/sys.html#sys.path
-    sys.path.insert(1, '/opt/openquake/lib/python' + str(sys.version_info[0]) +
-                    '.' + str(sys.version_info[1]) + '/site-packages')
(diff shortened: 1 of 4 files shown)


openquake/commands/webui.py
```diff
@@ -34,8 +37,13 @@ def webui(cmd, hostport='127.0.0.1:8800'):
     start the webui server in foreground or perform other operation on the
     django application
     """
-    dbserver.ensure_on()  # start the dbserver in a subproces
+
+    db_owner = os.path.expanduser(config.get('dbserver', 'file'))
+    if not os.access(db_owner, os.W_OK):
```
micheles: I would use the name `db_path` not `db_owner`

bin/oq
```diff
@@ -26,6 +26,7 @@ if os.path.realpath(__file__) == '/opt/openquake/bin/oq':
     # https://docs.python.org/2/library/sys.html#sys.path
     sys.path.insert(1, '/opt/openquake/lib/python' + str(sys.version_info[0]) +
                     '.' + str(sys.version_info[1]) + '/site-packages')
+    os.environ['PYTHONPATH'] = ":".join(sys.path)
```
micheles: How does this work on Windows?
BTW, I would write

```python
 sys.path.insert(1, '/opt/openquake/lib/python%d.%d/site-packages' % sys.version_info[:2])

micheles:
We could add in the packager a call to `oq webui` to test the message about not being the right user.

bin/oq
daniviga: On Windows we never enter that if, so this code will never executed (anyway, `os.environ` + `PYTHONPATH` works too, we use it in `.bat` entry points).

daniviga:
New tests: https://ci.openquake.org/job/zdevel_oq-engine/2451/

daniviga:
https://ci.openquake.org/job/zdevel_oq-engine/2453/

daniviga:
Tests are green: https://ci.openquake.org/job/zdevel_oq-engine/2454/

micheles:
Please update the changelog. For the rest, LGTM

PR #2687, opened 2017-03-31
Removed `rup_data` output

micheles:
It is redundant now that the ruptures are always stored. Here is a run on Ubuntu 16.04: https://ci.openquake.org/job/zdevel_oq-engine/2457. Closes https://github.com/gem/oq-engine/issues/2645.


openquake/calculators/tests/ucerf_test.py
```diff
@@ -32,8 +33,9 @@ def test_event_based(self):
         # check that we get the expected number of events
         with open(fname) as f:
             self.assertEqual(len(f.readlines()), 36)
-        # check the header and the first 18 events
-        self.assertEqualFiles('expected/ruptures.csv', fname, lastline=19)
+        if 'Ubuntu-16.04' in platform.platform():
```
ptormene: Please use a constant instead of all occurrences of this

Issue #2668, opened 2017-03-29
Scenario Risk csv outputs


CatalinaYepes:
- [x] Check the row index for the `agg_losses`
- [x] Include exporter for `losses_by_taxon`
- [x] Unify the exporters name (especially the GMPE name) in some cases the output has the name of the GMPE, in others the realization id.

PR #2693, opened 2017-04-03
Added output `losses_by_taxon-rlzs`

micheles:
Implements the second point in https://github.com/gem/oq-engine/issues/2668


openquake/qa_tests_data/scenario_risk/case_master/expected/losses_by_taxon-rlz-000.csv
```diff
@@ -0,0 +1,4 @@
+taxonomy,business_interruption,contents,nonstructural,occupants,structural,business_interruption_ins,contents_ins,nonstructural_ins,occupants_ins,structural_ins
+"tax1",8.48713E+04,5.26754E+05,7.59653E+05,NAN,3.93912E+05,8.48713E+04,5.26754E+05,7.59653E+05,NAN,3.93912E+05
```
ptormene: Is it correct that the `occupants` column contains NAN?

micheles: Yes, it is totally correct. There is no concept of insurance for occupants, so the column is left to NaN.

PR #2712, opened 2017-04-10
Sync deps with oq-hazardlib + hmtk

daniviga:
Sync with https://github.com/gem/oq-hazardlib/pull/623 to keep requirements homogenized.


micheles:
LGTM if tests on Mac are green

PR #2729, opened 2017-04-21
Added a CSV exporter for the benefit-cost-ratio calculator

micheles:
Since this calculator is never used, we forgot to implement a CSV exporter for it. The next step will be to remove the XML exporter. Closes https://github.com/gem/oq-engine/issues/2697.


CatalinaYepes:
LGTM

PR #2735, opened 2017-04-24
Made the slow classical tests fast by increasing the mesh spacing

micheles:
So that there is no need to skip them.


openquake/qa_tests_data/classical/case_5/job.ini
```diff
@@ -17,7 +17,7 @@ number_of_logic_tree_samples = 0
 
 # km
 rupture_mesh_spacing = 0.01
-complex_fault_mesh_spacing = 0.01
+complex_fault_mesh_spacing = 0.1
```
ptormene: Please check if you might want to change also `rupture_mesh_spacing = 0.1` as in other cases

micheles: No, I kept them different on purpose. This calculation is dominated by the  `complex_fault_mesh_spacing` since it has a complex fault.

PR #2742, opened 2017-04-27
Use CDN ftp mirror instead of upstream

daniviga:
Download the wheels from an ftp proxy on Linode instead of directly from our internal ftp. This reduces the load on our uplink and the cdn can provide stale cache if upstream is down.


micheles:
LGTM

Issue #2673, opened 2017-03-29
Event based outputs


CatalinaYepes:
- [x] The losses in the `loss_curves` output are missing (only poes are exported)
- [x] The loss curves should be generated in `csv` (quantiles and mean are in `xml`)
- [x] Unify the name for the loss curve exporter: in the classical calculator they are called `loss_curves`, while in this calculator they are called `rcurves-rlz-...`
- [x] Modify the `event_tag` to be a simple number
- [x] The `agg_losses` output should:
     - Include a column for `rup_id` that matches with the `ruptures.csv` and `ruptures.xml` file
- [x] For the `ruptures.csv` file:
     - Rename the column `serial` for `rup_id`
     - Investigate why the strike values are sometimes NaN (like in the master_case)
     - Remove the column `eid`
     - reorder the columns: first `rupid` then `multiplicity`
 - [x] For the `ruptures.xml` file:
     - Remove duplicate ruptures
     - Include a `rupid` consistent with the csv file
     - Include the multiplicity
     - Include the list of all `event_id` in which the rupture is used

PR #2751, opened 2017-04-28
Renamed the datasets `avg_losses-` to `losses_by_asset-`

micheles:
See https://github.com/gem/oq-engine/issues/2673.


micheles:
After discussion with Catalina and Anirudh we decided not to do this.

PR #2772, opened 2017-05-05
Use a real venv, don't mess up PYTHONPATH

daniviga:
**WIP**
Thanks to the work we've done with `oq-libs` and with the inits this is now finally possible.

This change has several advantages:
- Get rid of any manipulation of the PYTHONPATH
- PYTHONPATH doesn't need to be specified anymore, instead it's possible to call the proper binaries in `/opt/openquake/bin` (celery, mock, pip...) direclty
- Makes the codebase portable to python3 (we'll need just to update `oq.patch`)
- Sysadmins can easily add dependencies in the OQ venv calling `/opt/openquake/bin/pip` directly, with no other changes
- `pth` files are properly managed
- We don't need to relay anymore on 'magic' pip flags (see oq-libs), everything is more standard
- Binary packages are consistent with manual installations on clusters

Backward compatibility is kept, since PYTHONPATH can still be used, if properly configured.

Marked as 2.5.0 because it will not block a release, but I would like to get this in soon.

Ubuntu: https://ci.openquake.org/job/zdevel_oq-engine/2499/
RHEL: https://ci.openquake.org/job/builders/job/rpm-builder/78/

Requires: https://github.com/gem/oq-libs/pull/20


daniviga:
This work will be superseded by #2803 Phase 2

PR #2775, opened 2017-05-06
Removed the .txt exporter for the GMFs

micheles:
For several years we had an internal .txt exporter for the ground motion fields, used only in the tests. Since now we finally have a good .csv format, it is time to remove it.


openquake/calculators/tests/event_based_test.py
```diff
@@ -147,36 +144,27 @@ def test_case_1(self):
...
     def test_minimum_intensity(self):
-        out = self.run_calc(case_2.__file__, 'job.ini', exports='txt',
+        out = self.run_calc(case_2.__file__, 'job.ini', exports='csv',
                             minimum_intensity='0.4')
 
-        [fname] = out['gmf_data', 'txt']
-        self.assertEqualFiles(
-            'expected/minimum-intensity-SadighEtAl1997.txt', fname, sorted)
+        [fname] = out['gmf_data', 'csv']
+        self.assertEqualFiles('expected/mininim-intensity-gmf-data.csv', fname)
```
ptormene: Typo

raoanirudh:
Is the CSV format for GMFs documented somewhere?

micheles:
No, Anirudh, the event based CSV export format is not documented, but I am fine with it since it is not official (the XML is still the official format and it is not deprecated yet). On the other hand the CSV export format for scenario (which is totally different) should be documented but I am not sure it is.

PR #2780, opened 2017-05-08
Avoid duplication in the SES XML exporter

micheles:
We removed the duplication of ruptures in the exported XML file, as requested in https://github.com/gem/oq-engine/issues/2673. The price to pay is an incompatible change in the format. Now each rupture contains the stochastic event sets it occurs in, as well as its event IDs.

Here is an example of output: https://github.com/gem/oq-engine/blob/ses/openquake/qa_tests_data/event_based/case_17/expected/ses.xml


openquake/commonlib/hazard_writers.py
```diff
@@ -479,20 +481,17 @@ class SESXMLWriter(object):
     def __init__(self, dest):
         self.dest = dest
 
-    def serialize(self, data):
+    def serialize(self, data, investigation_time):
```
ptormene: Please add the description of investigation_time to the docstring

micheles: Done

PR #2784, opened 2017-05-09
Fixed the event IDs in the test case_miriam

micheles:
Fixes the error on Jenkins: https://ci.openquake.org/job/master_oq-engine/3466/

The numeric event IDs are unique within a calculation but they depend on the number of tasks (*and this should be documented in the manual*). So the only way to get the tests green on different platforms (with different number of cores) is to set concurrent_tasks on the `job.ini`. The event tags did not have this problem, but nobody liked them except me.

While at it, I have also fixed the order of the events according to the event ID, since it is efficient now that it is numeric.


micheles:
LGTM

Issue #2768, opened 2017-05-05
Change the format of the exported aggregate losses


micheles:
We should extend the work of https://github.com/gem/oq-engine/pull/2684 to the other calculators, as Catalina wants.

micheles:
This is more difficult to do than it could be expected. For instance for scenario damage  we have two cases in the calculators, two exports, a data structure which has mean and stddev and it is not exactly the same as for scenario risk, all this multiplied three times (by asset, by taxonomy, total).

micheles:
Cata says that the by taxonomy and by asset csv are fine, the total_dmg could be improved instead.

PR #2791, opened 2017-05-10
Made more readable dmg_total.csv

micheles:
As requested by Cata in https://github.com/gem/oq-engine/issues/2768.. Here is an example of output:
https://github.com/gem/oq-engine/blob/daae3cd8fb0af23d6b156be4e0b8758fb90c4bf6/openquake/qa_tests_data/scenario_damage/case_7/expected/dmg_total.csv
It could be improved further, by using four columns `loss_type, damage_state, mean, stddev`, but I will leave that for the next release.


CatalinaYepes:
I will prefer to change the header "damage_kind" to "damage_state"

PR #2797, opened 2017-05-12
Fast report info

micheles:
`oq info --report` became slow in release 2.3 because I tried to give precise information about the exact number of ruptures involved in the computation. Here I went back to the original approach, i.e. using the fast source filter and not the slow rupture filter. The numbers will not be exact but good enough for an estimate. Also, I am removing the source prefiltering and the source splitting, so that the generation of the report becomes a lot faster.


openquake/calculators/views.py
```diff
@@ -582,8 +582,8 @@ def view_required_params_per_trt(token, dstore):
     for grp_id, gsims in gsims_per_grp_id:
         maker = ContextMaker(gsims)
         distances = maker.REQUIRES_DISTANCES
```
daniviga: Why this is not sorted like the two below?

micheles: Because I forgot to do it! Good catch.

PR #2800, opened 2017-05-16
Make it possible to specify multiple file names in <uncertaintyValue/>

micheles:
The multiple files are collected together sequentially and are managed as a single list of source groups.


openquake/commonlib/readinput.py
```diff
@@ -372,38 +372,41 @@ def get_source_models(oqparam, gsim_lt, source_model_lt, in_memory=True):
...
-                src_groups.extend(sourceconverter.SourceGroup.collect(smodel))
+        src_groups = []
+        for name in sm.name.split():
+            fname = possibly_gunzip(
+                os.path.abspath(os.path.join(oqparam.base_path, name)))
+            if in_memory:
+                apply_unc = source_model_lt.make_apply_uncertainties(sm.path)
+                try:
+                    logging.info('Parsing %s', fname)
+                    src_groups.extend(psr.parse_src_groups(fname, apply_unc))
```
mmpagani: There might be a case here where an error is generated. You are applying uncertainties separately to the various components (i.e. sources in the different files).  In case of an uncertainty applied to a specific source, that source might not be necessarily in the chunk of sources you are processing. Did I understand correctly?

micheles: You understand correctly, but trying to exercise this error situation I discovered that the validation of the logic tree is disabled for unknown reasons (it was already disabled 4 years ago at the time of engine 1.0). I assume in this case the uncertainty is not applied for the missing source, but in any case we should restore the validation or at least give a good reason for removing it. I am opening a ticket for that: https://github.com/gem/oq-engine/issues/2801

mmpagani: Leaving validation aside ...  LGTM

PR #2804, opened 2017-05-18
Remove the default filtering on date

daniviga:
It's not intuitive to have a 30 days limit on calculation list, and we already have a limit of last 100 calcs in the WebUI. For non-frequent users (like me) this could be an issue.


micheles:
LGTM

PR #2806, opened 2017-05-19
Fixed the command `oq export loss_curves/rlz-XXX`

micheles:
Now it works both for classical_risk and event_based_risk. Tests for the event based risk case have been added.

PS: while at it, I have also added a test for the command `oq export hcurves/rlz-XXX`.


openquake/commonlib/calc.py
```diff
@@ -184,6 +184,8 @@ def items(self, kind=''):
             for rlzi in range(num_rlzs):
                 hcurves = self.get(self.sids, rlzi)
                 yield 'rlz-%03d' % rlzi, hcurves
+        elif 'poes' in self.dstore and kind.startswith('rlz-'):
```
ptormene: Please add a test for this if needed

micheles: Done

Issue #2636, opened 2017-03-20
Generate PDF has broken references


daniviga:
In the last manual (and in previous version, with less occurrences) reference to pictures are broken (Fig. `??`)

daniviga:
The issue is probably related to the warnings we get from LaTeX
```
LaTeX Warning: Reference `fig:single_rupture' on page 18 undefined on input lin
LaTeX Warning: Reference `fig:single_rupture' on page 19 undefined on input lin
LaTeX Warning: Reference `fig:point_source_multiple_ruptures' on page 19 undefi
LaTeX Warning: Reference `fig:point_source_ruptures' on page 19 undefined on in
LaTeX Warning: Reference `page:point_source_nrml' on page 19 undefined on input
LaTeX Warning: Reference `page:point_source_nrml' on page 19 undefined on input
LaTeX Warning: Citation 'petersen2008' on page 19 undefined on input line 107.
LaTeX Warning: Citation 'frankel1995' on page 19 undefined on input line 107.
LaTeX Warning: Citation 'woo1996' on page 19 undefined on input line 107.
LaTeX Warning: Reference `subsubsec:point_sources' on page 20 undefined on inpu
LaTeX Warning: Citation 'schwartz1984' on page 23 undefined on input line 15.
LaTeX Warning: Citation 'aki2002' on page 23 undefined on input line 50.
LaTeX Warning: Citation 'aki2002' on page 23 undefined on input line 53.
LaTeX Warning: Citation 'chiou2014update' on page 23 undefined on input line 81
Package caption Warning: \captionsetup{type*=...} or \captionof
Package caption Warning: The option `hypcap=true' will be ignored for this
Package caption Warning: The option `hypcap=true' will be ignored for this
LaTeX Warning: Reference `fig:char_fault_source' on page 28 undefined on input 
LaTeX Warning: Citation 'aki2002' on page 29 undefined on input line 30.
Package caption Warning: The option `hypcap=true' will be ignored for this
Package caption Warning: The option `hypcap=true' will be ignored for this
Package caption Warning: The option `hypcap=true' will be ignored for this
Package caption Warning: The option `hypcap=true' will be ignored for this
Package caption Warning: The option `hypcap=true' will be ignored for this
Package caption Warning: The option `hypcap=true' will be ignored for this
LaTeX Warning: Reference `fig:evenly_discretized_mfd' on page 35 undefined on i
[...]

PR #2809, opened 2017-05-19
Fix manual PDF generation

daniviga:
Thanks to @nastasi-oq references in PDF are fixed.

Fixes https://github.com/gem/oq-engine/issues/2636

See https://ci.openquake.org/job/builders/job/pdf-builder/34/


micheles:
LVGTM!

PR #2814, opened 2017-05-23
Update manual todo list for release 2.4

raoanirudh:


micheles:
I think there is not CSV output for the aggregate losses yet. The reason is that we plan to change the algorithm in the next release so I wanted to wait before doing that. For the rest, LGTM.

PR #2816, opened 2017-05-24
Updates to the engine manual (risk) for release 2.4

raoanirudh:
Highlights of the changes:

- Updated version number from 2.3 to 2.4
- Updated publication date and citation
- Updated authors to include @rcgee 
- §9.1: Updated the scenario damage datastore output names
- §9.1.1.1: Replaced the asset damage xml example with csv
- §9.1.1.2: Replaced the taxonomy damage xml example with csv
- §9.1.1.3: Replaced the total damage xml example with csv
- §9.1.2: Removed the collapse map xml example
- §9.2: Updated the scenario risk output names
- §9.2.2: Replaced the scenario loss map xml example with csv

The default export file format for all loss curves outputs is still xml, so I propose maintaining the status quo for the examples for those outputs until the csv versions are stable enough to replace them as the default.


raoanirudh:
PDF build for reference: [oq-manual.pdf](https://github.com/gem/oq-engine/files/1025358/oq-manual.pdf)

micheles:
There is a line in the manual
```
(a list of available GMPEs are given in Appendix ??)
```
I would change that line as
```
(the list of available GSIMs can be extracted with the command oq info --gsims)
```

micheles:
This branch was detached from master, so it is no good. I am reopening the PR detaching from engine-2.4

Issue #2793, opened 2017-05-11
Restore the classical_tiling calculator


micheles:
8 months ago we removed the tiling calculator for sound reasons (see https://github.com/gem/oq-engine/pull/2253). However, now we have the `shared_dir` functionality, which has changed the game completely The problem of the tiling was the data transfer, since the same sources had to be transferred multiple times. Now that we have the shared_dir, the source model can be read directly from the workers, thus bypassing the celery/rabbitmq combo. This looks very attractive, especially if the points sources are stored efficiently in .hdf5 format and become fast to read. A tiling calculator would allow for further optimizations, such as avoiding recomputing multiple times the same sources, as it happens in the Canada model. Also, a worker-based new tiling calculator could avoid filtering the sources in the controller node, which was slow. In short, it makes sense to restore it now.

PR #2821, opened 2017-05-26
Restored the tiling functionality

micheles:
To enable the tiling feature set the parameter `num_tiles` in the job.ini file. Notice that tiling is experimental and often slower than the regular calculator, so it best not to use it, unless I say otherwise (there are cases where it is better than the regular calculator). The performance depends strongly on the number of tiles generated and it is difficult to get it right, this is why this feature is experimental. In the future it will hopefully improve. If for instance `num_tiles=10` you should understand that:

- the data transfer due to the sources will be 10 times worse
- the data transfer due to the site collection will be 10 times better
- the data transfer due to the probability maps will be better
- the overhead when iterating over a fault rupture will be 10 times worse
- the filtering overhead will be worse
- the make_context part i.e. distance calculations will be better (shorter arrays)
- the dominating factor will often be the presence of slow tasks, i.e. unsplit sources

Depending on the computation parameters any kind of performance result is possible (can be better, can be worse; because of Murphy's law it will be usually worse).

Closes https://github.com/gem/oq-engine/issues/2793

### DIFF
--- a/openquake/qa_tests_data/classical_tiling/case_1/job.ini
+++ b/openquake/qa_tests_data/classical_tiling/case_1/job.ini
@@ -2,8 +2,7 @@
 
 description = Classical PSHA using Area Source
 calculation_mode = classical
-sites_per_tile = 1
-concurrent_tasks = 4
+num_tiles = 3
 random_seed = 23
 
 [geometry]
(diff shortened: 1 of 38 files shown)


openquake/qa_tests_data/classical_tiling/case_1/job.ini
```diff
@@ -2,8 +2,7 @@
 
 description = Classical PSHA using Area Source
 calculation_mode = classical
-sites_per_tile = 1
-concurrent_tasks = 4
+num_tiles = 3
```
ptormene: Perhaps we should re-add concurrent_tasks (here and in case_2)

PR #2824, opened 2017-05-28
oq webui start tries to open the browser

daniviga:
`oq webui start` now tries to open a browser tab pointing to the WebUI.

The browser starts only when the WebUI is up and running and responding to the requests. On headless servers this does not change normal behavior; the open browser feature can be also turned off passing the `--skip-browser` flag to the command:

`oq webui start --skip-browser`

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,7 @@
+  [Daniele Viganò]
+  * Now the command `oq webui start` tries to open a browser tab
+    with the WebUI loaded
+
 python-oq-engine (2.4.0-0~precise01) precise; urgency=low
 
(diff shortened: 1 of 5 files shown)


debian/changelog
```diff
@@ -1,3 +1,9 @@
+python-oq-engine (2.5.0-0~precise01) precise; urgency=low
+
+  [Daniele Viganò]
+  * Now the command `oq webui start` tries to open a browser tab
+    with the WebUI loaded
+
```
micheles: The changelog message should go *above* the line `python-oq-engine (2.5.0-0~precise01) precise; urgency=low``

PR #2829, opened 2017-05-29
Add more commands to webui plus some fixes to doc

### LINKED ISSUE #2827: local_settings.py.pam is not included in debs
The `local_settings.py.pam` template is missing from the debian package

daniviga:
This PR adds more commands to the WebUI command because those commands are useful when packages are used and the `python -m` way is hard to use (because of `/opt` and the `PYTHONPATH`).

Improves also the documentation

Closes #2827


micheles:
Please update the changelog. For the rest, LGTM.

Issue #2815, opened 2017-05-23
unicode error in filenames


CatalinaYepes:
Example of error message

```
2017-05-23T14:52:00.33,INFO,MainProcess/2749,Reading the exposure
2017-05-23T14:52:00.81,INFO,MainProcess/2749,Read 1512 assets on 411 sites
2017-05-23T14:52:00.83,CRITICAL,MainProcess/2749,
2017-05-23T14:52:00.84,CRITICAL,MainProcess/2749,
2017-05-23T14:52:00.85,CRITICAL,MainProcess/2749,Traceback (most recent call last):
  File "/Users/catalinayepes/OpenQuake/oq-engine/openquake/engine/engine.py", line 198, in run_calc
    _do_run_calc(calc, exports, hazard_calculation_id, **kw)
  File "/Users/catalinayepes/OpenQuake/oq-engine/openquake/engine/engine.py", line 231, in _do_run_calc
    close=False, **kw)  # don't close the datastore too soon
  File "/Users/catalinayepes/OpenQuake/oq-engine/openquake/calculators/base.py", line 200, in run
    self.pre_execute()
  File "/Users/catalinayepes/OpenQuake/oq-engine/openquake/calculators/scenario_damage.py", line 144, in pre_execute
    base.RiskCalculator.pre_execute(self)
  File "/Users/catalinayepes/OpenQuake/oq-engine/openquake/calculators/base.py", line 439, in pre_execute
    self.precalc = (self.compute_previous() if precalc_id is None
  File "/Users/catalinayepes/OpenQuake/oq-engine/openquake/calculators/base.py", line 375, in compute_previous
    precalc.run(close=False)
  File "/Users/catalinayepes/OpenQuake/oq-engine/openquake/calculators/base.py", line 200, in run
    self.pre_execute()
  File "/Users/catalinayepes/OpenQuake/oq-engine/openquake/calculators/scenario.py", line 39, in pre_execute
    super(ScenarioCalculator, self).pre_execute()
  File "/Users/catalinayepes/OpenQuake/oq-engine/openquake/calculators/base.py", line 444, in pre_execute
    self.basic_pre_execute()
  File "/Users/catalinayepes/OpenQuake/oq-engine/openquake/calculators/base.py", line 398, in basic_pre_execute
    self.read_risk_data()
  File "/Users/catalinayepes/OpenQuake/oq-engine/openquake/calculators/base.py", line 541, in read_risk_data
[...]

PR #2830, opened 2017-05-29
Fixed an encoding error in a log.warn

micheles:
Closes https://github.com/gem/oq-engine/issues/2815 (hopefully: we must test it with Python 2 on Windows and Mac).


micheles:
Here is the file to exercise the bug:
[CITYC.zip](https://github.com/gem/oq-engine/files/1038318/CITYC.zip)

Issue #2731, opened 2017-04-21
Make it possible to specify the kind of disaggregation output to export


micheles:
For instance on could write in the job.ini something like

```
   disagg_outputs = Mag Dist TRT Mag_Dist Lon_Lat Mag_Lon_Lat
```

where the set of valid kinds is limited (there are at most 8 choices).
This was requested by Laurentiu and Marco to reduced the number of the exported CSV files.
It would be nice to reduce the computation time if fewer disaggregation outputs are requested, but this is difficult and will be left for the future.

PR #2831, opened 2017-05-29
disagg_outputs feature

micheles:
Companion of https://github.com/gem/oq-hazardlib/pull/646. Closes https://github.com/gem/oq-engine/issues/2731.


rcgee:
LGTM

Issue #2713, opened 2017-04-10
Remove loss_curves and loss_maps XML exporters


micheles:
This was requested by @raoanirudh . If somebody is using those exporters we should deprecate them for one release before removal. If nobody is using them (as it seems) we can just drop them.

micheles:
Also remove the outdated examples from `nrml_examples`

PR #2832, opened 2017-05-29
Removed the deprecated XML outputs

micheles:
Closes https://github.com/gem/oq-engine/issues/2713.


micheles:
Yes, since they are not used.

Issue #2534, opened 2017-02-07
We should not perform logic tree reduction based on the ruptures in the event based calculators


micheles:
Otherwise the logic tree will depend on the seed and the investigation time. Logic tree reduction based on the maximum distance is fine instead, according to M. Pagani.

PR #2840, opened 2017-06-02
Removed excessive logic tree reduction in event based

micheles:
Closes https://github.com/gem/oq-engine/issues/2534. Now the logic tree realizations are:

1. independent from the investigation time and the random seed
2. exactly identical to the ones in an equivalent classical calculation
3. computable in a relatively fast way with `oq info --report`

If a source model generates zero ruptures, it will still contribute to the realizations with 0 values: the net effect is that the statistics are different. For instance, if there are 4+4 realizations, with the last 4 having zero effect (i.e. two source models with the second one generating zero ruptures within the given investigation time and with the given seed), then when computing the mean one has to divide by 8, not by 4. Since our tests were specifically testing such corner cases, most of them had broken statistic. The situation is even worse for event based calculations: in that case not only the statistics but also the individual curves are different if the logic tree reduction is different. The reason is that the ground motion seeds are associated to the realizations: more realizations -> more seeds -> different individual curves.
In real calculations, 99.999% of the times there are more than 0 ruptures, so this is not a serious issue in practice. The previous behavior was intended: it was meant to reduce the logic tree as much as possible for performance reasons; however, it was too surprising and now that the engine is fast enough it is better to do the less surprising thing.

NB: the logic tree is still dependent on the hazard sites. Suppose there is a source that affects a single site at a distance of 199.999 km from it. By moving the source or the site by 1 meter, with a maximum distance of 200 km, the source will be discarded and the corresponding tectonic region will be discarded too, if the source if the only one of that type: then the logic tree will be reduced and the number of realizations will change, with a big effect on all statistics. While surprising, this has to be considered acceptable, otherwise we should through away the concept of logic tree reduction which will entail other very serious drawbacks, i.e. performance, storage and duplicated outputs.

The tests are running here: https://ci.openquake.org/job/zdevel_oq-engine/2512


micheles:
I have updated the documentation. Feel free to copy the relevant bits to the manual.

Issue #2805, opened 2017-05-18
Add to the engine server API a method to get the parameters of a calculation


ptormene:
It would be useful, for instance, when styling a hazard map in QGIS, to set the size of points depending on their distance in the grid.

PR #2844, opened 2017-06-05
Added an end point `v1/calc/XXX/oqparam` to extract the calculation parameters as a JSON dictionary

micheles:
Closes https://github.com/gem/oq-engine/issues/2805


micheles:
After some checking, we discover that this is an issue of Firefox. The error 404 is correctly raised, Chrome is showing a 404 error page, but Firefox is hiding it and showing an empty page. I would do nothing.

PR #2847, opened 2017-06-06
Removed the obsolete risk writers

micheles:
Second part of https://github.com/gem/oq-engine/issues/2713. I have added CSV exporters for the agg_curve outputs instead of the deprecated XML ones.


openquake/qa_tests_data/event_based_risk/case_1/agg_curve-rlz-000_6879.csv
```diff
@@ -0,0 +1,41 @@
+loss_type,loss,poe
```
ptormene: Should the final number of the filename (6879) be removed? Please check about it here and for the other output files.

PR #2848, opened 2017-06-06
Restored the parameter sites_per_tile

micheles:
Tiling starts automatically when there are more than `sites_per_tile` hazard sites (default 20,000).


klunk386:
LGTM

Issue #2852, opened 2017-06-06
Added a special case for MultiMFDs with homogeneous parameters


micheles:
This is the most common case, as the hazard team pointed out. We should save space in the XML.
As an example, take the source model for Argentina, hazard_models_v1/sara_source_model_1_arg.xml.

PR #2856, opened 2017-06-08
Changed the MultiMFD class to manage homogeneous inputs

micheles:
Closes https://github.com/gem/oq-engine/issues/2852.


klunk386:
Given the example in openquake/qa_tests_data/classical/case_28/GridSources.xml, it should be now possible to specify a single (constant) value in a xml when using MultiMFD.
The method write_source_model should therefore produce a consistent xml output when a same value is used for all the points. Presently, an array of identical values is produced.

micheles:
Can you send me a script exhibiting the problem?

klunk386:
I tested the modifications with the North Africa model. Everything works as expected.

klunk386:
All new calculations from the engine are OK.  LGTM

Issue #2861, opened 2017-06-08
Collects the exported data into a single array for hazard outputs


micheles:
This makes it easier to manage them in QGIS. In practice, instead of .npz exports, we will have .npy exports for hcurves, hmaps and uhs. Each export will correspond to a single QGIS layer containing all the hazard information without duplication. This is possible now because we do not export anymore the full realizations, only the statistics, i.e. a reasonable about of data. The .npy files will never be too big.

Issue #2568, opened 2017-02-20
.npz outputs should contain investigation time


g-weatherill:
The .npz files for storing hazard output do not contain information about the investigation time set for the calculation. This should be added.

PR #2862, opened 2017-06-09
Changed the hazard exports to use one column per each IML

micheles:
Closes https://github.com/gem/oq-engine/issues/2861 and https://github.com/gem/oq-engine/issues/2568. Before merging, we have the see the impact on the QGIS plugin and the Colombia project.
NB: I have added the vs30 map to the hazard map output, because later on we may want to visualize it.
Here is an example of use of the hazard curves:

```python
In [1]: import numpy
In [2]: npy = numpy.load('hcurves_27521.npy')
In [3]: npy.dtype.names
Out[3]: ('lon', 'lat', 'rlz-000')   # or ('lon', 'lat', 'mean', 'quantile-0.15', 'quantile-0.85') etc
In [4]: npy.dtype['rlz-000'].names  # extracts the IMTs
Out[4]: 
('PGA',
 'PGV',
 'SA(0.025)',
 'SA(0.05)',
 'SA(0.1)',
 'SA(0.2)',
 'SA(0.5)',
 'SA(1.0)',
 'SA(2.0)')
In [5]: npy.dtype['rlz-000']['PGA'].names  # extract the IMLs to be used for plotting
Out[5]: 
('0.005',
 '0.007',
 '0.0098',
 '0.0137',
 '0.0192',
 '0.0269',
 '0.0376',
 '0.0527',
 '0.0738',
 '0.103',
 '0.145',
 '0.203',
 '0.284',
 '0.397',
 '0.556',
 '0.778',
 '1.09',
 '1.52',
 '2.13')
```

There is an advantage in using .npy, since the client can read the array directly with a function like the following:

```python
def get_array(url):
      resp = requests.get(url, dict(export_type='npy'))
      return numpy.load(io.BytesIO(resp.content))
```


daniviga:
I'm re-running my calculation for Colombia to test it. It will require some hours.

micheles:
It is useless now Daniele. First Paolo has to change the QGIS side, only after you can look at this.

daniviga:
I need anyway to re-run the calculation since it was made quite along ago with an old version of the engine. And I want to see how fast is the export (which I guess I can already run manually)

PR #2872, opened 2017-06-12
Don't force the isa in sub packages

### LINKED ISSUE #2871: CentOS7 and dependency
Please review dependency as :
```
yum install python-oq-engine-worker

Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * base: mirrors.td512.io
 * epel: mirror.xnet.co.nz
 * extras: mirrors.td512.io
 * updates: mirrors.td512.io
Resolving Dependencies
--> Running transaction check
---> Package python-oq-engine-worker.noarch 0:2.4.0-1 will be installed
--> Processing Dependency: python-oq-engine(x86-64) = 2.4.0-1 for package: python-oq-engine-worker-2.4.0-1.noarch
--> Finished Dependency Resolution
Error: Package: python-oq-engine-worker-2.4.0-1.noarch (gem-openquake-stable)
           Requires: python-oq-engine(x86-64) = 2.4.0-1
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
```

Of course there is no package for x86-64 arch, only a noarch package.

daniviga:
There is a slight different between packages build on COPR and locally by mock. This was causing an error when a subpackage, from the official stable repo, was installed. This was not caught by tests using locally built packages.

Fixes #2871

### DIFF
--- a/rpm/python-oq-engine.spec.inc
+++ b/rpm/python-oq-engine.spec.inc
@@ -48,7 +48,7 @@ Copyright (C) 2010-2017 GEM Foundation
 %package master
 Summary: OpenQuake Engine multi-node cluster support (master node)
 Group: Applications/Engineering
-Requires: %{name}%{?_isa} = %{version}-%{release} rabbitmq-server
+Requires: %{name} = %{version}-%{release} rabbitmq-server
 
 %description master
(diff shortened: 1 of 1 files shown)


daniviga:
LGTM

Issue #2518, opened 2017-02-03
Ask for confirmation when deleting calc in webui


daniviga:
See #2514 

In the modal we should ask for confirmation

micheles:
Thinking a bit more about this issue, I realized that we do not need a confirmation dialog. When the user wants to delete a calculation? When it failed, otherwise she has little motive to remove it. So here is my proposal:

1. if the calculation failed, there is a "remove" button and it works without confimation
2. if the calculation did not fail, there is no "remove" button: the only way to remove the calculation is from the command-line.

This should be easy to implement, since the Javascript knows if a calculation failed or not (this info is used in the colors, red for failure, green for success).

daniviga:
mmm I've seen the delete button used quite often during workshops to clean up successful calculations too, and during wss scientists tends to don't use the command line (and event to not explain in at all) so probably we have to keep it.

PR #2874, opened 2017-06-12
Confirm Dialog before remove calculation

hascar:
Confirm Dialog before remove calculation

Issue https://github.com/gem/oq-engine/issues/2518
tests are green here: https://ci.openquake.org/job/zdevel_oq-engine/2518/ on 14.04

### DIFF
--- a/openquake/server/static/js/engine.js
+++ b/openquake/server/static/js/engine.js
@@ -448,7 +448,7 @@ var CalculationTable = Backbone.View.extend(
             calculations.fetch({reset: true});
             setTimer();
 
-            $(".btn-danger").click(function (e) {
+            $(document).on("click", 'button[class=btn-danger]',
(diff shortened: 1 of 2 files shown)


daniviga:
May I suggest to make a new branch/PR with a couple of meaningful commits? Here we have lot of commits with almost the same comment.
@micheles what's you opinion?

hascar:
Superseded by #2876

micheles:
No problems with the what's new: I never use the commit messages for it. Let's keep the commits as they are.

daniviga:
@micheles I'm fine keeping all the original commits if they are a kind of meaningful. I don't like then having one hundred commits in the history with mostly the same commit message. It would be then hard to come back to real history if we add so many commits. We used the same approach on oq-platform and derivates.

PR #2881, opened 2017-06-13
Add a zipped copy of demos in the packages

### LINKED ISSUE #2822: Provide a zipped copy of each demo in packages
Like we do in the standalone installers. Those are needed by the webui.

A zips can be easily made during installation (to save space) or during package creation.

daniviga:
As we do in standalone installers. This has been asked by the users to make use of the WebUI more easy.

Zips are made at build time since all the zipped demos add less than 2MB to the final size. If the size will increase we can move this process into `post_install`.

https://ci.openquake.org/job/zdevel_oq-engine/2527/

Closes #2822

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: python-oq-engine
 Section: python
 Priority: extra
 Maintainer: Matteo Nastasi (GEM Foundation) <nastasi@openquake.org>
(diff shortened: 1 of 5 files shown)


micheles:
This a good idea, LGTM.

PR #2884, opened 2017-06-13
Added commands `oq dump` and `oq restore`

micheles:
Here is an example of use:

```bash
$ oq dump /tmp/oqdata.zip
...
Archiving /mnt/ssd/oqdata/calc_29304.hdf5
Archiving /mnt/ssd/oqdata/calc_29305.hdf5
Archiving /home/michele/oqdata/db.sqlite3
Archived 193 files into /tmp/x.zip in 212 seconds
WARNING: there were calculations executing during the dump
They have been not copied; please check the correctness of the db.sqlite3 file
3423 executing Virtual Island Seismic Risk - City C, ses=1
21895 executing LEONARD BACKGROUND RUN - NATIONAL MAPS
$ oq restore /tmp/oqdata.zip /tmp/oqdata
$ OQ_DATADIR=/tmp/oqdata oq webui start
```

### DIFF
--- a/openquake/commands/dump.py
+++ b/openquake/commands/dump.py
@@ -17,6 +17,7 @@
 #  along with OpenQuake.  If not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import print_function
+import time
 import os.path
 from openquake.baselib import sap
 from openquake.server.manage import db
@@ -29,11 +30,14 @@ def dump(archive):
     Dump the openquake database and all the complete calculations into a zip
     file. In a multiuser installation must be run as administrator.
     """
+    t0 = time.time()
     assert archive.endswith('.zip'), archive
     query = 'select ds_calc_dir || ".hdf5" from job where status="complete"'
     fnames = [fname for fname, in db(query) if os.path.exists(fname)]
     fnames.append(db.path)
     zipfiles(fnames, archive, print)
-    print('Archived %d files into %s' % (len(fnames), archive))
+    dt = time.time() - t0
+    print('Archived %d files into %s in %d seconds'
+          % (len(fnames), archive, dt))
 
 dump.arg('archive', 'path to a zip file')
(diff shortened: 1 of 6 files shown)


openquake/commands/dump.py
```diff
@@ -0,0 +1,38 @@
...
+@sap.Script
+def dump(archive):
+    """
+    Dump the openquake database and all datastores into a archive zip file.
+    In a multiuser installation must be run as administrator.
+    """
+    assert archive.endswith('.zip'), archive
+    fnames = [fname for fname, in db('select ds_calc_dir || ".hdf5" from job')
+              if os.path.exists(fname)]
+    fnames.append(db.path)
```
daniviga: We should stop (and restart?) the dbserver here to make sure that the dump is consistent.

daniviga: I would suggest to check if the DB is running and if that's the case:
- to stop it if `multi_user = False` 
- to exit with a message telling the user to stop the dbserver first if `multi_user = True`

This approach is consistent to what we do elsewhere.

micheles:
Actually the dump script works perfectly well even if the dbserver is on, if nobody is running a calculation. It probably works even if somebody is running a calculation. The copy of that computation will not be complete but the other datastores would be copied fine. I would say that it is the sysadmin responsibility to make sure that nobody is running a calculation during the dump.

micheles:
I have changed the query so that only "complete" calculations are copied. It is safe to copy such files even if there are calculations running. Copying the db.sqlite file while processes are writing on it can produce a corrupted file, at least in theory, but I was not able to produce one.

daniviga:
As we were discussing we can do the follow:

- Always dump the database (in bot single and multi user mode) without stopping the DbServer
- Check the DB first, if there's any calculation 'executing', print a warning on log saying that the backup was done when a computation was running and the sqlite DB may be unclean/corrupted.

micheles:
Done

daniviga:
Isn't this already part of #2880 ?

PR #2888, opened 2017-06-14
Allow installation of Django 1.11 LTS

daniviga:
Django 1.11 it's an LTS (see https://www.djangoproject.com/weblog/2015/jun/25/roadmap/) so it's the best version to be used as our ref. We are already 100% compatible with it.


micheles:
Please update the changelog

PR #2890, opened 2017-06-14
Do not propagate Django logs into the DbServer

### LINKED ISSUE #2889: Django log messages are pushed in the DB when django >= 1.10
@ptormene discovered that when django >= 1.10 is used the django log messages are pushed to the dbserver too (!!).

daniviga:
Fixes #2889 

This feature has been introduced in Django 1.10: https://docs.djangoproject.com/en/1.11/topics/logging/#django-server


micheles:
Please update the changelog

PR #2902, opened 2017-06-19
Simplified the NoFlush mechanism

micheles:
The engine has a mechanism to forbid the call to `Monitor.flush` in the workers, since the workers cannot write on the file system of the controller node. Such mechanism is more complex than needed, here I remove 16 lines of code.


openquake/baselib/parallel.py
```diff
@@ -235,7 +235,7 @@ def safely_call(func, args, pickle=False, conn=None):
         else:
             mon = child
         check_mem_usage(mon)  # check if too much memory is used
-        mon.flush = NoFlush(mon, func.__name__)
+        mon._flush = False
```
ptormene: Please add a comment explaining that this mechanism does not work with threads

Issue #2611, opened 2017-03-07
If deleting a calculation fails. the error message is not shown to the user


micheles:
Try to delete a calculation owned by another user.  You will get the right error message in the logs of the server, but the user will see a cryptic `Failed: error`.

ptormene:
The same happens when you try to delete a calculation, and the deletion is refused because there are other calculations that are based on it.

PR #2905, opened 2017-06-19
Fix webui error remove other user

hascar:
Fix webui error remove other user
Tests are green here: https://ci.openquake.org/job/zdevel_oq-engine/2535/
Closes https://github.com/gem/oq-engine/issues/2611


openquake/server/db/actions.py
```diff
@@ -268,31 +268,28 @@ def del_calc(db, job_id, user):
...
-        except OSError as exc:  # permission error
-            print('Could not remove %s: %s' % (fname, exc))
-        else:
-            fnames.append(fname)
-    return fnames
+    fname = path + ".hdf5"
+    try:
+        os.remove(fname)
+    except OSError as exc:  # permission error
+        print('Could not remove %s: %s' % (fname, exc))
```
micheles: Here we must return

```python
{"error": 'Could not remove %s: %s' % (fname, exc)}
```

openquake/server/static/js/engine.js
```diff
@@ -187,23 +187,27 @@
...
-                         }
-                     ).error(
-                         function(jqXHR, textStatus, errorThrown)
-                         {
-                             if (jqXHR.status == 404) {
-                                 diaerror.show(false, "Removing calculation", "Removal command for:<br><b>(" + calc_id + ") " + calc_desc + "</b> failed.");
+                         {  
+                             err = data.error; 
+                             if(err) {
+                                 err = data.error;
```
micheles: This `if` is useless.

openquake/server/db/actions.py
hascar: Fixed here https://github.com/gem/oq-engine/pull/2905/commits/85e73703cf664c94f5c6b8e55ff35093d64b1be1

micheles:
There are also several PEP8 errors; run

```bash
$ flake8 server/db/actions.py ... <all the change .py files here>
```

openquake/server/static/js/engine.js
hascar: Fixed here https://github.com/gem/oq-engine/pull/2905/commits/afb647f09e998c9d253bafc5cdfa9e62fab6eb76

openquake/server/db/actions.py
```diff
@@ -268,31 +268,30 @@ def del_calc(db, job_id, user):
...
-        except OSError as exc:  # permission error
-            print('Could not remove %s: %s' % (fname, exc))
-        else:
-            fnames.append(fname)
-    return fnames
+    fname = path + ".hdf5"
+    try:
+        os.remove(fname)
+    except OSError as exc:  # permission error
+        {"error": 'Could not remove %s: %s' % (fname, exc)}
```
micheles: Where is the `return` ??

hascar: Forgot it, forgive me.
Added here https://github.com/gem/oq-engine/pull/2905/commits/df1a92af8f8acfce0d5d94b8f40b06a279899ec2

openquake/server/views.py
```diff
@@ -326,13 +326,8 @@ def calc_remove(request, calc_id):
...
         return HttpResponseNotFound()
-    if isinstance(message, list):  # list of removed files
-        return HttpResponse(content=json.dumps(message),
-                            content_type=JSON, status=200)
-    else:  # FIXME: the error is not passed properly to the javascript
-        logging.error(message)
-        return HttpResponse(content=message,
-                            content_type='text/plain', status=500)
+    return HttpResponse(content=json.dumps(message),
+                        content_type=JSON, status=200)
```
daniviga: If the REST request fails because calculation is owned by another user it should return a 403 instead (IMHO):

```
403 - Forbidden

The server understood the request, but is refusing to fulfill it. Authorization will not help and the request SHOULD NOT be repeated. If the request method was not HEAD and the server wishes to make public why the request has not been fulfilled, it SHOULD describe the reason for the refusal in the entity.  If the server does not wish to make this information available to the client, the status code 404 (Not Found) can be used instead.
```

micheles:
Unfortunately, it seems that Backbone has no access to the request data if the status is different from 200, this is why we went that way.

daniviga:
That's strange. There are several examples on the web with backbone using 403. Anyway, I would not block the PR for this.

daniviga:
When I delete a calculation that does not belongs to me, I get the error, but the calculation has disappeared from the list. It comes back when I refresh the page.

It seems that the refresh loop is kept of after the modal. 

First modal, the calculation is still there (in background)
screenshot from 2017-06-20 09-08-37

Second modal (error) the calculation is gone (should not)
screenshot from 2017-06-20 09-08-50

Closing the modal the calculation is still gone (should not)
screenshot from 2017-06-20 09-09-00

After a manual refresh the calculation is back (correctly displayed)
screenshot from 2017-06-20 09-09-10

hascar:
Fixed here: https://github.com/gem/oq-engine/pull/2905/commits/9739593bd8f7f5a165897972edec2477cbff126c

daniviga:
This is an example on how to manage `403`:

```javascript
                var myXhr = $.ajax({url: gem_oq_server_url + "/v1/calc/" + calc_id + "/remove",
                                    type: "POST",
                                    error: function (jqXHR, textStatus, errorThrown) {
                                        if (jqXHR.status == 403) {
                                            diaerror.show(false, "Error:", JSON.parse(jqXHR.responseText).error);
                                        }
                                    },
                                    success: function(data, textStatus, jqXHR) {
                                        err = data.error;
                                        if(!err) {
                                            err = "removed.";
                                        }
                                        var hide_or_back = (function(e) {
                                            closeTimer();
                                            this.conf_hide = $('#confirmDialog' + calc_id).hide();
                                            this.back_conf_hide = $('.back_confirmDialog' + calc_id).hide();
                                            setTimer();
                                        })();
                                        diaerror.show(false, "Calculation removed", "Calculation:<br><b>(" + calc_id + ") " + calc_desc + "</b> " + err );
                                        view.calculations.remove([view.calculations.get(calc_id)]);
                                    }});
            },
```

Code is buggy and needs cleanup, but the main part does what we want. I'm using the verbose way, instead of the shorthand (```.post()```) for readability. 

ref: https://api.jquery.com/jquery.post/

hascar:
I tried it now and it does not work because with error 403 is not possible to display any message
https://github.com/gem/oq-engine/pull/2905/commits/85c547647e6c0f6c5d23dc493894202a89c121e7

daniviga:
The code pushed is actually wrong. I work on it later, I'm in a call right now.

openquake/server/views.py
```diff
@@ -334,7 +336,6 @@ def calc_remove(request, calc_id):
         return HttpResponse(content=message,
                             content_type='text/plain', status=500)
```
daniviga: This should be `403` not 500.

hascar: But did you read my comment and read my commit here? https://github.com/gem/oq-engine/commit/85c547647e6c0f6c5d23dc493894202a89c121e7 
There is error 403 and not 500

daniviga: See https://github.com/gem/oq-engine/pull/2911

micheles:
There is one thing I do not like. `db.actions.del_calc` may return different kind of errors ("ID does not exists", "There are dependent calculations", "You are not the owner", "OSError: you cannot remove") and they all end up to be errors 403. This is not much better that returning always 200 that at least had the advantage of removing the code duplication on the javascript side. Also, there is branch

```python
   else:
        # This is an untrapped server error
          logging.error(message)
￼          return return HttpResponse(content=message,
￼                              content_type='text/plain', status=500)	
```
that cannot never happen, given how db.actions.del_calc is implemented.  I am not sure what to do, there should be best practices for errors that do not much HTTP error codes, let's the PR wait and think again after the governing board.

daniviga:
> ("ID does not exists", "There are dependent calculations", "You are not the owner", "OSError: you cannot remove") and they all end up being errors 403.

'ID does not exists' may be a 404 but the other three are correctly 403. 

> 403 - Forbidden
> The server understood the request, but is refusing to fulfill it. Authorization will not help and the request SHOULD NOT be repeated. 

> This is not much better that returning always 200 that at least had the advantage of removing the code duplication on the javascript side.
 
Why? For a REST API 200 and 403 have specific meanings. 200 reports a successfully operation. The ones above aren't. The duplication in the javascript are _only_ in the timer management that can be easily refactored with a function and are not related to the error management.

The block about 500 can be removed, it's only to pass a plain text message to the client (but this in currently unmanaged on client side). We can just keep the error in the logs.

daniviga:
I see that 'ID does not exists' is already managed as 404 from the beginning of time.

micheles:
It is the "There are dependent calculations" error that I did not see as a 403 error (it is the violation of a ForeignKey constraint). However, I think it is not worth the effort of debating such details. Your point that we want to have an error code in the server logs (so no 200) is well taken. I do not really like having an error 500 and the 403 is probably the lesser of the evils.

LGTM.

PR #2909, opened 2017-06-20
Changed the message printed by `oq dump`

micheles:
Before it was "Archived 16 files" (correct, 15 calculations + the db.sqlite3 file), now it is "Archived 15 calculations" which is more consistent with the message printed by `oq restore` that only mention calculations.

PS: I am also fixing run-demos to run the 17th demo with the engine, not with oq-lite.

### DIFF
--- a/bin/run-demos.sh
+++ b/bin/run-demos.sh
@@ -26,4 +26,4 @@ MPLBACKEND=Agg python -m openquake.commands plot_uhs -3
 # fake a wrong calculation still in executing status
 python -m openquake.commands db set_status 1 executing
 # repeat the failed/executing calculation, which is useful for QGIS
-python -m openquake.commands run $1/hazard/AreaSourceClassicalPSHA/job.ini
+python -m openquake.commands engine --run $1/hazard/AreaSourceClassicalPSHA/job.ini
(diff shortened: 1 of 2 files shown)


ptormene:
I still don't understand why, according to https://travis-ci.org/gem/oq-engine/jobs/244881723

* 17 calculations are completed
* the first one is not exported (so we still have 16)
* in the end we have only 15 exported calculations (from 2 to 16), instead of 16 (from 2 to 17)

ptormene:
It looks like you are archiving things before running the 17th calculation, so it will not be included in the dump (and the first one was discarded)

micheles:
No, I am archiving after, it is a total mystery why the 17th calculation is not archive. Perhaps it is an unlucky number!

daniviga:
We should compare http://artifacts.openquake.org/demos/dump.zip with http://artifacts.openquake.org/travis/oqdata.zip

`dump.zip` has been used by Paolo and I with our first integration tests and it was working fine. It has been manually generated by me.

daniviga:
It looks like that if there are two calculations which are the same, but one is marked as `executing`, both are skipped by `oq dump`

micheles:
Solved. I  was trivial.
```
python -m openquake.commands run $1/hazard/AreaSourceClassicalPSHA/job.ini
```
has to be replaced by

```
python -m openquake.commands engine --run $1/hazard/AreaSourceClassicalPSHA/job.ini
```

daniviga:
Good catch!

PR #2913, opened 2017-06-21
Fixing precision error in gsim_table.py

treviallen:
Minor edit to fix a rounding error in gsim_table.py in the implementation of NGA-East GMM tables

### DIFF
--- a/openquake/hazardlib/gsim/gsim_table.py
+++ b/openquake/hazardlib/gsim/gsim_table.py
@@ -525,7 +525,7 @@ class GMPETable(GMPE):
             low_period = round(periods[0], 7)
             high_period = round(periods[-1], 7)
 
-            if imt.period < low_period or imt.period > high_period:
+            if round(imt.period, 7) < low_period or round(imt.period, 7) > high_period:
                 raise ValueError("Spectral period %.3f outside of valid range "
                                  "(%.3f to %.3f)" % (imt.period, periods[0],
                                                      periods[-1]))
--- a/openquake/hazardlib/gsim/yenier_atkinson_2015.py
+++ b/openquake/hazardlib/gsim/yenier_atkinson_2015.py
@@ -1,0 +1,559 @@
+# -*- coding: utf-8 -*-
+
+# The Hazard Library
+# Copyright (C) 2012-2014, GEM Foundation
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+Module exports :class:`YenierAtkinson2015`.
+"""
+from __future__ import division
+
+import numpy as np
+
+from openquake.hazardlib.gsim.base import GMPE, CoeffsTable
+from openquake.hazardlib import const
+from openquake.hazardlib.imt import PGA, PGV, SA
+
+
+class YenierAtkinson2015(GMPE):
+        """
+        Implements GMPE developed by Emrah Yenier and Gail M. Atkinson
+        and published as "Regionally Adjustable Generic Ground-Motion Prediction 
+        Equation Based on Equivalent Point-Source Simulations: Application to Central 
+        and Eastern North America" (2015, Bulletin of the Seismological Society of America,
+        Volume 105, No. 4, pages 1989-2009).
+        """
+        
+        #: Define for tectonic region type stable shallow crust. 
+        DEFINED_FOR_TECTONIC_REGION_TYPE = const.TRT.STABLE_CONTINENTAL
+
+        #: Supported intensity measure types are spectral acceleration,
+        #: peak ground velocity and peak ground acceleration, see pag 1990.
+        DEFINED_FOR_INTENSITY_MEASURE_TYPES = set([
+               PGA,
+               PGV,
+               SA
+        ])
+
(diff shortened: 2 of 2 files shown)


micheles:
We need tests for the new GSIMs, similar to the ones you can find in oq-engine/openquake/hazardlib/tests/gsim for the GSIM classes.

g-weatherill:
Can you assign the review to me? This GMPE is on my to-do list so I can check it and add the tables if necessary.

openquake/hazardlib/gsim/yenier_atkinson_2015.py
```diff
@@ -0,0 +1,559 @@
...
+                else:
+                    return C['e0'] + \
+                        C['e1'] * (rup.mag - C['Mh'])
+
+        def _compute_stress_adjustment(self, rup, C_SA):
+                """
+                Compute stress adjustment term, equation (4), pag 1991.
+                """
+                # delta is the reference stress in bar
+                if rup.stress_drop <= 100.:
```
g-weatherill: The stress drop property needs to be added to the REQUIRES_RUPTURE_PARAMETERS set. This opens up the need to add stress drop as a source attribute though. Given that is most regions this is likely unknown the attribute needs to be optional, but is there an appropriate default value?

openquake/hazardlib/gsim/yenier_atkinson_2015.py
```diff
@@ -0,0 +1,559 @@
...
+                4.000	6.85	3.44E-1	    1.93E+0	-7.47E-2	1.51E+0	-5.51E-1	7.43E-2 -9.69E-01	844.48	760	    0	0.1	    -3.21E-03	-1.52E-03	0.583 0.385 0.698
+                5.000	6.85	-7.92E-2	1.98E+0	-6.21E-2	1.59E+0	-5.80E-1	7.90E-2 -9.20E-01	793.13	760	    0	0.1	    -2.55E-04	-1.44E-03	0.601 0.437 0.744
+                6.500	7.15	-6.67E-3	1.97E+0	-5.45E-2	1.63E+0	-5.96E-1	8.12E-2 -8.34E-01	775.6	760	    0	0.1	    -3.64E-05	-1.37E-03  	0.0	  0.0   0.0
+                8.000	7.50	2.56E-1 	1.94E+0	-5.23E-2	1.59E+0	-6.09E-1	8.30E-2 -7.50E-01	760.81	760	    0	0.1	    1.00E-04	-1.37E-03	0.0	  0.0   0.0
+                10.000	7.45	-2.76E-1	1.97E+0	-4.63E-2	1.72E+0	-6.20E-1	8.42E-2 -6.56E-01	775	    760 	0	0.1	    0.00E+00	-1.36E-03	0.645 0.477 0.801
+                PGA		5.85	2.22E+0	    6.86E-1	-1.39E-1	7.66E-1	-6.19E-1	6.03E-2 -6.00E-01	1500	760	    0	0.1	    -1.50E-01	-7.01E-03	0.502 0.260 0.564
+                PGV		5.90	5.96E+0	    1.03E+0	-1.65E-1	1.08E+0	-5.79E-1	5.74E-2 -8.40E-01	1300	760	    0	0.1	    -1.00E-01	-8.44E-03   0.500 0.256 0.560
+        """)
+
```
g-weatherill: I'm seeing some over-indenting and alignment issues. Could be due to tabs in the .py files. If you can, please check this and replace with spaces.

openquake/hazardlib/gsim/gsim_table.py
```diff
@@ -525,7 +525,7 @@ def _return_tables(self, mag, imt, val_type):
             low_period = round(periods[0], 7)
             high_period = round(periods[-1], 7)
 
-            if imt.period < low_period or imt.period > high_period:
+            if round(imt.period, 7) < low_period or round(imt.period, 7) > high_period:
```
g-weatherill: Probably this is not the only point at which precision issues may cause problems. The fix looks good to me, and I guess we will find the others in due course.

openquake/hazardlib/gsim/yenier_atkinson_2015.py
```diff
@@ -0,0 +1,559 @@
...
+                """
+                # Computing Fm from equation (3), used in Fe. 
+                if rup.mag <= C['Mh']:
+                    return C['e0'] + \
+                        C['e1'] * (rup.mag - C['Mh']) + \
+                        C['e2'] * (rup.mag - C['Mh']) ** 2
+                else:
+                    return C['e0'] + \
+                        C['e1'] * (rup.mag - C['Mh'])
+
```
g-weatherill: I think this should be `C['e0'] + C['e3'] * (rup.mag - C['Mh'])` according to Eqn 3 of Yenier & Atkinson (2015) - this may be affecting the numbers you are getting

openquake/hazardlib/gsim/yenier_atkinson_2015.py
```diff
@@ -0,0 +1,559 @@
...
+                        cvc[i] = C['Vc']
+                
+                return_array = np.ones_like(vs30)
+
+                for i in range(len(vs30)):
+                        if vs30[i] <= cvc[i]:
+                                return_array[i] = C['c'] * np.log(vs30[i] / C['vref'])
+                        else:
+                                return_array[i] = C['c'] * np.log(cvc[i] / C['vref'])
+                return return_array
```
g-weatherill: As `C["Vc"]` is scalar this could be revised avoiding loops:

```python
    def _get_site_amplification_linear(self, vs30, C):
        """
        Compute site amplification linear term,
        equation (11), pag 1992.
        """
        # Inititalise with the Vs30 > Vc case as the default
        return_array = (C["c"] * np.log(C["Vc"] / C["vref"])) *\
            np.ones_like(vs30)
        # If any sites have Vs30 <= Vc then apply only to that subset
        idx = vs30 < C["Vc"]
        if np.any(idx):
            return_array[idx] = C["c"] * np.log(vs30[idx] / C["vref"])

        return return_array
```

g-weatherill:
I think this PR can be closed for the time being. The precision issue was fixed in #2967 whilst the complete NGA East suite is part of #3130 (noting that the author did not intend to push this particular GMPE here!)
The parametric form of Yenier & Atkinson (2015) should be added with tests in future, but it would make sense to move that to a new PR.

Issue #2919, opened 2017-06-22
Produce a `damages-stats` output for the `classical_damage` calculator


raoanirudh:
The `classical_damage` calculator, if used with a non-trivial logic-tree currently, produces outputs for each of the individual branches realized in the calculation. These are saved in the datastore with the name `damages-rlzs`. Unlike all of the other hazard and risk calculators however, mean and quantile results are currently not provided.

The mean calculated across all realized branches should be provided for all classical damage calculations, and stored with the name `damages-stats`, similar to the other risk calculators. For consistency with the other calculators, the calculation of the mean results can be skipped by specifying `mean_damage_maps = false` in the job configuration file.

Users of the classical damage calculator should also be able to request quantiles calculated across all realized branches, by using the parameter `quantile_damage_maps` in the job configuration file to specify a set of quantiles (eg. `quantile_damage_maps = 0.15, 0.50, 0.85`).

micheles:
Again, I would reuse the parameters quantile_loss_curves and mean_loss_curves to describe the risk stats, as we did in the classical_bcr case.

PR #2921, opened 2017-06-22
Implemented stats for classical_damage

micheles:
Closes https://github.com/gem/oq-engine/issues/2919.

### DIFF
--- a/openquake/commonlib/writers.py
+++ b/openquake/commonlib/writers.py
@@ -121,6 +121,7 @@ htranslator = HeaderTranslator(
     '(slight):float32',
     '(moderate):float32',
     '(extensive):float32',
+    '(extreme):float32',
     '(complete):float32',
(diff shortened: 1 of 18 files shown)


openquake/commonlib/writers.py
```diff
@@ -121,6 +121,7 @@ def write(self, descrs):
     '(slight):float32',
     '(moderate):float32',
     '(extensive):float32',
+    '(extreme):float32',
```
raoanirudh: The limit states defined in the fragility model need not necessarily always be ∈ {slight, moderate, extensive, extreme, complete}. They are user-defined strings.

micheles: I know. This is just to fix the headers in the .csv files in our tests and demos. Otherwise the headers will be littered with `:float32` postfixes. Solving the issue in general is outside the scope of this PR.

PR #2925, opened 2017-06-26
Changed the way the GMFs are stored

micheles:
They are now stored in a single dataset `gmf_data/data` instead of one dataset per group. Moreover, there is a now dataset `gmf_data/indices` storing start-stop indices, similarly to what done for ` all_loss_ratios`. The indices will be used in the `gmf_ebrisk` calculator. There is also some refactoring.


openquake/risklib/riskinput.py
```diff
@@ -659,56 +652,29 @@ class GmfDataGetter(GmfGetter):
...
+        Returns gmv records from the datastore, if any
         """
-        key = 'grp-%02d' % self.grp_id
-        try:
-            dset = self.gmf_data[key]
-        except KeyError:
-            return
-        for rec in dset[self.start:self.stop]:
-            yield rec
+        return self.gmf_data['data'][self.start:self.stop]
```
ptormene: Since this is not a generator anymore, you might rename it into `get_gmv`

micheles: The name `gen_gmv` has to stay, since it is the name of the API (all getters have a `gen_gmv` method). Sometimes it is a generator, sometimes it is not.

Issue #2734, opened 2017-04-22
Add a Web API to export the hazard curves of a specific site


micheles:
Similarly to what has been proposed for the loss curves (https://github.com/gem/oq-engine/issues/2710).
The API could be the following:

```
- hazard_curves/rlzs/sid-42  # export all realizations of site ID 42
- hazard_curves/stats/sid-42  # export all statistics of site ID 42
- hazard_curves/rlz-001/sid-42  # export the realization 1 of site ID 42
- hazard_curves/quantile-0.1/sid-42  # export the quantile-0.1 curves of site ID 42
```

micheles:
I am closing this. At the moment the Colombia project works by storing the hazard curves statically, without any call to the WebUI and there is no need to do things differently for the foreseeable future.

Issue #2710, opened 2017-04-10
Add a web API to export the loss_curves


micheles:
Here is the API:

```
- loss_curves/sid-42/rlzs   # export loss curves of site #42 for all realizations
- loss_curves/sid-42/rlz-003   # export all loss curves of site #42, realization 3
- loss_curves/sid-42/stats   # export statistical loss curves of site #42
- loss_curves/sid-42/mean   # export mean loss curves of site #42
- loss_curves/sid-42/quantile-0.1   # export quantile loss curves of site #42

- loss_curves/ref-a1/rlzs   # export loss curves of asset a1 for all realizations
- loss_curves/ref-a1/rlz-003   # export loss curves of asset a1, realization 3
- loss_curves/ref-a1/stats     # export statistical loss curves of asset a1
- loss_curves/ref-a1/mean     # export mean loss curves of asset a1
- loss_curves/ref-a1/quantile-0.1    # export quantile loss curves of asset a1
```

micheles:
The exported data could be in JSON format, something like this:

```python
{"rlz-000": [{"ref": "a1", "structural": {"losses": [...], "poes": [...]}}, ....]}
```

micheles:
The risk scientists tells me that this not so important. I am closing it until it will return relevant.

PR #2929, opened 2017-06-27
Added a generic extraction API to the WebUI

micheles:
This will be used by QGIS and other clients. This is preliminary to https://github.com/gem/oq-engine/issues/2734 and https://github.com/gem/oq-engine/issues/2710.
Clients should make calls like this one:

```python
sitecol = pickle.loads(requests.get('http://localhost:8000/v1/calc/999/extract/sitecol').content)
```


doc/web-api.md
```diff
@@ -113,6 +113,19 @@ A list of error lines extracted from the log. If the calculation was
 successfull, the list is empty.
 
 
+#### GET /v1/calc/:calc_id/export/:out_key/:out_kind
+
+Get the exported file for the given output key and output kind.
+
+Parameters:
+
+    * export_type: the desired format for the file (`npz`, `csv`, etc.)
```
ptormene: Please update the description of parameters

openquake/calculators/extract.py
```diff
@@ -0,0 +1,137 @@
...
+    def __iter__(self):
+        return iter(self.array)
+
+    def __len__(self):
+        return len(self.array)
+
+    def __getitem__(self, idx):
+        return self.array[idx]
+
+    @property
```
ptormene: The only difference between `DatasetWrapper` and `DatagroupWrapper` is this `dtype` property. Please see if we can unify those objects

micheles: We can. Done.

PR #2931, opened 2017-06-28
Split the event_based_risk demos in hazard and risk

micheles:
Closes https://github.com/gem/oq-platform-ipt/issues/38. @ptormene , you will need to change the number of expected demos on the QGIS side.


CatalinaYepes:
Also the `oq-risk-tests` need to be updated

PR #2932, opened 2017-06-28
Upload artifacts only if a [dump] tag is provided or branch is master

daniviga:
To avoid uploading artifacts after every build, for every branch (~30MB per run) we can make travis upload `oqdata.zip` only if a `[dump]` tag is provided at the end of the commit message, or if `$branch` is master.

The same approach may be used in the future to turn on/off specific steps in Travis at commit time (i.e. instead of `[skip ci]` just `[skip demos]`)


micheles:
LGTM

PR #2936, opened 2017-06-29
Increased the `sz` field in the ruptures from 1 to 2 bytes

micheles:
Marco says that there are legitimate use cases where the surface discretization can produce more than 256 values.


mmpagani:
LGTM!

PR #2939, opened 2017-07-01
Changed the ordering of the fields in loss_curve_dt and loss_maps_dt

micheles:
For consistency with the ordering in the `loss_dt` dtype. The change only affects computations with insured losses and more than one loss type. For instance

`structural structural_ins contents contents_ins ->  structural contents structural_ins contents_ins`

PS: I am also adding a check on the loss types: the loss types in the loss_ratios (if any) must be the same as the loss types in the risk functions. If not, before an error was raised during the calculation, now it is raised upfront.


openquake/commonlib/oqvalidation.py
```diff
@@ -564,6 +564,15 @@ def is_valid_complex_fault_mesh_spacing(self):
...
 
+    def is_valid_loss_ratios(self):
+        """
+        The loss types in the loss_ratios dictionary {loss_ratios} are not
+        the ones for which there are risk functions: {_risk_files}
+        """
+        ltypes = sorted(self.loss_ratios)
+        expected_ltypes = sorted(self.risk_files)
+        return not ltypes or ltypes == expected_ltypes
+
```
ptormene: Please add a test

PR #2941, opened 2017-07-03
Smart resolution of the heisenbug in the calculation of the loss maps

micheles:
This was hard. There was a change of philosophy from "it is better a working slow computation than a failing fast calculation" to "it is better to fail early than to be too slow". We cannot wait for 256 hours (~11 days) to complete a single core calculation, it is better to fail early and to have to user repeat the calculation with the --hc option, so that the parallelism fully works and the calculation can be done in 1 hour (numbers are made up but meaningful).

In order to enable (partial) parallelism even without the --hc option, it was necessary to change the LossRatiosGetter so that the loss ratios are read upfront in the controller node. In this way the reading part is still sequential but at least the computation part is parallel. Also, if the reading part runs out of memory, or celery dies because of the excessive data transfer, there is an early error which is good.

Moreover, there were strange errors in the cluster when computing the loss maps: this kind of bugs happens with large computations with structured arrays in the workers and never happens in the tests and demos. I solved the damn bug by replacing the loss maps structured arrays with plain vanilla arrays. This implied changing the way the loss_maps are stored and therefore changing the exporters.
Both the .csv and .npz outputs are unchanged.

While working at this PR, I also implemented the following improvements:

1. fixed a bug in `split_in_block` when the number of blocks is zero;
2. improved the error message for non-pickleable objects
3. allowed the existence of calculators not defining `.rlzs_assoc`
4. renamed `scientific.build_loss_dtypes` to `scientific.build_loss_curve_dt`
   and simplified it to return only the `loss_curve_dt` and not `loss_maps_dt`
5. renamed `fromworker` to `lazy`
6. we are passing to the workers just the realization weights and not the
   realization objects, thus reducing the data transfer
7. we are passing to the workers just the asset values and not the asset
   objects, thus reducing the data transfer
8. we are passing to `build_all_poes` just the number of realizations and
   not the realizations
9. the string representation of the DataStore now also contains the status
   of the datastore (`open` or `close`)
10. changed `oq db find` to returns the calculation ordered by runtime

Computing the loss maps with the --hc option is still faster. Here are some numbers for Chile on our cluster:
```
   with --hc
   ======================= ======== ========= ======
   operation               time_sec memory_mb counts
   ======================= ======== ========= ======
   total build_loss_maps   55,310   36        1,236 
   getting loss ratios     37,766   43        1,236 
   ======================= ======== ========= ======
   without --hc
   ============================== ========= ========= =======
   operation                      time_sec  memory_mb counts
   ============================== ========= ========= =======
   total build_loss_maps          16,770    43        1,236
   building lrgetters             841       18,115    1
   ============================== ========= ========= =======

37766s divided 256 cores means 148s, a lot better than the 841s spent sequentially.
```

The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2546.


openquake/risklib/riskinput.py
```diff
@@ -99,17 +99,19 @@ def assets_by_site(self):
             assets_by_site[ass['site_id']].append(self[i])
         return numpy.array(assets_by_site)
 
-    def values(self):
+    def values(self, aids=None):
         """
+        :param: asset indices where to compute the values (default all)
```
ptormene: Please add the name of the parameter. We could also change the description of the default value.

PR #2946, opened 2017-07-04
Fixed bug when exporting hazard outputs by realization

micheles:
As reported by [Elisabeth  on the mailing list](https://groups.google.com/d/msg/openquake-users/F5FVBVzqcgs/A41UC1vJAwAJ), the command `oq export hazard_output/rlz-XXX` does not work as intended. Here I am fixing it.


openquake/qa_tests_data/classical/case_18/expected/uhs-rlz-1.csv
```diff
@@ -0,0 +1,5 @@
+# source_model_tree_path=('b1',),gsim_tree_path=('b12',),investigation_time=1.0
```
ptormene: It looks like we have an inconsistency in the format of the comment line that contains parameters. Here parameters are separated by `","` but somewhere else we have parameters separated by `", "`
Please check about this.

micheles:
They are green except for the heisenbug, so I am merging and fix the heisenbug immediately after.

PR #2948, opened 2017-07-04
Made the demo LogicTreeCase2ClassicalPSHA faster

micheles:
By increasing the mesh spacing and the area discretization it is now 4 times faster. This was the slowest demo we had before. @mmpagani please confirm that the parameters are still reasonable.


mmpagani:
I think that for a demo the discretisation spacing selected is fully acceptable so LGTM

PR #2953, opened 2017-07-12
Extends declustering algorithms to work in 3D

g-weatherill:
Previously the declustering algorithms in hmtk were working only in 2D, this extends them to work in 3D


micheles:
LGTM

mmpagani:
I think it would be advisable to give the user the possibility of deliberately choosing an option. In this sense, I suggest creating a `_3d.py` version of the two algorithms; hopefully this will not imply too much code duplication. The alternative for using the 2D option would be to merge the current PR and create catalogues with all depths == 0. However, I find this option inefficient and a bit dangerous since it would require changes to all the code is already using these functions in order to maintain the current behaviour. In the afternoon, if I'll find a better connection I'll also check the tests (sorry for this).

g-weatherill:
I can break the distance check into a separate method and create a new subclass for the 3D case, so you can have GardnerKnopoff3D. Alternatively, the choice between the 2D and 3D options could be made in the configuration file (defaulting to 2D if not specified). I can think of cases where one might wish to use either option.

mmpagani:
Graeme, I'm happy with the strategy of having separate subclasses for the 3D case.

g-weatherill:
Sorry, after a long delay I have made the changes and added the 3D options as subclasses. Tests are added

openquake/hmtk/tests/seismicity/declusterer/dec_gardner_knopoff_test.py
```diff
@@ -86,3 +86,21 @@ def test_dec_gardner_knopoff(self):
...
+        """
+        config = {'time_distance_window': GardnerKnopoffWindow(),
+                  'fs_time_prop': 1.0}
+        # Instantiate the declusterer and process the sample catalogue
+        dec = GardnerKnopoffType1_3D()
+        vcl, flagvector = dec.decluster(self.cat, config)
+        print('vcl:', vcl)
+        print('flagvector:', flagvector, self.cat.data['flag'])
+        self.assertTrue(np.allclose(flagvector, self.cat.data['flag']))
+
```
micheles: You want `numpy.testing.assert_allclose` here.

openquake/hmtk/seismicity/declusterer/dec_afteran.py
```diff
@@ -159,6 +159,14 @@ def decluster(self, catalogue, config):
 
         return vcl, flagvector
 
+    @staticmethod
```
mmpagani: If this method is used only for the 2D declustering, you can remove the depths since they are not used, or (perhaps even better) you can directly use the haversine function.

openquake/hmtk/seismicity/declusterer/dec_afteran.py
```diff
@@ -235,3 +243,20 @@ def _find_foreshocks(self, vsel, year_dec, time_window, imarker, neq):
...
+@DECLUSTERER_METHODS.add(
+    "decluster",
+    time_distance_window=TIME_DISTANCE_WINDOW_FUNCTIONS,
+    time_window=np.float)
+class Afteran3D(Afteran):
+    """
+    Adaptation of the Afteran method with hypocentral distance
+    """
+    @staticmethod
+    def _get_distance(longitude, latitude, lon_i, lat_i, depth, depth_i):
```
mmpagani: maybe you can call this function '_get_hypo_distance'

openquake/hmtk/seismicity/declusterer/dec_gardner_knopoff.py
```diff
@@ -155,3 +158,30 @@ def decluster(self, catalogue, config):
...
+    "decluster",
+    time_distance_window=TIME_DISTANCE_WINDOW_FUNCTIONS,
+    fs_time_prop=np.float)
+class GardnerKnopoffType1_3D(GardnerKnopoffType1):
+    """
+    Adaptation of the Gardner & Knopoff Type 1 algorithim taking into account
+    depth
+    """
+    @staticmethod
+    def get_distance(longitude, latitude, lon_i, lat_i, depth=None,
```
mmpagani: This function is repeated. Wouldn't it be more efficient to put it in a utils.py file?

g-weatherill:
Further changes are underway to optimise the declustering using the numpy datetime tools. Suggest closing the PR and will create new one with further revisions in future.

PR #2955, opened 2017-07-17
Added experimental support for SCOOP

micheles:
We tried it on our cluster and it works. Here is an example

```
$ OQ_DISTRIBUTE=scoop /opt/python35/bin/python -m scoop --hostfile hosts /usr/local/openquake/oq-engine/bin/oq engine --run /home/michele/Tanzania/job.ini
```

where the file `hosts` has the following content:
```
wilson 1
mercury 64
marley 64
dylan 64
cobain 64
```


micheles:
After some testing we discovered that:

1. apparently scoop does not shutdown cleanly
2. if the computation runs out of memory it hangs
3. the project looks abandoned (see https://github.com/soravux/scoop/issues/60)

Despite all that, it works for Tanzania when celery hangs. I will reopen this PR if the project will be revived.

PR #2961, opened 2017-07-26
Refactor the reset command and make it work for multi_user

daniviga:
With this PR `oq reset` now deletes the DB and the WebUI datastore even if `multi_user = true`. To be able to use it in `multi_user` the command must be executed by the `openquake` user. If not, only `oqdata` for the user running the command will be deleted:

```bash
[daniele@centos7-oq ~]$ sudo -u openquake oq reset
Do you really want to destroy all your data? (y/n) y
Removed /var/lib/openquake/db.sqlite3. You must re-apply any existing migration.
Removed /var/lib/openquake/oqdata
```

```bash
[daniele@centos7-oq ~]$ oq reset
Do you really want to destroy all your data? (y/n) y
Removed /home/daniele/oqdata
```

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,6 @@
+  [Daniele Viganò]
+  * Extended the 'oq reset' command to work on multi user installations
+
   [Michele Simionato]
   * Fixed a bug: if there are multiple realizations and no hazard stats,
     it is an error to set hazard_maps=true or uniform_hazard_spectra=true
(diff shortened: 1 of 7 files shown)


micheles:
I would remove the sentence `You must re-apply any existing migration.` If I remember correctly migrations are automatically run by system at database creation, there is no need to confuse the user.

daniviga:
@micheles migrations are referred to Django migrations: if you have the authentication enabled you must run `oq webui migrate` before you can use the WebUI again. This is not done automatically (and it happened this morning)

micheles:
Ah, thanks, I was not thinking of the authentication. LGTM, but please add a line to the changelog.

Issue #2954, opened 2017-07-13
Proposed Uncertainty Model for NGA East GMPEs


g-weatherill:
One forthcoming task is the implementation of the GMPEs for Central and Eastern North America (CENA) published as part of the NGA East Project (http://peer.berkeley.edu/publications/peer_reports/reports_2015/webPEER-2015-04_NGA-East_8.26.16.pdf). One GMPE is already in the library (http://docs.openquake.org/oq-engine/master/openquake.hazardlib.gsim.html#module-openquake.hazardlib.gsim.shahjouei_pezeshk_2016) and another is waiting inside a PR (https://github.com/gem/oq-engine/pull/2913). There are more to come.

In contrast to the majority of existing GMPEs in the OpenQuake library the NGA East models have a separate, and common, model for the standard deviations, as described by Al Atik (2015) (http://peer.berkeley.edu/publications/peer_reports/reports_2015/webPEER-2015-07-Alatik.pdf). This contains coefficients for inter-event, intra-event (single station and ergodic) and total standard deviations (presented in Chapter 5 of Al Atik 2015).

Here is the problem. The full standard deviation model takes into account epistemic uncertainties, and sets of coefficients are presented for the following cases:
1. Central, Low and High branches
2. Global, Constant (for CENA) and Magnitude-Dependent (for CENA)
3. Ergodic and Single Station

Considering all of these possibilities would result in 18 subclasses for each NGA East model (of which there are about 20). Implementing all of this would produce around 360 new GMPE classes. This doesn't seem like a good approach!

Instead I propose to implement the single common standard deviation model, with the corresponding options passed to the GMPE at instantiation (and set by the user in the nrml). For this I would create a new subclass of the openquake.hazardlib.gsim.base.GMPE class that accepts three optional keyword arguments at instantiation:

`model` : The choice of model from `global` (default), `cena_constant` (the magnitude independent model for CENA) and `cena` (the magnitude-dependent model)
`branch`: Choosing from `central` (default), `low`, `high`
`ergodic`: The choice of ergodic (True, default) or single station (False) models

The user could set their choice in the nrml via something like:
```xml
<logicTreeBranch branchID="b13">
    <uncertaintyModel model="global" branch="central" ergodic="True">DarraghEtAl20151CCSP</uncertaintyModel>
     <uncertaintyWeight>1.0</uncertaintyWeight>
</logicTreeBranch>
```

[...]

PR #2963, opened 2017-07-27
Modified the Coefficients Table to be instantiated with dictionaries as well as strings

g-weatherill:
It is increasingly common to modify coefficients of published GMPEs depending on the application. This adapts the Coefficients Table object to support input of the coefficients in terms of a tuple of dictionaries (one for SA coefficients the other for Non-SA coefficients), in addition to the current means of definition using a string. This can allow for the coefficients table to be modified dynamically without the need for multiple tables to be hard coded. This is needed for https://github.com/gem/oq-engine/issues/2954

### DIFF
--- a/debian/changelog_hazardlib
+++ b/debian/changelog_hazardlib
@@ -1,6 +1,3 @@
-  [Graeme Weatherill]
-  * Adapts CoeffsTable to be instantiated with dictionarys as well as strings
-  
   [Michele Simionato, Daniele Viganò]
   * Merged the `oq-hazardlib` repository into `oq-engine`.
     The `python-oq-hazardlib` package is now provided by `python-oq-engine`
--- a/debian/changelog_hazardlib
+++ b/debian/changelog_hazardlib
@@ -1,3 +1,6 @@
+  [Graeme Weatherill]
+  * Adapts CoeffsTable to be instantiated with dictionarys as well as strings
+  
   [Michele Simionato, Daniele Viganò]
   * Merged the `oq-hazardlib` repository into `oq-engine`.
     The `python-oq-hazardlib` package is now provided by `python-oq-engine`
--- a/openquake/hazardlib/gsim/base.py
+++ b/openquake/hazardlib/gsim/base.py
@@ -940,8 +940,8 @@ class CoeffsTable(object):
                 else:
                     self.non_sa_coeffs[key] = table[key]
         else:
-            raise ValueError('CoeffsTable cannot be constructed with inputs '
-                             'of the form %s' % type(table))
+            raise TypeError("CoeffsTable cannot be constructed with inputs "
+                            "of the form '%s'" % table.__class__.__name__)
 
     def _setup_table_from_str(self, table, sa_damping):
         """
--- a/openquake/hazardlib/gsim/base.py
+++ b/openquake/hazardlib/gsim/base.py
@@ -916,15 +916,11 @@ class CoeffsTable(object):
     It is also possible to instantiate a table from a tuple of dictionaries,
     corresponding to the SA coefficients and non-SA coefficients:
 
-    >>> sa_coeffs = {SA(0.1): {"a": 1.0, "b": 2.0},
-    ...              SA(1.0): {"a": 3.0, "b": 4.0}}
-    >>> non_sa_coeffs = {PGA(): {"a": 0.1, "b": 1.0},
-    ...                  PGV(): {"a": 0.5, "b": 10.0}}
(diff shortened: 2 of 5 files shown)


micheles:
Why force the user to distinguish SA coefficients and non-SA coefficients? The user could pass a single dictionary and inside the code could distinguish if the IMT is a SA one or not.

openquake/hazardlib/tests/gsim/base_test.py
```diff
@@ -634,3 +635,49 @@ def test_ordering_and_equality(self):
...
+            table1.sa_coeffs,
+            {SA(period=0.1, damping=5): {"a": 1.0, "b": 2.0},
+             SA(period=1.0, damping=5): {"a": 5.0, "b": 10.0},
+             SA(period=10.0, damping=5): {"a": 10.0, "b": 20.0}}
+            )
+
+    def test_table_dict_instantiation(self):
+        # Check that the table instantiates with pre-defined dictionaries
+        table1 = CoeffsTable(sa_damping=5, table=self.coefficient_string)
+        # Create target dictionatu
```
mmpagani: dictionary

openquake/hazardlib/gsim/base.py
```diff
@@ -912,20 +912,46 @@ class CoeffsTable(object):
...
             raise TypeError('CoeffsTable requires "table" kwarg')
-        table = kwargs.pop('table').strip().splitlines()
+        table = kwargs.pop('table')
+        self.sa_coeffs = {}
+        self.non_sa_coeffs = {}
         sa_damping = kwargs.pop('sa_damping', None)
         if kwargs:
             raise TypeError('CoeffsTable got unexpected kwargs: %r' % kwargs)
+        if isinstance(table, str):
+            self._setup_table_from_str(table, sa_damping)
```
micheles: Did you run the tests both with Python 2 and Python 3? I wonder if the str/bytes change may cause trouble here.

daniviga:
```python
test_table_bad_instantiation (openquake.hazardlib.tests.gsim.base_test.CoeffsTableTestCase) ... FAIL

======================================================================
FAIL: test_table_bad_instantiation (openquake.hazardlib.tests.gsim.base_test.CoeffsTableTestCase)
----------------------------------------------------------------------
  File "/home/dvigano/openquake/oq-engine/openquake/hazardlib/tests/gsim/base_test.py", line 682, in test_table_bad_instantiation
    "CoeffsTable cannot be constructed with "
AssertionError: "CoeffsTable cannot be constructed with inputs of the form <class 'int'>" != "CoeffsTable cannot be constructed with inputs of the form <type 'int'>"
- CoeffsTable cannot be constructed with inputs of the form <class 'int'>
?                                                            ^^^^^
+ CoeffsTable cannot be constructed with inputs of the form <type 'int'>
?                                                            ^^^^


----------------------------------------------------------------------
Ran 528 tests in 78.733s
```

daniviga:
Hazardlib tests on python3 are broken:

```bash
test_table_bad_instantiation (openquake.hazardlib.tests.gsim.base_test.CoeffsTableTestCase) ... FAIL

======================================================================
FAIL: test_table_bad_instantiation (openquake.hazardlib.tests.gsim.base_test.CoeffsTableTestCase)
----------------------------------------------------------------------
  File "/home/dvigano/openquake/oq-engine/openquake/hazardlib/tests/gsim/base_test.py", line 682, in test_table_bad_instantiation
    "CoeffsTable cannot be constructed with "
AssertionError: "CoeffsTable cannot be constructed with inputs of the form <class 'int'>" != "CoeffsTable cannot be constructed with inputs of the form <type 'int'>"
- CoeffsTable cannot be constructed with inputs of the form <class 'int'>
?                                                            ^^^^^
+ CoeffsTable cannot be constructed with inputs of the form <type 'int'>
?                                                            ^^^^
```

debian/changelog_hazardlib
```diff
@@ -1,3 +1,6 @@
+  [Graeme Weatherill]
+  * Adapts CoeffsTable to be instantiated with dictionarys as well as strings
+
```
mmpagani: Move this to the 'changelog' file. See https://github.com/gem/oq-engine/blob/master/debian/changelog

openquake/hazardlib/gsim/base.py
g-weatherill: What is the python 2 and 3 compatible way to do this?

Also, the error Daniele notes is due to the use of `type(foo)` is giving a different string between Python 2 and 3. Is `x.__class__.__name__` safer/cleaner to use?

micheles:
LGTM

PR #2965, opened 2017-08-02
Fix some typos and dev installation now uses bin wheels as default

daniviga:
As suggested by @ptormene


daniviga:
Backported to `engine-2.5`

PR #2967, opened 2017-08-02
Minor fixes for floating precision in gsim tables and tests

g-weatherill:
Some minor errors were turning up when running the OQ-engine test suite on my machine, which seem to relate to the erroneous use of floating point values when specifying numpy array dimensions (I'm using numpy 1.13.1). This PR fixes these errors.

Also, resolves the gsim_table period precision issue contained within https://github.com/gem/oq-engine/pull/2913 (recommend to close that PR as the other issue may be resolved in due course)

### DIFF
--- a/openquake/hazardlib/gsim/gsim_table.py
+++ b/openquake/hazardlib/gsim/gsim_table.py
@@ -524,8 +524,8 @@ class GMPETable(GMPE):
 
             low_period = round(periods[0], 7)
             high_period = round(periods[-1], 7)
-
-            if imt.period < low_period or imt.period > high_period:
+            if (round(imt.period, 7) < low_period) or\
+                (round(imt.period, 7) > high_period):
                 raise ValueError("Spectral period %.3f outside of valid range "
                                  "(%.3f to %.3f)" % (imt.period, periods[0],
                                                      periods[-1]))
--- a/openquake/hazardlib/tests/gsim/chiou_youngs_2014_test.py
+++ b/openquake/hazardlib/tests/gsim/chiou_youngs_2014_test.py
@@ -163,7 +163,7 @@ class ChiouYoungs2014NearFaultDistanceTaperTestCase(BaseGSIMTestCase):
             imts=[PGV()],
             gsim=ChiouYoungs2014NearFaultEffect(),
             truncation_level=0,
-            realizations=1.
+            realizations=1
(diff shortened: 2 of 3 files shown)


mmpagani:
The changes are really minor so this PR LGTM. Nonetheless, I wonder if we should design a more specific test addressing the issue you're trying to resolve.

micheles:
LGTM

g-weatherill:
Not sure if there is a test that would catch this. I think it is a change between numpy 1.11 (packaged w/OQ wheels) and numpy 1.12 (I'm running 1.13) - https://docs.scipy.org/doc/numpy/release.html#id5 wherein this has been turned from a deprecation warning to an error. 
Any further action needed to close this PR?

daniviga:
It would be good to test OQ with the "officially supported" dependency versions and with the latest one, but this would require lot of computational resources and/or time.
Anyway `setup.py` is already forcing `numpy < 1.12` which means that OQ should refuse to install with newer `numpys` (it can do nothing if installation is forced or `PYTHONPATH` is used):

```python
install_requires = [
    'mock >=1.0, <1.4',
    'h5py >=2.2, <2.8',
    'nose >=1.3, <1.4',
    'numpy >=1.8, <1.12',
    'scipy >=0.13, <0.18',
    'psutil >=1.2, <4.5',
    'shapely >=1.3, <1.6',
    'docutils >=0.11, <0.14',
    'decorator >=3.4, <4.1',
    'django >=1.6, <1.12',
    'matplotlib >=1.5, <2.0',
    'requests >=2.2, <2.13',
    # pyshp is fragile, we want only versions we have tested
    'pyshp >=1.2.3, <1.2.11',
    'PyYAML',
]
```

g-weatherill:
Thanks Daniele. The small bug (if it can be called that) this fixes is something that is incorrectly implemented to begin with, it is simply that in numpy versions < 1.12 this has been tolerated but from 1.12 onwards numpy throws an error. With this fix I have the oq-hazardlib running on 1.13 (presumably it would work on 1.12 as well), so you could extend the range of supported numpy versions.

daniviga:
Thanks @g-weatherill! That's a real good news. I'll make some extended test (on Mac, PPC64...) and then I will increase the allowed version range for numpy in the `setup.py`

PR #2972, opened 2017-08-09
Add a deprecation warning to changelog_hazardlib and risklib

daniviga:
I'm using an invalid syntax to check that those files are *not* processed by dpkg/CI.

https://ci.openquake.org/job/zdevel_oq-engine/2566/


mmpagani:
LGTM

PR #2975, opened 2017-08-15
Changed the weighting algorithm of the sources

micheles:
Currently the source weight does not depend on the number of sites affected by the source. This is intuitively wrong: a source affecting 1000 sites must take more computational time than a source affecting 1 site. I changed the weighting algorithm and performed several experiments to show that the new weighting is better than the old one. I will list here the figures for the SHARE calculation:
```
master, total run time: 3h26m
master, average task time: 797±1,027 seconds with 2,309 tasks

this branch, total run time: 2h45m
this branch, average task time: 1,154±1,154 seconds with 1,588 tasks 
```
The task distribution has improved and there are less extra-small tasks than before. Information about the sources (number of sites, weight and task) are stored in `task_info/source_data`.

NB1: the numbers are slightly different because the change in the weights implies a change in the splitting and that has a numeric effect. This is normal.

NB2: there is now a maximum weight of 5,000,000 per task. The limit is there to reduce the issue of slow tasks. For the moment it is hard coded but it could become a configuration parameter in the future.

The demos are running here: https://ci.openquake.org/job/zdevel_oq-engine/2568/


micheles:
Task distribution in master:
old

Task distribution in this branch:
new

openquake/calculators/ucerf_classical.py
```diff
@@ -127,7 +127,8 @@ def ucerf_classical(
     pmap = poe_map(ucerf_source, s_sites, imtls, cmaker,
                    truncation_level, ctx_mon, pne_mons)
     nsites = len(s_sites)
-    pmap.calc_times = [(ucerf_source.source_id, nsites, time.time() - t0)]
+    pmap.calc_times = [
+        (ucerf_source.source_id, nruptures, nsites, time.time() - t0)]
```
ptormene: Please check if we need to use nruptures * nsites as weight

openquake/commonlib/source.py
```diff
@@ -654,7 +660,7 @@ def add_infos(self, sources):
         for src in sources:
             self.infos[src.src_group_id, src.source_id] = SourceInfo(src)
 
-    def split_sources(self, sources, src_filter, maxweight=MAXWEIGHT):
+    def split_sources(self, sources, src_filter, maxweight=MINWEIGHT):
```
ptormene: Please add a comment about maxweight=MINWEIGHT

openquake/hazardlib/source/base.py
```diff
@@ -52,7 +52,8 @@ def weight(self):
         """
         if not self.num_ruptures:
             self.num_ruptures = self.count_ruptures()
-        return self.num_ruptures * self.RUPTURE_WEIGHT
+        nsites = getattr(self, 'nsites', 1)
```
ptormene: Please add a comment about the fact that nsites could be missed only in tests

openquake/calculators/ucerf_classical.py
micheles: Done

openquake/commonlib/source.py
micheles: I turned `maxweight` into a mandatory argument

Issue #2974, opened 2017-08-10
Newer numpy is now supported in the HMTK


daniviga:
Following https://github.com/gem/oq-engine/pull/2967 `numpy` > 1.11 is now supported in the HMTK.

We need to update the changelog too (credits to @g-weatherill).

daniviga:
With `numpy==1.13` I still get some failing tests:

```======================================================================
ERROR: Basic test of the entire completeness analysis using a synthetic
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/openquake/oq-engine/openquake/hmtk/tests/seismicity/completeness/comp_stepp1971_test.py", line 301, in test_complete_stepp_analysis_basic
    self.process.completeness(self.catalogue, self.config))
  File "<decorator-gen-4>", line 2, in completeness
  File "/home/openquake/oq-engine/openquake/hmtk/registry.py", line 110, in caller
    return fn(obj, catalogue, config, *args, **kwargs)
  File "/home/openquake/oq-engine/openquake/hmtk/seismicity/completeness/comp_stepp_1971.py", line 183, in completeness
    n_times))
  File "/home/openquake/oq-engine/openquake/hmtk/seismicity/completeness/comp_stepp_1971.py", line 343, in get_completeness_points
    np.log10(sigma[id0, iloc]))
  File "/home/openquake/oq-engine/openquake/hmtk/seismicity/completeness/comp_stepp_1971.py", line 373, in _fit_bilinear_to_stepp
    x_0 = [-1.0, xdata[len(xdata) / 2], xdata[0]]
IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices

======================================================================
ERROR: Tests the function get_completeness_points using a synthetic
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/openquake/oq-engine/openquake/hmtk/tests/seismicity/completeness/comp_stepp1971_test.py", line 269, in test_get_completeness_points
    ntime)
  File "/home/openquake/oq-engine/openquake/hmtk/seismicity/completeness/comp_stepp_1971.py", line 343, in get_completeness_points
    np.log10(sigma[id0, iloc]))
  File "/home/openquake/oq-engine/openquake/hmtk/seismicity/completeness/comp_stepp_1971.py", line 373, in _fit_bilinear_to_stepp
    x_0 = [-1.0, xdata[len(xdata) / 2], xdata[0]]
[...]

PR #2976, opened 2017-08-16
Changed a test to make it compatible with numpy 1.13

micheles:
Part of https://github.com/gem/oq-engine/issues/2974.


micheles:
It was a bad idea. I am back to comparing the arrays (it requires to write more, since there there is a comparison for each IMT).

PR #2985, opened 2017-08-21
Refactored the RlzsAssoc object

micheles:
The final goal would be to simplify the RlzsAssoc object but there is a lot more to do. Some unused code has been removed at least.


openquake/commonlib/tests/source_test.py
```diff
@@ -810,15 +800,15 @@ def count_ruptures(src_group):
...
-0,SadighEtAl1997(): ['<0,b1_b3_b6~@_b3,w=0.04>']
-2,SadighEtAl1997(): ['<1,b1_b3_b7~@_b3,w=0.12>']
-4,SadighEtAl1997(): ['<2,b1_b3_b8~@_b3,w=0.04>']
-6,SadighEtAl1997(): ['<3,b1_b4_b6~@_b3,w=0.12>']
-8,SadighEtAl1997(): ['<4,b1_b4_b7~@_b3,w=0.36>']
-10,SadighEtAl1997(): ['<5,b1_b4_b8~@_b3,w=0.12>']
-12,SadighEtAl1997(): ['<6,b1_b5_b6~@_b3,w=0.04>']
-14,SadighEtAl1997(): ['<7,b1_b5_b7~@_b3,w=0.12>']
-16,SadighEtAl1997(): ['<8,b1_b5_b8~@_b3,w=0.04>']>"""
+0,SadighEtAl1997(): [0]
```
ptormene: This has to be approved by the scientific team

micheles: Why? This is the internal `__repr__` of the RlzsAssoc object which is never used by the scientists as far as I know. It is an implementation detail.

ptormene: Ok, I was only wondering if scientists might benefit from a more informative representation. If they don't use it at all, it's ok to make this change.

micheles: The information on the realizations is available in an official place which is the output realizations.csv. That one is unchanged by this PR.

PR #2984, opened 2017-08-21
Experimental support for zmq

micheles:
To be tested on the cluster.

PR #2990, opened 2017-09-01
Refactoring of parallel.py

micheles:
This is preliminary to the zmq move (contains the zmq-independent parts of https://github.com/gem/oq-engine/pull/2984).

### DIFF
--- a/openquake/baselib/parallel.py
+++ b/openquake/baselib/parallel.py
@@ -640,7 +640,7 @@ class Starmap(object):
 
     def add_task_no(self, iterargs, pickle=True):
         """
-        Add .task_no and .weight to the monitor and yields back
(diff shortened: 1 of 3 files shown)


openquake/baselib/parallel.py
```diff
@@ -608,55 +602,64 @@ def wait(self):
...
         """
         try:
-            nargs = len(self.task_args)
+            return len(self.task_args)
         except TypeError:  # generators have no len
-            nargs = ''
-        if nargs == 1:
-            [args] = self.task_args
-            add_task_no(args, 0)
+            return ''
```
ptormene: Please return -1 or something that makes explicit that you can't extract the len from a generator

micheles: Done

PR #2991, opened 2017-09-03
Using zmq in the dbserver

micheles:
Two reasons:

1) the DbServer is simplified
2) we can gain experience with zmq in a simple case.

I am removing some unused code in performance.py too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,6 @@ install:
   - pip -q install wheels/*
   - pip -q install -e .
 
-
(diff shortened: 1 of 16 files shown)


daniviga:
more tests:
- Windows: https://ci.openquake.org/job/windows/job/zdevel_oq-engine/14/console
- Mac: https://ci.openquake.org/job/macos/job/zdevel_mac_oq-engine/96/console

PR #2992, opened 2017-09-04
Add a note about encoding issue in macOS

daniviga:
A known issue (already addressed in the installer), which was missing in the doc.

Reported by Laurentiu.


micheles:
LGTM

PR #2993, opened 2017-09-05
Turned the DbServer into a multithreaded server added a `get_data` action

micheles:
The goal is to be able to extract numpy arrays from the datastore. Here is a simple example of usage:

```bash
$ oq db get_data 10175 performance_data
[(b'reading site collection', 0.0867161750793457, 0.66015625, 1)
 (b'reading composite source model', 0.09082245826721191, 0.16015625, 1)
 (b'prefiltering source model', 0.035729408264160156, 0.703125, 1)
 (b'managing sources', 98.83192563056946, 0.18359375, 1)]
```

I will need to add a test script extracting large arrays in parallel, to see how well the server works.


openquake/server/dbserver.py
```diff
@@ -44,25 +46,44 @@ class DbServer(object):
     """
     A server collecting the received commands into a queue
     """
-    def __init__(self, db, address, authkey):
+    def __init__(self, db, address, authkey,  num_workers=5):
```
daniviga: Maybe in the future we can put something between 2 and CPU/2

PR #2996, opened 2017-09-06
Commit upgrade_db in the dbserver

micheles:
The migrations were not committed when starting the DbServer, causing very surprising results.


micheles:
This does not seems to be enough, but it needed anyway so I am merging it.

PR #2997, opened 2017-09-06
Set relevant default to 1 instead of true

ptormene:
Otherwise, some calculations were not found by the filter in the web api, and the integration tests with the QGIS plugin were broken.


micheles:
LGTM

PR #3002, opened 2017-09-07
Trying to close and reopen conn after migration

daniviga:
It seems that the only way to see a change (ALTER TABLE) in a column default is to close and re-open the connection to the DB (commit isn't enough).

I'm also reverting the change in the original migration for consistency.

https://travis-ci.org/gem/oq-irmt-qgis/builds/272813819

The problem can be reproduced with
```
$ oq reset -y
$ oq dbserver start
$ oq engine --run job.ini
```
looking inside the job table in the db the field `relevant` must be `1` and not `true`, since the upgrade script 0001 must have changed the default from `true` to 1.

### DIFF
--- a/bin/check_demos
+++ b/bin/check_demos
@@ -18,7 +18,7 @@
 #  along with OpenQuake.  If not, see <http://www.gnu.o']rg/licenses/>.
 import os.path
 import getpass
-from openquake.server.manage import db
+from openquake.server.dbserver import db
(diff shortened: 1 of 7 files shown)


micheles:
I did some cleanup. The integration tests are running here: https://travis-ci.org/gem/oq-irmt-qgis/builds/272813819

PR #3004, opened 2017-09-08
Saved memory in event_based_risk

micheles:
Some simplification done 3 months ago when implementing the new gmf_ebrisk calculator caused the engine to use more memory than before. I was aware of the fact, but the impact was minor on the calculations I studied at the time, so I was not worried. However Mabe' calculation is particularly bad since the hazard part is extremely heavy (mostly due to a large maximum_distance and lots of sites), has a big logic tree and the memory effect is particularly severe. With the change implemented in this PR we are back computing the hazard for each GSIM and not for several GSIMs at once. In the case of Mabe' this reduces by 3 the amount of needed memory. Still the computation does not run on the cluster since it requires way more than 2 GB per core.
A serious change of strategy is needed to perform such calculations, but this fix is good enough for the moment. The release of engine 2.6 is not blocked anymore.

The demos with Python 2.7 are running here: https://ci.openquake.org/job/zdevel_oq-engine/2578

### DIFF
--- a/openquake/calculators/event_based.py
+++ b/openquake/calculators/event_based.py
@@ -334,7 +334,7 @@ def compute_gmfs_and_curves(getter, oq, monitor):
         duration = oq.investigation_time * oq.ses_per_logic_tree_path
         with monitor('building hazard', measuremem=True):
             gmfdata = numpy.fromiter(getter.gen_gmv(), getter.gmf_data_dt)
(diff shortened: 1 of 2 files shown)


micheles:
This a reduced version of Mabe' computation, using 45 GB on my workstation:

master

This is the same computation with the new branch; the maximum memory consumption is just
save_memory

PR #3005, opened 2017-09-08
Remove check_mem_usage from the safety_call cause is giving troubles

daniviga:
Warn logs are working with both futures and celery. Check is done in `submit` and `iter`.


micheles:
LGTM

PR #3006, opened 2017-09-09
Renamed the `agg_loss` outputs as `agg_curves`

micheles:
As discussed with Anirudh. I am also removing some obsolete code and doing a small refactoring.


micheles:
I will fix that in a different PR.

micheles:
It turns out the behaviour was correct. The number were like this:
```
array([  896.4173584 ,   938.59576416,   973.93951416,  1054.1640625 ,
                  nan,            nan,            nan,            nan], dtype=float32)
>>> numpy.interp(quantile, cum_weights, sorted_data)
938.5957574560808
```
which is correct for the quantile 0.15 on 8 realizations. For higher realizations one gets the NaNs.
Notice that numpy orders the NaNs on the right.

Issue #2926, opened 2017-06-26
Fix the loss curves exporter for statistics


micheles:
Right now we have

```
$ oq export loss_curves/mean
NotImplementedError
``

PR #3008, opened 2017-09-11
Implement asset loss curves in event based risk

micheles:
The numbers in the aggregate curves are different in some cases since I have fixed a bug with zero losses (they were ignored, now they are not). Also fixes https://github.com/gem/oq-engine/issues/2926.


openquake/risklib/scientific.py
```diff
@@ -1446,21 +1422,25 @@ class LossesByPeriodBuilder(object):
 
     :param insured_losses: insured losses flag from the job.ini
     """
-    def __init__(self, return_periods, loss_dt, num_rlzs, eff_time):
+    def __init__(self, return_periods, loss_dt, weights, num_events, eff_time):
```
ptormene: Please update the docstring (that even before these changes was not aligned with the signature)

micheles: Done

Issue #3009, opened 2017-09-11
Use the new loss curves algorithm to compute the loss maps


micheles:
The trick is to extract the PoEs from the annual frequency of exceedence,
```
poe = 1 - exp(-afe * t)
```

where `t` is the risk investigation time.

PR #3011, opened 2017-09-11
Generated the event based loss maps from the new loss curves

micheles:
Closes https://github.com/gem/oq-engine/issues/3009. Here is a run with Python 2.7: https://ci.openquake.org/job/zdevel_oq-engine/2581


openquake/risklib/scientific.py
```diff
@@ -1500,9 +1382,21 @@ def build(self, agg_loss_table, stats=()):
...
-        else:
-            array_stats = None
-        return array, array_stats
+        return self.pair(array, stats)
+
+    def build_maps(self, losses, clp, stats=()):
+        """
+        :param losses: an array of shape (A, R, P)
+        :param clp: a list of C conditional loss poes
+        :param stats: list of pairs [(statname, statfunc), ...]
```
ptormene: Please change `list` with `sequence`

PR #3012, opened 2017-09-11
Manage ctrl-c in a better way, killing children

daniviga:
Currently `CTRL-C` isn't working very well. It needs to be pressed several times and does not always kill the children (on Linux this is mitigated by `prctl`[1]) leaving orphans.

This simple change makes `CTRL-C` working at the first shot [2] and concurrent.futures child process will be brought down.

This doesn't change the behavior (revoke) when celery is used.

Currently it produces a stacktrace from every thread. ~process. I've been unable to change this behavior even registering the signal after the fork. This may be an issue on single nodes with lot of cores.~

[1] prctl has a different goal, which is to bring down children processes when master is killed by `SIGKILL`.
[2] more hits are still required when the dbserver is started by oq. I still need to manage this case.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,7 @@
+  [Daniele Viganò]
+  * Changed the 'CTRL-C' behaviour to make sure that all children
+    processes are killed when a calculation in interrupted
+
   [Michele Simionato]
   * Replaced the agg_curve outputs with losses by return period outputs
   * Turned the DbServer into a multi-threaded server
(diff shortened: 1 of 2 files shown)


micheles:
Can we remove the dependency on `prctl` now?

daniviga:
No, we don't: see [1] it's used for a different scope. It makes sure that not children are left if master is killed via `SIGKILL`. It can be still useful on clusters.

micheles:
But now we are intercepting SIGTERM and killing the workers, so it seems that we are doing the same thing on a single machine. On a cluster there is no forking on the master node so it is no needed anyway.

daniviga:
Yes, but this would work only with `SIGTERM`, not with `SIGKILL`.

micheles:
Just add a line in the changelog, LGTM

PR #3014, opened 2017-09-13
Fixed a bug in the calculation of distances

micheles:
This was discovered by Julio when calculating distances between a 1D mesh and a 2D mesh. `geodetic.pure_distances` was working only for 1D meshes. now it works also for multidimensional meshes.


julgp:
LGTM

klunk386:
LGTM

PR #3015, opened 2017-09-13
Make the sites.csv file mandatory in the calculator gmf_ebrisk

micheles:
As discussed with Anirudh. In this way the correspondence between lons, lats and site IDs is explicit. The file must have the format `site_id,lon,lat`.


micheles:
The fixes here are enough to manage scenario-like GMFs, when `number_of_ground_motion_fields` is fixed, but not enough to manage event_based-like GMFs. These will be considered in a future PR.

PR #3016, opened 2017-09-13
Added a check for the SourceGroup tag into nrmlSourceModelParser

klunk386:
The xml source parser of htmk works only with nrml4. With this modification the parser can also be used with nrml5 files, but a proper parser (including the SourceGroup information) should be better implemented for that.

We would suggest also to create a generic parser which is version independent, working also with old nrml3 files.


micheles:
We already have a version independent parser, the issue is that the HMTK is not using it (for historical reasons). It should be an easy fix.

micheles:
Superseded by https://github.com/gem/oq-engine/pull/3017

PR #3017, opened 2017-09-14
Removed the obsolete nrmlSourceModelParser

micheles:
The old parser was not able to parse the NRML 0.5 format.


klunk386:
We have to make sure that no production code (e.g. for old existing models) still exists that depends on this Parser. If so, it might be perhaps more useful to keep it, but modified as a wrapper of the new parser, to preserve its interface for back compatibility. We should get the opinion of @mmpagani first.
For the rest it LGTM

micheles:
Unfortunately it is impossible to write a fully working compatibility wrapper, since the new parser has a different API and returns a list of source groups instead of a list of sources. So anybody having notebooks using the old parsers in the HMTK has to change his code. The error messages gives clear instructions of how to do that. Otherwise we could write a compatibility wrapper working only for NRML 0.4. Not sure if it is worth the effort, probably @g-weatherill has an opinion.

g-weatherill:
No problem in principle with the parser getting changed. The point of having this parser separate, as opposed to just calling the oq-engine parser directly (which would have been a lot easier), is that for the hmtk there is a need to load in source model files that have incomplete information (e.g. missing MFD, nodal plane dist) except for the geometry. The point being that the hmtk will give you the values you need. As the oq-engine parser always did the validation this was not possible without writing a separate parser. If this workflow (i.e. the ability to read incomplete xmls and build source models with missing attributes) is still possible with the proposed change then it is OK with me. If you are removing that functionality you remove a key feature of the toolkit.

micheles:
I missed that point, @g-weatherill . `nrml.parse(fname, converter)` is doing all validations, so it is no good in that respect. You can use `nrml.read(fname)` instead, which is doing a lot less validations but returns `Node` objects, not hazardlib objects. If this not enough for you then this PR has to be closed and we will have to keep the HMTK parser. In that case it has to be updated for NRML 0.5 as Valerio wanted to do originally.

mmpagani:
I think we should simple remove and replace. There’s no reason to 
maintain this old code and its outdated dependencies.

g-weatherill:
Correct me if I am mistaken Michele, but the parser as it was already parsed the xml to return a set of node objects. Most of the work was translating the node objects into the mtk source classes. Does anything more need to be changed beside replacing the call here https://github.com/gem/oq-engine/pull/3017/files#diff-d5e6c4112632e2398eefdf0d2c6c7bb4L460 with whatever is now the current way to parse the xml to a node set that is compatible with nrml 0.5?

micheles:
The parser `nrmlSourceModelParser` has the unique feature of parsing invalid source models, so it cannot be removed for people using that feature. For other people I strongly suggest NOT to use it.
Use `openquake.hazardlib.nrml.parse` instead.

 I am opening another pull request upgrading `nrmlSourceModelParser` to support NRML 0.5. Still it is ugly to have another parser in our code base that will inevitably go out of sync with the changes in NRML, but let be it.

PR #3022, opened 2017-09-15
Upgraded the `nrmlSourceModelParser` to support NRML 0.5

micheles:
This will not work in all cases, i.e. the new MultiPointSources are not supported. It should be better than before, though.


klunk386:
Tested on a xml fault source file. It works without problem. LGTM

mmpagani:
I'm fine with this change. Nonetheless - unless there's something I'm missing - I don't see advantages in keeping code that's duplicating functionalities already available.

PR #3027, opened 2017-09-18
Fixed a bug with MultiMFD/arbitraryMFD

micheles:
In that case one gets a list of arrays that must be concatenated.


klunk386:
Works perfectly. LGTM

PR #3028, opened 2017-09-18
Fixing the length check for arbitrary mfd

klunk386:
Using the iterator of MultiMFD, the method mfd_class fails with arbitrary mfd type. This is due to the check_constraints method of the ArbitraryMFD class, not working when self.occurrence_rates is a list.

### DIFF
--- a/openquake/hazardlib/mfd/arbitrary_mfd.py
+++ b/openquake/hazardlib/mfd/arbitrary_mfd.py
@@ -52,7 +52,7 @@ class ArbitraryMFD(BaseMFD):
         * Each number in occurrence rates list is non-negative.
         * Minimum magnitude is positive.
         """
-        if not self.occurrence_rates:
+        if not len(self.occurrence_rates):
             raise ValueError('at least one bin must be specified')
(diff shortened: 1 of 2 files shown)


klunk386:
I have also modified the generator output of the method get_annual_occurrence_rates of the MultiMFD class, to provide a full MFD in each iteration

micheles:
The double `for loop` here was on purpose, if you remove it I expect the tests to break.

klunk386:
Indeed, you're right. However, what is the purpose to output each bin separately in the loop? It would make more sense to me to get "blocks" of occurrence rates in output, as for the call of get_annual_occurrence_rates on a single mfd instance.

micheles:
In this way a MultiMFD instance behaves like a regular MFD when calling `get_annual_occurrence_rates`, so it does not break other parts of the code. Otherwise you will have to put `ifs` in several places in client code, distinguishing MultiMFDs from regular MFDs, which is probably worse than the current situation.

micheles:
LGTM

PR #3034, opened 2017-09-21
What's new in engine 2.6

micheles:
See https://github.com/gem/oq-engine/blob/doc/doc/whats-new.md


ptormene:
You can refer to the changelog of the QGIS plugin (plugin versions 2.6.x are aligned with engine 2.6): https://plugins.qgis.org/plugins/svir/version/2.6.1/

doc/whats-new.md
```diff
@@ -1,144 +1,195 @@
...
+We removed an excessive check from the WebUI: now if an output exists,
+it can be downloaded even if the calculation was not successful.
+
+Now we show the user the error message in the case of a calculation
+that cannot be deleted from the WebUI.
+
+We changed the .npz exporters of the hazard outputs to make the import
+into a single QGIS layer easier.
+
+We have now in place an automatic testing mechanism making sure that
```
ptormene: I would change this into something like:
"We have now in place an automatic integration testing mechanism making sure that all the outputs of the demos that are loadable by the QGIS plugin are actually imported and visualized correctly."

doc/whats-new.md
```diff
@@ -1,144 +1,215 @@
...
+We changed the behaviour of 'CTRL-C' to make sure that all children
+processes are killed when a calculation is interrupted.
+
+Python 3.6 is not officially supported, however we have fixed the only test
+which was breaking and we know for sure that it works. We also fixed some
+tests breaking with numpy 1.13 which however is still not officially supported.
+
+We fixed some tests breaking on macOS; the numbers there are slightly
+different, since the scientific libraries are compiled differently.
+This platform should be trusted less than Linux and Windows.
```
daniviga: I would skip this statement. Or at least just say 'This platform should be trusted less than Linux, which remains the reference'.

PR #3038, opened 2017-09-21
Fixed gmf_data export when the GMFs are read from a file

micheles:
Discovered by Catalina. This only affects the .npz export. We need to add a demo for this situation, otherwise it will be untested in the QGIS side.

### DIFF
--- a/openquake/calculators/base.py
+++ b/openquake/calculators/base.py
@@ -702,12 +702,12 @@ def get_gmv_data(sids, gmfs):
     return numpy.fromiter(it, gmv_data_dt)
 
 
-def get_gmfs(self):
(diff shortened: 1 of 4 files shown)


openquake/calculators/base.py
```diff
@@ -702,12 +702,12 @@ def get_gmv_data(sids, gmfs):
     return numpy.fromiter(it, gmv_data_dt)
 
 
-def get_gmfs(calc):
+def get_gmfs(self):
```
ptormene: I suggest to rename the parameter, otherwise it might be counter-intuitive

micheles: Done

Issue #3007, opened 2017-09-11
Calculation is not stopped when a process is killed/terminated 


daniviga:
When one of the subprocesses is killed (by the user, by the OOM killer) the computation isn't stopped and other processes continues to crunch data until the tasks have been completed.

This is quite dangerous because it leaves several processes alive using resources (or even idle but with memory allocated) and in the case of OOM killer intervention the situation get works because of the high memory pressure (i.e OOM kills one process base on scoring but this does not stop other processes to leak, depending on the memory usage rate, the OOM killer may be too slow).

One solution could be (on unix) using:
`os.killpg(os.getpgrp(), signal.SIGKILL)` in `raiseMasterKilled`

Please note the `SIGKILL`: because how `MultiProcess` works (no cancel, no revoke) a KILL is required to stop process without waiting for a task to end.

micheles:
Notice that the PIDs of all workers are known and stored in the list `openquake.baselib.parallel.executor.pids`, so one could loop on those and kill them.

daniviga:
It seams that the only clean solution would be catch the `BrokenProcessPool` exception when submitting new tasks or when iterating on the results. If the exception is raised we can kill all the processes and die:

```python
    try:
        r = f.result()
    except futures.process.BrokenProcessPool as e:
        for pid in parallel.executor.pids:                                         
            try:                                                                   
                os.kill(pid, signal.SIGKILL)                                       
            except:                                                                
                pass   
        raise futureKilled
```

micheles:
I suggest to change the base except to an `except OSError:  # pid not found`

PR #3052, opened 2017-09-26
Used `setproctitle` in the multiprocessing workers

micheles:
In this way one can use `killall oq-worker` to kill all workers at once. It is up to Daniele to decide if we want to include setproctitle in our distribution. It is an optional dependency. It is installed already in our cluster ~~via apt-get~~. It is a workaround for https://github.com/gem/oq-engine/issues/3007.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -8,13 +8,13 @@ Homepage: http://github.com/gem/oq-engine
 
 Package: python-oq-engine
 Architecture: all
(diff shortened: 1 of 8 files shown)


daniviga:
@micheles if you want to test please remember that our cluster does not use apt's Python distribution and pip3 must be used instead

micheles:
Now that you mention it, I probably installed it on the cluster via pip3. But it was before the summer vacations, I forgot. Anyway, it works on the cluster.

daniviga:
Nice feature. I would be in favor of having this enabled by default in our installers.

https://github.com/gem/oq-builders/pull/74/
https://github.com/gem/oq-libs/pull/24
https://github.com/gem/oq-engine/pull/3057

Issue #3058, opened 2017-09-27
Change `extract` so that it can return a key-value generator


micheles:
This will save memory. For instance `oq extract hazard/rlzs` may extract one realization at the time and save it in the .hdf5 file. As it is now, all realizations must be kept in memory. It requires changes in `openquake.server.views.extract` and `openquake.commands.extract` too.

micheles:
After testing on the cluster, I checked we can extract the 3200 realizations of SHARE in 2 hours consuming only 10 GB. Still, implementing this feature is useful: with lazy extraction on can kill the operation, save a few realizations and see what it is inside the HDF5 file, while with eager extraction this is impossible: the file will be empty.

PR #3059, opened 2017-09-27
Lazy extraction

micheles:
Closes https://github.com/gem/oq-engine/issues/3058. Also, I am storing the numpy array in compressed format to save data transfer.


ptormene:
I've also checked that the QGIS plugin can keep reading the compressed data, without any change in the code.

Issue #3050, opened 2017-09-25
OQ_DISTRIBUTE=celery behavior


daniviga:
The behavior of `OQ_DISTRIBUTE=celery` is different from changing the `openquake.cfg`

in case of `OQ_DISTRIBUTE=celery` no
```
INFO:root:Using celery@localhost.localdomain, 16 cores
```
is reported and the availability of celery is evaluated only when tasks are submitted.

PR #3060, opened 2017-09-27
Fixed `openquake.engine.USE_CELERY` flag

micheles:
Closes https://github.com/gem/oq-engine/issues/3050. The flag was read from the configuration file and the environment variable OQ_DISTRIBUTE was ignored.

### DIFF
--- a/openquake/engine/engine.py
+++ b/openquake/engine/engine.py
@@ -35,7 +35,7 @@ from openquake.commonlib import logs
 TERMINATE = valid.boolean(
     config.get('distribution', 'terminate_workers_on_revoke') or 'false')
 
-USE_CELERY = config.get('distribution', 'oq_distribute') == 'celery'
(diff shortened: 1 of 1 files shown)


openquake/engine/engine.py
```diff
@@ -35,7 +35,7 @@
 TERMINATE = valid.boolean(
     config.get('distribution', 'terminate_workers_on_revoke') or 'false')
 
-USE_CELERY = config.get('distribution', 'oq_distribute') == 'celery'
+USE_CELERY = os.environ.get('OQ_DISTRIBUTE') == 'celery'
```
daniviga: Shouldn't be both? From env and from config?

micheles: No, look inside `openquake.commands.__main__`: the environment variable is set from the config file if the user does not specify it. It is perverse, but I could not find another way to transmit the information to baselib .parallel which has no access to the configuration file, which is in commonlib.

PR #3067, opened 2017-09-29
Added the ability to extract attributes from the datastore

micheles:
This is useful for Paolo. Now he can extracts the damage_states with `/extract/composite_risk_model.attrs` which returns the attributes of the composite_risk_model object as an .npz file, without building the full object. The dictionary contains the damage_states among other things.


micheles:
Test added @ptormene

PR #3071, opened 2017-09-29
Extract losses by asset

micheles:
With this the QGIS plugin will be able to make calls like `/extract/losses_by_asset/<rlz>/<loss_type>`
and get the losses, which will be aggregated by using the information in `/extract/assetcol/aids_by_tag`.

PS: probably a better idea is to perform the aggregation server side, see https://github.com/gem/oq-engine/issues/3080


openquake/calculators/extract.py
```diff
@@ -176,3 +177,22 @@ def extract_hazard(dstore, what):
...
+    "rlz-0/structural", "rlz-1/contents_ins", "mean/structural" etc
+    """
+    rlz_or_stat, lt = what.split('/')
+    r = int(rlz_or_stat[4:]) if rlz_or_stat.startswith('rlz-') else None
+    lti = dstore['oqparam'].lti
+    if 'losses_by_asset' in dstore:  # scenario_risk
+        return dstore['losses_by_asset'][:, r, lti[lt]]['mean']
+    elif 'avg_losses-rlzs' in dstore:  # event_based_risk
+        if r is None:
+            assert rlz_or_stat == 'mean', rlz_or_stat
```
ptormene: Please add a comment about the fact that only the mean can be extracted, and no other stat (e.g. quantiles)

Issue #3064, opened 2017-09-27
baselib.parallel depends from hazardlib


micheles:
If OQ_DISTRIBUTE=celery, parallel.py imports celeryconfig.py that imports commonlib.config that imports hazardlib. The solution is to move config inside baselib. While at it, I could also rewrite the configuration code, which is quite ugly.

PR #3073, opened 2017-09-29
Replaced the old configuration mechanism

micheles:
One of the last remnants of the past is now gone, and ~270 lines have been saved. As a bonus, closes https://github.com/gem/oq-engine/issues/3064. Jenkins is happy: https://ci.openquake.org/job/zdevel_oq-engine/2588/


openquake/baselib/__init__.py
```diff
@@ -47,8 +47,22 @@ def __getattr__(self, key):
...
 
 
-def _read(*paths, **validators):
-    # load the configuration file by looking at the given paths
+def read(*paths, **validators):
+    """
+    Load the configuration, make each section available in a separate dict.
+
+    The configuration location can specified via an environment variable:
+       - OQ_CONFIG_FILE
```
daniviga: `s/OQ_CONFIG_FILE/OQ_CONFIG_FILE_VAR/g`

or if you like more `OQ_CONFIG_FILE` (I do) we can change it in the code

micheles: Changed in the code to be consistent with the documented name.

Issue #3000, opened 2017-09-06
`oq run` is not logging correctly sometimes


micheles:
I discovered that
```bash
$oq run oq-engine/openquake/qa_tests_data/event_based_risk/case_4/job.ini   # if split_sources=true
```
logs fine with Python 3.5, but logs only the warnings if rtree is missing. The problem is probably that
the "missing rtree" log is called too soon. Notice that everything works with `oq engine`.

PR #3078, opened 2017-10-02
Fixed logging when missing rtree

micheles:
Closes https://github.com/gem/oq-engine/issues/3000.

The problem was rather subtle: since `openquake.hazardlib.calc.filters` has a toplevel definition
`source_site_noop_filter = SourceFilter(None, {})`, in absence of rtree the `logging.info` was triggered at import time, BEFORE calling `logging.basicConfig`. As a consequence, the logger was configured at the WARN level (the default) and messages at INFO level were not printed anymore for missing rtree.

It looks like a "feature" of the logging module, `logging.basicConfig` if called too late is not honored. Did I ever mention that I hate the logging module?

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,6 @@
   [Michele Simionato]
+  * Fixed a bug with logging configured a WARN level instead of INFO level
+    if rtree was missing (affecting only `oq run`)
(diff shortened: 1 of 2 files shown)


daniviga:
Yes, you did.

PR #3085, opened 2017-10-03
Added an API to aggregate the damages

micheles:
The API requires to specify loss type and realization and returns an array of losses per damage state.


ptormene:
If I make a call like this: http://localhost:8800/v1/calc/29/extract/aggdamages/contents?city=Benicia&tract=338302
I get the following error:
`IndexError at /v1/calc/29/extract/aggdamages/contents`
`arrays used as indices must be of integer (or boolean) type`

micheles:
Good catch Paolo, there was a bug in the case of an empty selection (i.e. there are no assets with city=Benicia and tract=339302).

openquake/calculators/extract.py
```diff
@@ -178,10 +178,30 @@ def extract_hazard(dstore, what):
...
+
 @extract.add('agglosses')
 def extract_agglosses(dstore, loss_type, *tags):
     """
-    Aggregate losses of the given loss type and tags
+    Aggregate losses of the given loss type and tags.
+
+    :returns:
+        array of shape (R,), being R the number of realizations or
+        array of length 0 if there are no data for the given tags
```
ptormene: Please change `are` with `is`

PR #3090, opened 2017-10-04
Added an experimental distribution mode of kind "zmq"

micheles:
This has the potential to replace celery in the far future. For the moment it is not enabled in openquake.cfg.


daniviga:
Please also fix the `openquake.cfg` patch in debian, or let me know if I have to do it.

See https://ci.openquake.org/job/zdevel_oq-engine/2603/

micheles:
Yes, please fix the debian patch. The circular import is fixed now.

PR #3092, opened 2017-10-04
Add an API to aggregate loss curves by tag for event based

micheles:
The URL to call is `/v1/calc/1234/extract/aggcurves?tagname1=tagvalue1...`


openquake/calculators/extract.py
```diff
@@ -230,3 +230,20 @@ def extract_aggdamages(dstore, loss_type, *tags):
...
+
+@extract.add('aggcurves')
+def extract_aggcurves(dstore, loss_type, *tags):
+    """
+    Aggregate loss curves of the given loss type and tags for
+    event based risk calculations.
+
+    :returns:
+        array of shape (S, P), being P the number of return periods
+        and S the number of statistics
```
ptormene: It would be useful to extract with the same call, as attributes, also the names of stats and the list of return periods.

Issue #3084, opened 2017-10-03
"New version released" warning


daniviga:
We should add a warning, in the command line and in the webui, that notifies users that a new OQ Engine release is available (for stable releases). This:
 - Can be disabled via confing (`openquake.cfg`)
 - Should silently fail if no connection is available
 - Should not create delays: a cache should be used (a field in the DB is enough...)

In the past this issue was less problematic since everyone was using Linux packages that get updated with `apt` or `yum`. Nowadays most people are using installers which need manual upgrade.

Releases can be retrieved from here https://developer.github.com/v3/repos/releases/

pslh:
I think this is a good idea, do we have a time estimate?

micheles:
This check takes at least 0.5 seconds, so I would not add an option to disable it or to cache it.

PR #3093, opened 2017-10-04
Added a check on the latest version

micheles:
Closes https://github.com/gem/oq-engine/issues/3084

### DIFF
--- a/openquake/engine/engine.py
+++ b/openquake/engine/engine.py
@@ -20,19 +20,21 @@
 calculations."""
 
 import os
+import re
 import sys
 import signal
 import traceback
+import requests
 
 from openquake.baselib.performance import Monitor
-from openquake.baselib import parallel, config
+from openquake.baselib import parallel, config, __version__
 from openquake.commonlib.oqvalidation import OqParam
 from openquake.commonlib import datastore, readinput
 from openquake.calculators import base, views, export
 from openquake.commonlib import logs
 
+GITHUB = 'https://api.github.com/repos/gem/oq-engine'
 TERMINATE = config.distribution.terminate_workers_on_revoke
-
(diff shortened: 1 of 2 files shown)


openquake/engine/engine.py
```diff
@@ -244,3 +247,27 @@ def _do_run_calc(calc, exports, hazard_calculation_id, **kw):
...
+    """
+    try:
+        json = requests.get(GITHUB + '/releases/latest', timeout=0.5).json()
+        tag_name = json['tag_name']
+        current = version_triple(__version__)
+        latest = version_triple(json['tag_name'])
+    except:  # page not available or wrong version tag
+        return
+    if current < latest:
+        logs.LOG.warn('Version %s of the engine is available, but you are '
```
daniviga: Would it be possible to just return a string in the function instead of printing `log.LOG.warn`? In this way the function may be reused in the future by an API endpoint (i.e. via QGIS).

micheles: Done

openquake/engine/engine.py
```diff
@@ -244,3 +249,31 @@ def _do_run_calc(calc, exports, hazard_calculation_id, **kw):
...
+    groups = re.match(r'v?(\d+)\.(\d+)\.(\d+)', tag).groups()
+    return tuple(int(n) for n in groups)
+
+
+def check_obsolete_version(oqparam):
+    """
+    Check if there is a newer version of the engine.
+
+    :returns: a message if the current version is obsolete or the empty string
+    """
```
daniviga: Please specify what is returned is case of errors (like GH api is unreachable)

micheles:
Update changelog and docstring

PR #3096, opened 2017-10-05
Add a warning box to WebUI when new release available

daniviga:
The box is shown only is the check is successfully and it's made asynchronously.

### DIFF
--- a/doc/web-api.md
+++ b/doc/web-api.md
@@ -234,9 +234,10 @@ Logout
 Return a string with the OpenQuake engine version
 
 
-#### GET /engine_new_version
+#### GET /engine_latest_version
 
-Return a string with if new versions have been released
+Return a string with if new versions have been released.
+Return 'None' if the version is not available
 
 
 #### GET /v1/available_gsims
--- a/doc/web-api.md
+++ b/doc/web-api.md
@@ -234,6 +234,11 @@ Logout
 Return a string with the OpenQuake engine version
 
 
+#### GET /engine_new_version
+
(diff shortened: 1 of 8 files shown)


openquake/server/views.py
```diff
@@ -198,6 +198,15 @@ def get_engine_version(request):
 
 @cross_domain_ajax
 @require_http_methods(['GET'])
+def get_engine_new_version(request):
+    """
+    Return a string with if new versions have been released
+    """
+    return HttpResponse(engine.check_obsolete_version(None))
+
```
micheles: Please change `check_obsolete_version` and remove the argument `oqparam` which is not used anyway

openquake/server/urls.py
```diff
@@ -25,6 +25,7 @@
 urlpatterns = [
     url(r'^$', RedirectView.as_view(url='/engine/', permanent=True)),
     url(r'^engine_version$', views.get_engine_version),
+    url(r'^engine_new_version$', views.get_engine_new_version),
     url(r'^v1/calc/', include('openquake.server.v1.calc_urls')),
```
micheles: I prefer the name `engine_latest_version`

doc/web-api.md
```diff
@@ -234,6 +234,11 @@ Logout
 Return a string with the OpenQuake engine version
 
 
+#### GET /engine_new_version
+
+Return a string with if new versions have been released
+
```
micheles: add "returns 'None' if the version is not available"

micheles:
LGTM

PR #3097, opened 2017-10-05
Experimental support for Python 3.6

daniviga:
Just to avoid getting the branch stale. This is experimental and tests are still on Python 3.5 for now. It contains, at least, everything that is needed to easily run tests or setup an installation on Linux.


micheles:
LGTM

PR #3100, opened 2017-10-05
Use api.openquake.org instead of GitHub

daniviga:
Use https://api.openquake.org and provide some extra info via the user agent.

Currently blocked server-side

### DIFF
--- a/openquake/engine/engine.py
+++ b/openquake/engine/engine.py
@@ -23,6 +23,7 @@ import os
 import re
 import sys
 import signal
+import logging
 import traceback
 import requests
 import platform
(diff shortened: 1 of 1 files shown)


daniviga:
`api.openquake.org` deployed

micheles:
I have added some information about the calculation_mode (server side we should discard the calls without calculation_mode: they happens at every web UI refresh and are not interesting; the calls with calculation_mode instead are interesting since they correspond to actual calculations).

TODO: should we avoid doing any call for calculations running on Travis/Jenkins? It should be possible to do so by looking at some environment variable.

daniviga:
I would use something as 'WebUI' instead of a calculation mode. In this way I can filter out the requests server-side with non modifications to django-requests.

For the second point it would be easy to create an utils function `is_ci` than evaluates
```python
def is_ci():
    return True if os.environ.get('JENKINS_URL') or os.environ.get('TRAVIS') else False
```

micheles:
Yes, but I want calculation_mode ;-) It is very interesting for us to know which kind of calculators are most used.

daniviga:
No, I mean: you leave the calculation mode, but when is empty just put WebUI. But just skip it for now, I should be able, using regexp, to skip the agent when it has `^OpenQuake.*;;.*` without touching the `django.request` code.

micheles:
LGTM on the part you did

daniviga:
LGTM too

PR #3105, opened 2017-10-07
Activate Tools automatically if openquakeplatform module is installed

daniviga:
Evaluate the `openquakeplatform` module to activate Tools instead of using a custom `local_settings.py` that needed manual activation.


openquake/server/settings.py
```diff
@@ -162,6 +166,27 @@
...
+
+    STANDALONE_APPS += (
+        'openquakeplatform_ipt',
+        'openquakeplatform_taxtweb',
+        'openquakeplatform_taxonomy',
+    )
+
+    INSTALLED_APPS += STANDALONE_APPS
+
+    FILE_PATH_FIELD_DIRECTORY = os.path.join(os.path.expanduser('~'), 'oqdata')
```
daniviga: This should be probably taken form the `openquake.cfg`

micheles: The right function to call is `openquake.baselib.datastore.get_datadir()`

PR #3110, opened 2017-10-10
Add HTMK to Travis

daniviga:
We forgot them...


daniviga:
See https://github.com/gem/oq-engine/pull/3109

PR #3111, opened 2017-10-10
Great unification of the event based and scenario GMFs

micheles:
It is not exaggerated to say that this pull request took 5 years in the making. After 5 years of nearly continuous work on the problem, after uncountable failures and small but significant progress, we are now in a state where:

1. the GMFs are stored in the *same way* both for the event based and the scenario calculators
2. all the calculators reading the GMFs from a file store the GMFs in the datastore first, so they are finally doing the same thing as the regular calculators
3. there is a common procedure to read the GMFs for all GMF-related calculators
4. all risk calculators except the event based one use now the same way of building the riskinputs
5.  the calculators reading the hazard curves from a file (classical_risk, classical_damage) store the poes in the datastore first, so that they are unified with the case of generating the hazard curves from the sources

A lot more needs to be done but this is a crucial step towards https://github.com/gem/oq-engine/issues/2017.

Jenkins is happy too: https://ci.openquake.org/job/zdevel_oq-engine/2598/


raoanirudh:
The benefits of having a clean slate: https://github.com/gem/oq-risk-tests/blob/master/src/calculators.jl#L161-L166

Issue #3113, opened 2017-10-10
FragilityFunctionDiscrete can not handle list properly


dynaryu:
Noticed that FragilityFunctionDiscrete can not handle list or array properly as below:
`a =  FragilityFunctionDiscrete(limit_state='slight', 
                                                     `imls` = [0.0, 0.2, 0.4, 0.6, 0.8, 1, 1.2, 1.4],
                                                      poes = [0.0, 0.771, 0.95, 0.989, 0.997, 0.999, 1., 1.],
                                                     no_damage_limit=0.05)
`
It works okay when it takes a single value, but it fails when it takes array and yields a wrong value if it takes a list:

`a(0.4)` 
`a([0.2, 0.4])`
`a(numpy.array([0.2, 0.4]))`

I changed the code as below to fix the error:
`class FragilityFunctionDiscrete(object):

    def __call__(self, iml):
        """
        Compute the Probability of Exceedance (PoE) for the given
        Intensity Measure Level (IML).
        """
        highest_iml = self.imls[-1]

        # when the intensity measure level is above
        # the range, we use the highest one
        try:
            iml = [highest_iml if x > highest_iml else x for x in iml]
        except TypeError:
            temp = self.interp(highest_iml if iml > highest_iml else iml)
        else:
            temp = self.interp(iml)

        if self.no_damage_limit:
             temp[iml < self.no_damage_limit] = 0.0
        return temp
`

PR #3114, opened 2017-10-11
Extended the fragility functions to accept arrays

micheles:
At user request, closes https://github.com/gem/oq-engine/issues/3113. Now the fragility functions are vectorized, and the same is true for `scientific.scenario_damage`.The speedup is *very* significant.
For instance for the ScenarioDamage demo the `computing risk` time goes down from 90 seconds to 5 seconds, with `number_of_ground_motion_fields=100`.


openquake/risklib/scientific.py
```diff
@@ -529,12 +530,19 @@ def __call__(self, iml):
...
-        # the range, we use the highest one
-        return self.interp(highest_iml if iml > highest_iml else iml)
+        try:
+            len(iml)
+        except TypeError:  # is a scalar
+            if self.no_damage_limit and iml < self.no_damage_limit:
+                return 0.
+            return self.interp(highest_iml if iml > highest_iml else iml)
+        else:  # is a sequence
+            iml = numpy.array(iml)
```
raoanirudh: Might be better to rename the variable to `imls` since now it can be a sequence of intensity levels.

Issue #3001, opened 2017-09-06
Scenario outputs


raoanirudh:
## Scenario output updates
The following changes are proposed for the scenario_damage and scenario_risk outputs:
- [ ] Format `realizations` output
- [ ] Hide `fullreport` output from webui
- [ ] Update `gmf_data`
- [ ] Deprecate `_by_tag` and `_total` outputs
- [ ] Update `_by_asset` output to include tags
- [ ] Add event-loss aggregation by tag to `losses_by_event` and deprecate the csv output
- [ ] Add `dmg_by_event` and `losses_by_event` outputs for scenario_damage

### Realizations
- [ ] Modify the column headers as follows:
uid→path
model→source_model

- [ ] Modify source model names as follows:
fake⟷scenario

### Fullreport
Hide the fullreport output from the webui

### GMFs
Use the unified gmf format for scenarios and event-based calculators, for both input and output file formats. Particularly, for the scenario gmf output, currently, a separate file is produced for each realization.

- [ ] The current output format for gmfs can be replaced by the newly proposed format which includes the gmfs for all realizations in a single file.
- [ ] A second file listing the site ids and corresponding lon, lat coordinates will also have to be produced and exported.
- [ ] Modify the column headers as follows:
rlzi→realization
sid→site_id
eid→event_id
gmv_IMT→IMT

### _by_tag and _total outputs
- [ ] The _total output can be trivially obtained from the _by_asset output. Thus, we can deprecate this output in future releases.
- [ ] The information in the current _by_tag output can be included in the _by_asset output by simply adding the tags corresponding to each asset. The aggregation by a particular tag can be handled in post-processing by the QGIS plugin. Thus, the _by_tag output can also be deprecated in future releases.

### _by_asset output
- [ ] To the current output, add the tags associated with each asset in additional columns.

### losses_by_event output
- [ ] This output should include, for each event, the event losses total and per tag
- [ ] This output should be provided as an npz or hdf5 file
- [ ] The csv output should be deprecated

### dmg_by_event and losses_by_event outputs for scenario_damage
- [ ] (Low priority) The `dmg_by_event` should be added as an output for the scenario_damage calculator. This output is analogous to the `losses_by_event` output for the scenario_risk calculator, but listing the expected number of buildings in each damage state, for each event in the scenario calculation.
[...]

PR #3123, opened 2017-10-11
Remove some scenario_damage outputs

micheles:
As requested in https://github.com/gem/oq-engine/issues/3001


micheles:
This is also blocked by changes in the manual

PR #3126, opened 2017-10-12
Add support for groups in the WebUI

daniviga:
Very simple group management for the WebUI. Users in the same group can see all the calculations, logs and outputs from all the members. Computations can be deleted only by the owner (the delete button returns an error, I'll see to hide totally the button  in the future).

If you like the idea I will update the documentation and the changelog (I don't want to spend time on it if this does not get merged).

I also added a link to the admin area (only when logged in as super).

### DIFF
--- a/openquake/server/db/actions.py
+++ b/openquake/server/db/actions.py
@@ -430,7 +430,7 @@ def calc_info(db, calc_id):
     return response_data
 
 
-def get_calcs(db, request_get_dict, user_name, user_acl_on=False, id=None):
+def get_calcs(db, request_get_dict, allowed_users, user_acl_on=False, id=None):
     """
     :param db:
         a :class:`openquake.server.dbapi.Db` instance
@@ -448,11 +448,6 @@ def get_calcs(db, request_get_dict, user_name, user_acl_on=False, id=None):
     """
     # helper to get job+calculation data from the oq-engine database
     filterdict = {}
-
-    # user_acl_on is true if settings.ACL_ON = True or when the user is a
-    # Django super user
-    if user_acl_on:
-        filterdict['user_name'] = user_name
 
     if id is not None:
         filterdict['id'] = id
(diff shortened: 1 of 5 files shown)


openquake/server/views.py
```diff
@@ -308,9 +309,9 @@ def calc(request, id=None):
     base_url = _get_base_url(request)
 
     user = utils.get_user_data(request)
-
+    allowed_users = user['group_memebers'] or [user['name']]
```
micheles: `group_memembers` does not look right, perhaps it is `group_members`?

openquake/server/views.py
```diff
@@ -470,7 +471,8 @@ def calc_results(request, calc_id):
     # throw back a 404.
     try:
         info = logs.dbcmd('calc_info', calc_id)
-        if user['acl_on'] and info['user_name'] != user['name']:
+        allowed_users = user['group_memebers'] or [user['name']]
+        if user['acl_on'] and info['user_name'] not in allowed_users:
```
micheles: Same `memembers` as before

openquake/server/db/actions.py
```diff
@@ -480,8 +475,17 @@ def get_calcs(db, request_get_dict, user_name, user_acl_on=False, id=None):
     else:
         time_filter = 1
 
-    jobs = db('SELECT * FROM job WHERE ?A AND %s ORDER BY id DESC LIMIT %d'
-              % (time_filter, limit), filterdict)
+    # user_acl_on is true if settings.ACL_ON = True or when the user is a
+    # Django super user
+    if user_acl_on:
+        users_filter = ("user_name IN ('%s')" %
+                        "', '".join([str(item) for item in allowed_users]))
```
micheles: Let's hope we do not have a user called Bobby table... https://imgs.xkcd.com/comics/exploits_of_a_mom.png

daniviga: This is something I don't like either... there's any facility to sanitize input?

micheles: Yes there is, the `?X` syntax, like `db('select * from job where user_name IN (?X)', ['michele', 'openquake'])`

daniviga:
Another way of making the query can be:
```patch
diff --git a/openquake/server/db/actions.py b/openquake/server/db/actions.py
index aa8efc3af..2c77d3e09 100644
--- a/openquake/server/db/actions.py
+++ b/openquake/server/db/actions.py
@@ -477,10 +477,9 @@ def get_calcs(db, request_get_dict, allowed_users, user_acl_on=False, id=None):
 
     # user_acl_on is true if settings.ACL_ON = True or when the user is a
     # Django super user
-    if user_acl_on:
-        users_filter = "user_name IN (?X)"
-    else:
-        users_filter = 1
+    users_filter = "user_name IN (?X)"
+    if not user_acl_on:
+        users_filter = "(1 OR %s)" % users_filter
 
     jobs = db('SELECT * FROM job WHERE ?A AND %s AND %s'
               ' ORDER BY id DESC LIMIT %d'
```

micheles:
Lets leave things as they are not since it works. FYI, the simplest way to see the query performed on the database is to write ` db('SELECT ...', ..., debug=True)` and then the query template and arguments will be printed on stdout.

daniviga:
Thanks @micheles

daniviga:
Documentation done via https://github.com/gem/oq-engine/commit/e621307f5c830883e886a0fd9f3083179eca8eb1

PR #3128, opened 2017-10-14
Stored the size of the GMFs generated while the computation is running

micheles:
To see the progress. This is useful in ultra-large calculations like the one by Mabe'.


micheles:
LGTM to myself since this is really trivial and I need it for Mabe's computation

PR #3130, opened 2017-10-14
Implements complete suite of NGA East GMPEs

g-weatherill:
Test are running here: https://ci.openquake.org/job/zdevel_oq-engine/...

This PR implements the complete NGA East suite of ground motion models for Central and Eastern North America (CENA), with their common aleatory and epistemic uncertainty configurations, as referenced here (https://github.com/gem/oq-engine/issues/2954).

The details of the model suite are found in three PEER reports, two describing the median models and their adjustments (http://peer.berkeley.edu/publications/peer_reports/reports_2015/webPEER-2015-04_NGA-East_8.26.16.pdf and http://peer.berkeley.edu/publications/peer_reports/reports_2015/webPEER-2015-08-NGA-East-Various826.pdf), and a third describing the aleatory uncertainty model that is common to all NGA East GMPEs (http://peer.berkeley.edu/publications/peer_reports/reports_2015/webPEER-2015-07-Alatik.pdf)

These are amongst the most complex models implemented, not in terms of functional form but in terms of the complex epistemic uncertainty structure. The core body of models for the mean consists of 21 GMPEs (including some variants from several authors). The uncertainty model, however, defines three models and three branches for the inter-event standard deviation, three models and three branches for the intra-event standard deviation, a non-ergodic form and an ergodic form with three branches of the site-to-site term. Enumerating all possible branches results in 10,206 GMPEs! Clearly it is impossible to implement these in the current form. Further complicating the issue is that some GMPEs do not have a parametric form and can only be implemented in the form of tables. So, what has been done here ...:

1. New base classes have been created that should be common to all NGA East GMPEs. These base classes are inherited from the `GMPETable` object for calculating ground motions from hdf5 tables. Whilst the authors of some GMPEs provided parametric forms, after applying the adjustments described in http://peer.berkeley.edu/publications/peer_reports/reports_2015/webPEER-2015-08-NGA-East-Various826.pdf, all the resulting mean models are non-parametric. The base class therefore will calculate the mean ground motion from hdf5 tables. It is highly likely future models (particularly for this region) will operate in a similar way, so the user may which to select one of the 21 core models, or run an "NGA East-style" GMPE with their own tables. Therefore for the core models the hdf5 tables are supplied directly, called in the conventional manner, with their path to the hdf5 files hard-coded e.g.
```xml
<logicTreeBranch branchID="b1">
    <uncertaintyModel>AlNomanCramer2015NGAEast</uncertaintyModel>
    <uncertaintyWeight>##</uncertaintyWeight>
</logicTreeBranch>
``` 
alternatively, if the user wishes to supply their own hdf5 tables this should be called as:
```xml
<logicTreeBranch branchID="b1">
    <uncertaintyModel>NGAEastBaseGMPE(gmpe_table="./path/to/nga_east_gmpe_table.hdf5")</uncertaintyModel>
    <uncertaintyWeight>##</uncertaintyWeight>
</logicTreeBranch>
```

2. The aleatory uncertainty model proposed by Al Atik (2015) describes for each tau, phi_ss and phi_s2s a mean and variance for the epistemic uncertainty. These are translated into branches by taking quantiles according to a chi^2 distribution (see pages 101 and 102 of Al Atik (2015) for explanation. Al Atik proposes a three branch model for low, center and high of the epistemic uncertainty for each parameter, with the corresponding quantiles of 0.05, 0.5 (not equal to mean!) and 0.95 respectively. Although Al Atik provides tables for the low, center and high values for each parameter the chi^2 distribution is implemented directly to allow the user to control the quantile for each parameter (though the tables from the paper itself are used for testing the implementation). However, Al Atik also defines a total standard deviation epistemic model propagating the mean and variance of the random effects components into a mean and variance for total sigma. These too can be used to define similar quantiles for total sigma. Unfortunately, using this approach when selecting a given quantile of total sigma, the inter-and intra-event standard deviations cannot be retrieved. Therefore the aleatory uncertainty model must be used in one of two different ways: i) via direct definition of the quantiles for the tau, phi_ss and phi_s2ss terms (automatically generating the total sigma), or ii) by direct definition of the quantile for total sigma. The former may give more flexibility to the user, but it is the latter that is more likely to be adopted for applications. As the two aleatory models are different, we provide two different base classes: `NGAEastBaseGMPE` and `NGAEastBaseGMPETotalSigma`. All 21 core models are duplicated, one with the random effects control the other with the total sigma.

3. As we cannot hard-code all of the options the control of the aleatory uncertainty model is passed to the user via optional key-words. These will be enumerated below. For the tau and phi_ss terms there are three different models proposed: 1. a "global" magnitude-dependent, 2. a constant (magnitude-independent) model for use in CENA, and 3. a magnitude-dependent model for use in CENA. Furthermore, the choice of ergodic or non-ergodic model can be made by choosing to adopt the CENA delta_s2s model (for the ergodic case) or omit the delta_s2s (for the non-ergodic case). 

For the `NGAEastBaseGMPE` class the following are the optional keywords controlling the aleatory uncertainty model (default marked in `{}`):

`tau_model` - choosing from `{global}` | `cena_constant` | `cena`
`phi_model` - choosing from `{global}` | `cena_constant` | `cena`
`phi_s2ss_model` - choosing `cena` or left blank (default)

The remaining keywords `tau_quantile`, `phi_ss_quantile` and `phi_s2ss_quantile` define the selected quantile of the epistemic distribution, which can take a floating value between 0 and 1 (inclusive).

For the `NGAEastBaseGMPETotalSigma` the choice of `tau_model`, `phi_model` and `phi_s2ss_model` are the same as for `NGAEastBaseGMPE`; however, the only quantile that can be controlled is the total sigma quantile (`sigma_quantile`), also a float between 0 and 1 (inclusive)

So ... some example usages:

1. Select the Darragh et al (2015) 1CCSP model, with global tau (mean), CENA phi_ss (0.95 quantile) and the ergodic option with phi_s2ss set to 0.05:
```xml
<logicTreeBranch branchID="b1">  
    <uncertaintyModel>DarraghEtAl2015NGAEast1CCSP(tau_model="global",phi_ss_model="cena",phi_s2ss_model="cena",tau_quantile=None,phi_ss_quantile=0.95,phi_s2ss_quantile=0.05)</uncertaintyModel>
    <uncertaintyWeight>##</uncertaintyWeight>
</logicTreeBranch>
```

2. Select the Darragh et al. (2015) 1CCSP model with the global tau model, the CENA phi_ss model, the CENA phi_s2ss model but for the 0.9 quantile for the total sigma:

```xml
<logicTreeBranch branchID="b1">   
    <uncertaintyModel>DarraghEtAl2015NGAEast1CCSPTotalSigma(tau_model="global",phi_ss_model="cena",phi_s2ss_model="cena",sigma_quantile=0.9)</uncertaintyModel>
    <uncertaintyWeight>##</uncertaintyWeight>
</logicTreeBranch>
```

3. Construct a "user-defined" NGA East GMPE with the mean described in an hdf5 table (located in `./path/to/my_ngaeast_table.hdf5`, choosing the median CENA tau model, the mean CENA-constant phi_ss model and the non-ergodic option:
```xml
<logicTreeBranch branchID="b1"> 
     <uncertaintyModel>BaseNGAEastModel(gmpe_table="./path/to/my_nga_east_table.hdf5",tau_model="cena",phi_ss_model="cena_constant",tau_quantile=0.5)</uncertaintyModel>
    <uncertaintyWeight>##</uncertaintyWeight>
</logicTreeBranch>
```

4. Using the same table as in example 3, choose the mean global tau model, the mean CENA phi model and the non-ergodic option, selecting the 0.05 quantile of the total standard deviation
 ```xml
<logicTreeBranch branchID="b1">   
    <uncertaintyModel>BaseNGAEastModelTotalSigma(gmpe_table="./path/to/my_nga_east_table.hdf5",tau_model="global",phi_ss_model="cena",sigma_quantile=0.05)</uncertaintyModel>
    <uncertaintyWeight>##</uncertaintyWeight>
</logicTreeBranch>
```

Further notes:

* For testing the mean models the original spreadsheets have been rendered into test tables.

* For testing the aleatory models the values used in the tables for low, central and high tau, phi_ss, phi_s2ss are used, whilst for the total sigma case the tables for low, central and high single-station and ergodic sigmas are used

* The tests for the complete (random effects) aleatory uncertainty model enumerate every single combination of output. As a result they take about 10 - 12 minutes to run, and have been assigned as a `slow` test. The tests for the median and total sigma models run in only a few seconds.

* An adaptation to the `check_gsim` function for the GMPE tests was made to allow it to be passed instantiated GMPE classes (though uninstantiated remains the default behaviour)

* A common site amplification model for application with these GMPEs has been recently published (http://peer.berkeley.edu/publications/peer_reports/reports_2017/2017_05_final_hashash_4.2.17.pdf). This is currently being reviewed and may be added to the model in a future pull request.

* Scripts for generating the hdf5 tables (and test tables) from the Excel files supplied in the electronic supplements to the PEER reports will be uploaded onto my own github repository in due course (unless objections arise)

### CODE REVIEW

File: openquake/hazardlib/tests/gsim/check_gsim.py [hunk 1]
```diff
@@ -54,12 +55,18 @@ def check_gsim(gsim_cls, datafile, max_discrep_percentage, debug=False):
         and a message pointing to a line with a test that failed will show up.
         If ``False`` the GSIM is executed in a vectorized way (if possible)
         and all the tests are executed even if there are errors.
+    :param instantiated:
+        If ``True`` the gsim_cls object is an instantiated GSIM class,
+        otherwise ``False``
```
micheles: This is really ugly. Can we just pass a gsim instance, always? I would change `gsim_by_import_path` 
`return gsim_class` -> `return gsim_class()` and this would work for GSIMs without parameters. If you want this to work also with GSIMs with parameters you can try `return object.__new__(gsim_class)`.

g-weatherill: Sorry, I need more information here. How do I do this so as to avoid having to change every call to a GMPE in all of the other GMPE tests? If I do `return object.__new__(gsim_class)` then it passes instantiated objects (i.e. with the parameters) but then the uninstantiated objects (which is all the rest of them) don't work. Do I need to go and change them all to `NameEtAlYYYY()`? If not, and I have to place an if statement here (i.e. return `gsim_class()` if not with parameters, and `object.__new__(gsim_class)` with them) then I don't understand the advantages of that over the way this has been implemented in the current PR? Can you advise what I need to change and where?

micheles: Let's do the following, which is a very minor change. Remove the argument `instantiated` from the `check_gsim` function and then replace the check 

```
￼     if instantiated:
￼         gsim = copy.deepcopy(gsim_cls)
￼     else:
￼         gsim = gsim_cls()
```
with
```
￼     if isinstance(gsim_cls, GroundShakingIntensityModel):
￼         gsim = copy.deepcopy(gsim_cls)
￼     else:
￼         gsim = gsim_cls()
```

in this way the user does not need to pass an argument than Python can figure out. I still do not like much having an argument `gsim_cls` which sometimes is a class and sometimes is an instance, but this can be fixed later on in another PR.

For the rest, LGTM. Please fix the conflict in the changelog and we will merge it.

PR #3132, opened 2017-10-15
Make server functional tests more robust

daniviga:
This PR should mitigate two issues:

1. Warmup of the Django server can take some more time on macOS and (even more) on Windows compared to Linux. We have already an utils function (used by `oq webui start`) that waits, with a max timeout, for the webserver to be ready. This makes the test actually faster on linux ( < 5 secs) and more robust on other OSes and slower servers

2. Remove (or at least mitigate) issues when more tests are run at the same time on OSes with no process/network isolation (again macOS and Windows). Currently all the runs are using the same port for the functional tests, causing one job to run tests across the wrong webserver with unexpected behaviors.

Should fix https://github.com/gem/oq-engine/pull/3131
https://ci.openquake.org/job/macos/job/zdevel_macos_engine/3/ [Build Status](https://ci.openquake.org/job/macos/job/zdevel_macos_engine/3/)

### DIFF
--- a/openquake/server/tests/functional_test.py
+++ b/openquake/server/tests/functional_test.py
@@ -32,7 +32,6 @@ import subprocess
 import tempfile
 import requests
 import numpy
-import socket
 from openquake.baselib.general import writetmp
 from openquake.engine.export import core
 from openquake.server.db import actions
@@ -45,18 +44,10 @@ if requests.__version__ < '1.0.0':
 
 
 class EngineServerTestCase(unittest.TestCase):
-    hostport = ''
+    hostport = 'localhost:8761'
     datadir = os.path.join(os.path.dirname(__file__), 'data')
 
     # general utilities
-
-    @classmethod
-    def getsocket(cls):
-        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-        s.bind(('', 0))
-        cls.hostport = 'localhost:%s' % s.getsockname()[1]
-        s.close()
-        return cls.hostport
 
     @classmethod
     def post(cls, path, data=None, **params):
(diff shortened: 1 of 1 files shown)


micheles:
The problem affects also the tests in https://github.com/gem/oq-engine/blob/master/openquake/baselib/tests/workerpool_test.py#L34
Moreover, the function returning the free port is subject to race conditions. Until we find a really reliable solution my suggestion is to run on macOS only the tests for Python 2.7 and comment out the ones for Python 3.5. Then there is no more port contention. When we will switch to Python 3.5 we will run only the ones for Python 3.5, no port contention either.

daniviga:
> The problem affects also the tests in https://github.com/gem/oq-engine/blob/master/openquake/baselib/tests/workerpool_test.py#L34

I would be in favor of skipping this test on foreign OSes (macOS, Windows) since we are not using it and will not use it soon, at least there. Instead the API server test is critical since it's widely used on macOS and Windows too.

> Moreover, the function returning the free port is subject to race conditions

True, but I would expect this to be infrequent since port is allocated during the startup of the server, and not just saved in a var and used later on (when the port may be taken) as I do already in Jenkins for the DbServer. We can add a new flag to `oq webui start` to generate a random port, but I would not change the code for this. 

> Until we find a really reliable solution my suggestion is to run on macOS only the tests for Python 2.7 and comment out the ones for Python 3.5. Then there is no more port contention. When we will switch to Python 3.5 we will run only the ones for Python 3.5, no port contention either.

Unfortunately this is not affecting just Python 3.5 vs Python 2.7 but is also affecting master vs zdevel. Is currently dangerous to run a zdevel job: if master is triggered just after, master is very likely to fails). We can anyway reduce the # of executors to one loosing some flexibility.

daniviga:
Just skip the randomization of WebUI port for now, and keep the server check.

PR #3141, opened 2017-10-16
Fixed oq plot_sites in absence of rtree

micheles:
Fixes the error in https://ci.openquake.org/job/macos/job/master_macos_engine/label=macos,python=python3.5/46/console

### DIFF
--- a/openquake/hazardlib/calc/filters.py
+++ b/openquake/hazardlib/calc/filters.py
@@ -276,8 +276,9 @@ class SourceFilter(object):
         self.use_rtree = use_rtree and rtree and (
             integration_distance and sitecol is not None and
             sitecol.at_sea_level())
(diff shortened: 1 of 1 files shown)


openquake/hazardlib/calc/filters.py
```diff
@@ -276,8 +276,9 @@ def __init__(self, sitecol, integration_distance, use_rtree=True):
         self.use_rtree = use_rtree and rtree and (
             integration_distance and sitecol is not None and
             sitecol.at_sea_level())
-        if self.use_rtree:
+        if sitecol is not None:
```
ptormene: As it's written here, it is not so intuitive. The fact that `if self.use_rtree` implies that also sitecol is not None does not look obvious at a first sight.

Issue #3091, opened 2017-10-04
Maximum Distance Filter for GMPE calculation


klunk386:
I have tested the new filter for a couple of magnitude values (maximum_distance = [(8, 200), (7, 100)]) and obtained non realistic results. This is likely due to linear extrapolation beyond bounding values, which filters in this case the intermediate magnitudes (<6) at very short distances. Although such behaviour can be avoided using a better binning within the maximum_distance command, I would suggest to introduce instead a step-wise filtering. In that way, the above example would be interpreted as:
`(do not calculate at distances above 100km for all magnitudes below 7) & (do not calculate  at distances above 200km for all magnitudes below 8) & (no distance filter above magnitude 8)`

klunk386:
This is a sketch of what I would compute using the following settings:
`maximum_distance = [(8, 200), (7, 150), (6, 100), (5, 50)]`

![distancescaling](https://user-images.githubusercontent.com/22977002/31169670-af61d9d4-a8f9-11e7-873a-7b62763b2555.png)

mmpagani:
I'm unsure on the type of feedback you expect from me on this issue. The first test you did seems perfectly reasonable to me since you are excluding magnitudes below 7. On the linear scaling I already told Michele that in my opinion it should be removed since it does not allow the user to properly control the behavior of the filter.

klunk386:
In the first test, magnitudes below 7 are not excluded, but used only within a very narrow distance range, because of the extrapolation.

mmpagani:
The interpolation should be removed. @micheles how complex it is to complete this stuff?

micheles:
Not so complex and it is already market for engine-2.7, so it will fixed soon.

PR #3145, opened 2017-10-17
Implemented piecewise magnitude-distance filtering

micheles:
Here is an example of how it works:

```python
In [3]: oq.maximum_distance  # 'default' means there is the same filter for all TRTs
Out[3]: {'default': [(6, 100), (7, 150), (8, 200)]}
In [4]: oq.maximum_distance('default', mag=5.)  # magnitude < minimum
Out[4]: 100
In [5]: oq.maximum_distance('default', mag=6.)  # magnitude 6, filter above 100 km
Out[5]: 100
In [6]: oq.maximum_distance('default', mag=6.5)  # magnitude 6.5, filter above 150 km
Out[6]: 150
In [7]: oq.maximum_distance('default', mag=7)  # magnitude 7 filter above 150 km
Out[7]: 150
In [8]: oq.maximum_distance('default', mag=7.5)  # magnitude 7.5, filter above 200 km
Out[8]: 200
In [9]: oq.maximum_distance('default', mag=8)  # magnitude 8, filter above 200 km
Out[9]: 200
In [10]: oq.maximum_distance('default', mag=9)  # magnitude above the maximum, filter above 500 km
Out[10]: 2000
```
~~As a bonus, now a configuration like `maximum_distance = [(4, 200)]` automatically discards all ruptures below magnitude 4, while for the ones above magnitude 4 only the ones more distant than 200 km are discarded.~~

Closes https://github.com/gem/oq-engine/issues/3091.


klunk386:
I don't think the implementation is correct in this way.
1) In the example, magnitudes above the maximum (>8) are filtered above 200km. This is not how it should be, as these magnitudes should not be filtered at all. The filter should be applied only below the prescribed magnitude value (in this case only below 8)
2) The bonus is also not correct in my opinion. Discarding all magnitudes below a threshold is not necessary, as this is controlled by the minimum magnitude parameters of the source. Moreover, as in the previous case, the 200km filter should be applied below 4, and not above, as indicated.
I suspect you're filtering in the wrong direction. To summarise: magnitudes BELOW a give threshold value cannot generate sensible ground motion ABOVE a certain distance, and should be filtered out, not vice-versa.

mmpagani:
Hi Michele, as indicated by Valerio the filter should work the other way round. i.e. following your example: take everything below magnitude 6 within 100 km, take everything with magnitude lower than 7 within 150 km and so on. Apologises, I realise I might have caused some confusion in one of my previous comments.

micheles:
I see, so below the minimum magnitude we keep filtering. My question is above the maximum magnitude: it looks wrong to me remove any filtering about magnitude 8, otherwise we will have to consider effects coming from 10,000 km away.

klunk386:
Above the M max you can use a realistic (arbitrary large) default value, such as 500km. Ground motion is definitely irrelevant (for our purposes) above this distance.

klunk386:
Once the tests are green, LGTM

micheles:
Notice that this is blocked until the manual is not updated. It will go in release 2.8.

micheles:
The magnitude-distance filtering is a list of pairs (magnitude, distance) as the name says. But I am sure there will be users who will use the wrong ordering, like

maximum_distance = [(200, 8), (150, 7), (100, 6)]  # I want 200 km for magnitude 8, 150 km for 7, ...

which will be interpreted by the engine like 8 km for magnitude 200 etc, essentially filtering away most of the sites and giving bogus results without any error. So I would add a check on the magnitude. What about being a floating point in the range [2., 10.] ? Or [3., 10.]? What should be the minimum acceptable magnitude? 10. looks right for the upper limit.

mmpagani:
Let’s put -1 to 11

micheles:
-1??

micheles:
I have update the manual too. Please have a look here:
https://ci.openquake.org/job/builders/job/pdf-builder/61/artifact/oq-engine/doc/manual/oq-manual.pdf

PR #3163, opened 2017-10-19
Added a check on missing `source_model_logic_tree`

micheles:
This happened to Cata when running the event based risk demo `job_risk.ini` without having run the hazard first.


CatalinaYepes:
I personally prefer a message that gives more detail about the problem, for example: "job_risk.ini missing the hazard component. Specify hazard parameters, or provided an existing hazard calculation"

micheles:
Now the error message is `Missing source_model_logic_tree in job.ini or missing --hc option`. It is very difficult to distinguish between the two cases reliably, so I am leaving it as it is. It is better than before.

PR #3166, opened 2017-10-20
Fixed reading order in config.read

micheles:
The ConfigParser manages the .ini files in such a way that the last one wins, i.e. in the opposite order with respect to how it is implemented now (I simply forgot how it worked). Now it works correctly and in absence of virtualenv it is also able to read $HOME/openquake.cfg (if present) which takes the precedence over /etc/openquake,cfg but not with respect to OQ_CONFIG_FILE.

NB: currently things work well enough because `engine/openquake.cfg` is read first so it has the least precedence and `/etc/openquake.cfg` or the one in the virtualenv are read later and win.
OQ_CONFIG_FILE does not work in engine 2.7 since it is added first in the list, so with the least precedence while it should have the highest precedence.

### DIFF
--- a/openquake/baselib/__init__.py
+++ b/openquake/baselib/__init__.py
@@ -39,11 +39,14 @@ class DotDict(collections.OrderedDict):
 
 
 config = DotDict()  # global configuration
+d = os.path.dirname
+fallback = os.path.join(d(d(__file__)), 'engine', 'openquake.cfg')
 if 'VIRTUAL_ENV' in os.environ or hasattr(sys, 'real_prefix'):
     config.paths = [
-        os.path.join(os.environ.get('VIRTUAL_ENV', '~'), 'openquake.cfg')]
+        os.path.join(os.environ.get('VIRTUAL_ENV', '~'), 'openquake.cfg'),
+        fallback]
 else:  # installation from packages, search in /etc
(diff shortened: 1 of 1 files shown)


openquake/baselib/__init__.py
```diff
@@ -39,14 +38,17 @@ def __getattr__(self, key):
 
 
 config = DotDict()  # global configuration
-if 'VIRTUAL_ENV' in os.environ or hasattr(sys, 'real_prefix'):
```
daniviga: Why you removed `hasattr(sys, 'real_prefix')`

daniviga: Anyway it looks like the best way is https://stackoverflow.com/a/42580137

daniviga: This should solve also the issue when you call directly executables in the venv without loading it first.

micheles: I undestood `sys.real_prefix` was needed for old versions of virtualenv and not needed anymore.

daniviga: You're right, it works with `virtualenv` but not with `venv`. Anyway https://stackoverflow.com/a/42580137 seems the best solution. But for now I'm fine.

PR #3174, opened 2017-10-23
gmf_ebrisk: fixed rlzi in the agg_loss_table

micheles:
The bug was discovered by Catarina Costa. Also, the number of generated rows was wrong.


catarinaqmcosta:
LGTM!

PR #3175, opened 2017-10-23
Changed images on the page server.md

hascar:
I changed the gif images with the new images on the page server.md and i created a folder with the old gif images


micheles:
I would change `The possible formats are xml, csv, geojson, zip` with `The possible formats are xml, csv, npz, zip`. Apart from that, LGTM

daniviga:
Please remove all the `old_image` content

PR #3176, opened 2017-10-23
Improve venv detection

### LINKED ISSUE #3167: Make identification of a `venv` more robust
It's currently done via `if 'VIRTUAL_ENV' in os.environ` but this does not work when an entry point in the `venv` is called directly, i.e.:

```bash
$ ~/venv/bin/oq engine --run job.ini
```
or via a `systemd --user` unit (happened already on the VM for Colombia).

Can be resolved in this way:

```python
import sys

def is_venv():
    return (hasattr(sys, 'real_prefix') or
            (hasattr(sys, 'base_prefix') and sys.base_prefix != sys.prefix))
```
(from https://stackoverflow.com/questions/1871549/python-determine-if-running-inside-virtualenv/42580137#42580137)

daniviga:
Closes #3167 

`sys.prefix` always reflects the used venv (even when an entry point is used and the venv not loaded):

```bash
# VIRTUAL_ENV is _not_ set
$ echo $VIRTUAL_ENV
>
# systemwide python
$ python3 -c 'import sys;print(sys.prefix)'
> /usr
# made with `virtualenv`
$ ~/py27/bin/python -c 'import sys;print(sys.prefix)'
> /home/daniele/py27
# made with `venv` 
$ ~/py35/bin/python -c 'import sys;print(sys.prefix)'
> /home/daniele/py35
```

Systemd units will be happier.

### DIFF
--- a/openquake/baselib/__init__.py
+++ b/openquake/baselib/__init__.py
@@ -17,6 +17,7 @@
 # along with OpenQuake. If not, see <http://www.gnu.org/licenses/>.
 
 import os
+import sys
 import collections
 from openquake.baselib.python3compat import configparser
 from openquake.baselib.general import git_suffix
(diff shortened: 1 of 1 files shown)


micheles:
LGTM

PR #3187, opened 2017-10-25
Keep the realization weights even when sampling

micheles:
As requested by M. Pagani.


micheles:
We can always expand the test coverage, but the fact that I had to change the expected files for one test means that we are already sensible to changes in the weights, so the situation is not bad.

PR #3189, opened 2017-10-25
Removed the deprecated outputs losses_by_tag and losses_total

micheles:
As requested in https://github.com/gem/oq-engine/issues/3001


raoanirudh:
The page on the [engine outputs](https://github.com/gem/oq-engine/blob/master/doc/engine-outputs.rst) needs to be updated with these changes.

micheles:
Done

PR #3191, opened 2017-10-26
Changing the management of the realization weights again

micheles:
In the past the weights were considered when sampling the realizations but not when computing the statistics; now we are switching to a logic where the weights are considered when computing the statistics but ignored during the sampling (for the gsim logic tree). The goal is to treat the weights more correctly when the number of samples is smaller than the size of the logic tree, like as for the India model, where the gsim logic tree has a (full) size of 245,760 realizations.

NB: for the moment I am not changing the sampling logic for the source model logic tree, only for the gsim logic tree. Changing also for the source model logic tree would be too big of a change, and also it is not needed for models like the India model, when there are 3 models and it is easy to set `number_of_logic_tree_samples >> 3`.

NB: there are no changes at all for the case of full enumeration.

FYI, here is the source model data for the India model with 100 samples and weights 0.4, 0.27, 0.33:
```
name                      branch_id    num_samples
areal_collapsed.xml                     b1m1    45
NT2012_smoothed_collapsed_mmin4.5.xml	b1m2	27
NT2012_smoothed_collapsed_mmin5.5.xml	b1m3	28
```


micheles:
Not only tests. We need to run some real computation and perform an analysis of the dependence of the results from the sampling. For instance, given a site, how much the mean hazard curve change if we change the seed and we get different realizations? At which point there enough realizations that this dependency becomes reasonably small? There a lot of things to try out. Essentially all the real life computations that I have seen on our cluster were using full enumeration, not sampling. I do not think we have much experience with sampling in big models.

PR #3194, opened 2017-10-26
number_of_ground_motion_fields back to being optional if there is a file with the GMFs

micheles:
As requested by Mabe'. It is inferred from the GMF file. The user can still add it to the job.ini file, in this case there is a check that the parameter is consistent with the number of GMFs found in the file. It is a good practice to set the parameter, since in this way truncated GMFs files will raise a clear error (i.e. the file contains less lines than expected).


micheles:
For `gmf_ebrisk` I prefer to keep this parameter mandatory, since the calculator is still experimental and ma ny things can go wrong. For scenario_risk I have fixed the issue

PR #3196, opened 2017-10-26
Fixed yet another bug in gmf_ebrisk

micheles:
Catarina discovered that for zero losses the saved event loss table was missing the event IDs (everything was zero). This PR solves the issue and adds a lot more checking on the GMF csv file. In particular the event IDs must be ordered from 0 to E - 1 and must all be present. This assumption will have to be relaxed for GMFs coming from an event based, but this is left for future developments.
A test adapted from Catarina's reduced calculation has been added.


catarinaqmcosta:
Looks good!

PR #3199, opened 2017-10-27
Revoke functionality for the zmq workers

micheles:
This is stil a work in progress. Here is the current plan of work.

1. use setproctitle so that each running job has a name `oq-job-ID`
2. add a command `oq abort job_id`
3. add a handler for SIGTERM to the main process: when received, a JobAborted exception is
    raised if the job has status 'aborted' and then the job dies
4. add the same handler to the zmq workers: then when they get a SIGTERM they raise a JobAborted
   exception which is trapped and sent back to the main process. In this way the tasks die early, but
   the zmq worker does not die and does not need to be restarted.

`oq abort job_id` gets the pid from the job name by using psutil, changes the db status to 'aborted' and then send to the process the signal SIGTERM. Moreover, it sends to all zmq workers a SIGTERM.


micheles:
Superseded by https://github.com/gem/oq-engine/pull/3473

Issue #2362, opened 2016-11-16
Add a command oq abort <calc_id>


micheles:
In order to implement that, we need to store the pid in the job table or to be able to extract the pid from the process name. Also, we should enable the feature only if the job is running and only if it belongs to the user sending the kill command.
The tricky case is for processes started from the Web UI.

micheles:
It is fine to implement it only for zmq.

PR #3200, opened 2017-10-28
WebUI: now every job runs in its own process and has name oq-job-<ID>

micheles:
The goal is to make it easy to kill jobs started from the web UI (see https://github.com/gem/oq-engine/issues/2362). We are exposed to abuses, like people submitting hundreds of jobs at the same time, but so be it for the moment. If this becomes a real problem it is easy to keep track of how many calculations are running concurrently and put a limit. Notice that when there is task queue (celery or zmq) there are no issues for the parallel part of the computation, only for the initial part of the computation which can run out of memory if there are too many jobs. cgroups or other system tricks may help here.

Jenkins is running: https://ci.openquake.org/job/zdevel_oq-engine/2617/


daniviga:
It does not work for me. The job is tuck in "created" status, a process with the job name is allocated, but nothing is actually running.

I also noticed that when I stop the WebUI process (`oq webui start`) with `ctrl-c` all these `oq-job-<id>` processes are left alive and the first of them takes the `8800` port over.

daniviga:
- [ ] Check with gunicorn
- [ ] Check on macOS
- [ ] Check on Windows

micheles:
As it is now, killing the WebUI does not kill the processes spawned by it. This will be fixed in a future PR, after https://github.com/gem/oq-engine/pull/3199 is merged

PR #3203, opened 2017-10-31
Add a post-zip command in utils

daniviga:
A very useful script provided by @micheles to easily debug the API server


micheles:
I have added `sap` support. In this way if we wanted, this script could become another `oq` command (but I would not do it for the moment).

LGTM

PR #3212, opened 2017-11-03
Abort functionality for futures

micheles:
Implementing the abort functionality for zmq (https://github.com/gem/oq-engine/pull/3199) proved to be difficult, but I do not see any issue with the case of futures, even for large calculations. Perhaps the problems seen by Daniele in the past have disappeared. To test run a large calculation that may run out of memory and call `oq abort -1`.
Closes https://github.com/gem/oq-engine/issues/3007 and https://github.com/gem/oq-engine/issues/2362.


daniviga:
> Perhaps the problems seen by Daniele in the past have disappeared

The problem was related to `CTRL-C` where the main process was terminated before the loop on the spawned processes was completed.

micheles:
I cannot reproduce the problems with CTRL-C, at least with Python 3.

openquake/engine/engine.py
```diff
@@ -163,15 +163,11 @@ def raiseMasterKilled(signum, _stack):
...
-
-    # FIXME this code has been temporary disabled due issues with large
-    # computations and further investigation is need; code is left as reference
-    # for pid in parallel.executor.pids:
-    #     try:
-    #         os.kill(pid, signal.SIGKILL)
-    #     except OSError: # pid not found
-    #         pass
-
+    for pid in parallel.executor.pids:  # when using futures
```
daniviga: I'm experiencing  issues with Python 2: `oq abort` does not do anything and this is preventing `CTRL-C` from work (and `SIGKILL` is needed. Everything is fine with Python 3: I would suggest to skip this block if Python 2.7 is used. Soon or later Python 2 support will be removed :)

PR #3226, opened 2017-11-08
Removed the `poe` from the disaggregation outputs if `iml_disagg` is set in the job.ini

micheles:
Because it was giving confusing values. This was noticed by Robin. Now each output in the datastore does not contain the `poe` value anymore; however, it does contain a list `poe_agg` which is computed by using the formula
```
poe_agg = 1 - numpy.prod(1 - poe_array)
```
for each `poe_array` in the output. There are 8 values in the list since there are 8 arrays (Dist, Lon_Lat, Lon_Lat_TRT, Mag, Mag_Dist,Mag_Dist_Eps,Mag_Lon_Lat,TRT). If not for approximation errors all 8 values should give the same poe.


rcgee:
I ran a test disaggregation calculation with these changes. Each output in the datastore no longer contains the poe value; however, I cannot locate the list poe_prod in the hdf5 file.

micheles:
It is an attribute of the outputs.

rcgee:
OK, I see it now, thank you. LGTM

micheles:
NB: I have changed the name from `poe_prod` to `poe_agg` which I like more.

PR #3227, opened 2017-11-09
If `iml_disagg` is given, forbid `intensity_measure_types_and_levels`

micheles:
Because it is automatically inferred from `iml_disagg`. It is better to have the information in a single point rather than having it in two places and then check that the values are the same.
Also, now hazard curves with a single IML (which is the case for `iml_disagg`) are valid.


klunk386:
LGTM

PR #3228, opened 2017-11-09
Add a check for `poes_disagg` too big

micheles:
This happened in a test model provided by F. Martinelli. Now a clear error message is raised:

```python
ValueError: You are trying to disaggregate for poe=0.1.
However the source model #0, 'source_model_test_complex.xml',
produces at most probabilities of 0.0362320992703 for rlz=#0, IMT=PGA.
The disaggregation PoE is too big or your model is wrong,
producing too small PoEs.
```

The check is performed after the classical part of the calculation, but before the disaggregation starts.


openquake/calculators/disaggregation.py
```diff
@@ -191,9 +201,14 @@ def full_disaggregation(self):
...
             logging.info('%d mag bins from %s to %s', len(mag_edges) - 1,
                          min_mag, max_mag)
-            for sid, site in zip(sitecol.sids, sitecol):
-                curves = curves_dict[sid]
-                if not curves:
+            for i, site in enumerate(sitecol):
+                sid = sitecol.sids[i]
+                curve = curves[i]
+                for rlzi, imt in curve:
+                    maxc = curve[rlzi, imt].max()
```
mmpagani: I think lines 208-211 can be also written as
max_poe[rlzi][imt] = max(max_poe[rlzi][imt], curve[rlzi, imt].max())

micheles: Done

PR #3230, opened 2017-11-09
Update the documentation for Python3 installers

daniviga:
Also remove redundancy, we now suggest only Python 3.5 for development.


micheles:
LGTM

PR #3238, opened 2017-11-13
Added a test disagg/case_master

micheles:
We were missing a test for the case of a complex logic tree. I added one by adapting classical_risk/case_master.


openquake/qa_tests_data/disagg/case_master/expected/mean_disagg.rst
```diff
@@ -0,0 +1,68 @@
+=================================== ========= ========= =========== ========= ========= ============ =========== =========
+key                                 Dist      Lon_Lat   Lon_Lat_TRT Mag       Mag_Dist  Mag_Dist_Eps Mag_Lon_Lat TRT      
+=================================== ========= ========= =========== ========= ========= ============ =========== =========
+poe-0.01-rlz-0-PGA--121.5-38.3      7.618E-04 0.001     5.714E-04   4.572E-04 1.524E-04 3.810E-05    2.286E-04   0.001
```
ptormene: Please note that, in this file, TRT is always equal to Lon_Lat. Is it correct?

micheles: I am not sure that it is correct. It could be a long standing bug. I will investigate.

PR #3240, opened 2017-11-13
Changed `oq extract <calc_id> hazard/rlzs` to generate an .hdf5 file suitable for the  hazard scientists

micheles:
Since engine 2.8 we have a command  `oq extract <calc_id> hazard/rlzs` to extract hazard curves and maps from the datastore. The format used is designed for future use in the QGIS plugin and it is not convenient for hazard scientists. In this PR I change it to produce a more suitable .hdf5. The old command is kept, but now it is spelled `oq extract <calc_id> qgis-hazard/rlzs`.

Moreover, I am simplifying the `PmapGetter` class, i.e. the class used to extract programmatically the hazard curves and maps, so that code like the following works:

```python
from openquake.baselib.datastore import read
from openquake.commonlib.calc import PmapGetter, make_hmap
dstore = read('/home/michele/oqdata/calc_20683.hdf5')  # an example
oqparam = dstore['oqparam']
pgetter = PmapGetter(dstore)
rlzs = dstore['realizations'].value
for r, rlz in enumerate(rlzs):
   pmap = pgetter.get(r)
   hmap = make_hmap(pmap, oqparam.imtls, oqparam.poes)
   print(rlz, pmap, hmap)
```


g-weatherill:
LGTM

PR #3246, opened 2017-11-14
Improved the error message if the user forgets both sites and sites.csv

micheles:
In a calculation starting from predetermined GMFs. This happened to Mabe'.

### DIFF
--- a/openquake/commonlib/oqvalidation.py
+++ b/openquake/commonlib/oqvalidation.py
@@ -230,6 +230,11 @@ class OqParam(valid.ParamSet):
             raise ValueError('asset_correlation != {0, 1} is no longer'
                              ' supported')
 
+        # check for GMFs from file
+        if ('gmfs' in self.inputs and not self.sites and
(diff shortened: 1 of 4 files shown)


openquake/commonlib/oqvalidation.py
```diff
@@ -230,6 +230,11 @@ def __init__(self, **names_vals):
             raise ValueError('asset_correlation != {0, 1} is no longer'
                              ' supported')
 
+        # check for GMFs from file
+        if (self.inputs.get('gmfs', '').endswith('.csv') and not self.sites and
+                'sites' not in self.inputs):
+            raise ValueError('You forgot sites|sites_csv in the job.ini!')
```
raoanirudh: Job configuration file to be more general than `job.ini`?

micheles: I want to avoid confusions with the openquake.cfg configuration file. Perhaps I should just say `job .ini file`.

PR #3250, opened 2017-11-15
Event based

nastasi-oq:
Read list of web applications from `openquakeplatform.settings`
Tests are green here: https://ci.openquake.org/job/zdevel_oq-engine/2621/

### DIFF
--- a/openquake/server/settings.py
+++ b/openquake/server/settings.py
@@ -25,6 +25,7 @@ try:
     from openquakeplatform.settings import STANDALONE, STANDALONE_APPS
 except ImportError:
     STANDALONE = False
+    STANDALONE_APPS = ()
 
 INSTALLED_APPS = ('openquake.server.db',)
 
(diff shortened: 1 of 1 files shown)


openquake/server/settings.py
```diff
@@ -22,8 +22,7 @@
 from openquake.baselib import config, datastore
 
 try:
-    import openquakeplatform
-    STANDALONE = True
+    from openquakeplatform.settings import STANDALONE, STANDALONE_APPS
```
daniviga: `STANDALONE_APPS` must exist anyway in this configuration, because it's used here https://github.com/gem/oq-engine/blob/d6d8c6f3d157e7375ae40ae567d600d16eaa73d8/openquake/server/urls.py#L39

So, or we do something like:
```python
STANDALONE = False
STANDALONE_APPS = ()

from openquakeplatform import settings as standalone_settings

STANDALONE = standalone_settings.STANDALONE
STANDALONE_APPS = standalone_settings.STANDALONE_APPS
```

or we have to change `urls.py` too

nastasi-oq: Read list of web applications from openquakeplatform.settings.

PR #3263, opened 2017-11-20
Optimized the sending of the sources in the classical calculator

micheles:
Instead of sending the sources by SourceGroup, we can send them by tectonic region type, except for mutex sources. This makes the operation `managing sources` a lot faster. In the case of SHARE we went down from 3,159s to 1,229s. Also, the tasks become more homogeneous.

PS: while at it, I have also decouple the EventBasedCalculator from the ClassicalCalculator: there is no more an inheritance relationship. This is a lot cleaner conceptually and makes changing the calculators easier, at the cost of adding a few lines of code.

### DIFF
--- a/openquake/calculators/classical.py
+++ b/openquake/calculators/classical.py
@@ -200,7 +200,7 @@ class PSHACalculator(base.HazardCalculator):
         for trt, sources in csm.get_sources_by_trt(opt).items():
             gsims = csm.info.gsim_lt.get_gsims(trt)
             self.csm.add_infos(sources)  # update self.csm.infos
-            for block in csm.split_sources(maxweight, sources):
+            for block in csm.split_in_blocks(maxweight, sources):
                 yield block, csm.src_filter, gsims, param
 
     def post_execute(self, pmap_by_grp_id):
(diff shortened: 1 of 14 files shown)


openquake/calculators/classical.py
```diff
@@ -184,52 +181,37 @@ def gen_args(self, csm, monitor):
...
+    def _args_by_grp(self, csm, param, num_tiles, maxweight):
+        for sg in csm.src_groups:
+            if sg.src_interdep == 'mutex':
+                gsims = csm.info.gsim_lt.get_gsims(sg.trt)
+                self.csm.add_infos(sg.sources)  # update self.csm.infos
+                yield sg, csm.src_filter, gsims, param
+
+    def _args_by_trt(self, csm, param, num_tiles, maxweight):
+        opt = self.oqparam.optimize_same_id_sources
+        for trt, sources in csm.get_sources_by_trt(opt).items():
```
ptormene: Please add a comment about the fact that _args_by_grp considers all sources and it needs to select those that are mutex, whereas _args_by_trt uses get_sources_by_trt that only selects sources that are not mutex

PR #3271, opened 2017-11-21
SGS model

klunk386:
I have introduced a couple of subclasses for ZhaoEtAl2006Asc and AtkinsonBoore2006 to account for the distance filter used by SGS in their PSHA model for Saudi Arabia.
Distance threshold is hard coded to 5km in this implementation.

### DIFF
--- a/openquake/hazardlib/gsim/atkinson_boore_2006.py
+++ b/openquake/hazardlib/gsim/atkinson_boore_2006.py
@@ -32,6 +32,7 @@ import numpy as np
 # standard acceleration of gravity in m/s**2
 from scipy.constants import g
 from math import log10
+import copy
 
 from openquake.hazardlib.gsim.boore_atkinson_2008 import BooreAtkinson2008
 from openquake.hazardlib.gsim.utils import (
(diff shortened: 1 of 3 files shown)


openquake/hazardlib/gsim/atkinson_boore_2006.py
```diff
@@ -530,3 +530,25 @@ def _get_stress_drop_scaling_factor(self, magnitude):
...
+    by introducing a distance filter for the near field, as implemented
+    by SGS for the national PSHA model for Saudi Arabia.
+    """
+
+    def get_mean_and_stddevs(self, sites, rup, dists, imt, stddev_types):
+        """
+        Using a minimum distance of 5km for the calculation.
+        """
+
+        dists.rrup = np.array([d if d > 5. else 5. for d in dists.rrup])
```
mmpagani: Perhaps it's faster this: 
dists.rrup[dists.rrup<=5] = 5.

klunk386: I tried your suggestion, but this has the drawback to modify also the original reference to the distance variable I'm using when calling the function, which is undesirable to me.
In oder words, the input distance array is returned modified to [5., 5., 5., 5., 10., 20., .....]
For reasons not fully clear to me, the first implementation doesn't do it.

mmpagani: ok. I don't understand why it's a problem if dists.rrup is modified.

nastasi-oq: @klunk386 I wrote this small example to check your assertion about the two methods and, if I understood correctly, the drawback occur also with your solution (luckily, because an object is mutable so you can change back its attributes values). To test it try `fun` and than `fun2` 
```python
#!/usr/bin/env python
import numpy as np


class Pippo():
    def __init__(self):
        self.rrup = None


def fun(pi):
    pi.rrup[pi.rrup <= 5] = 5


def fun2(pi):
    pi.rrup = np.array([d if d > 5. else 5. for d in pi.rrup])
    print("inner %s" % pi.rrup)


if __name__ == '__main__':
    pi = Pippo()
    pi.rrup = np.array([1., 2., 3., 4., 5., 6., 7., 8., 9.])
    print(pi.rrup)
    fun(pi)
    print(pi.rrup)
```

micheles:
LGTM, but you need to update the changelog.

PR #3274, opened 2017-11-22
Improved prefiltering in tiling calculations

micheles:
Right now the sources are prefiltered (i.e. filtered with rtree) once per each tile; moreover, they are prefiltered against the whole site collection once at the beginning. With this branch this initial redundant prefiltering is removed.


micheles:
Superseded by https://github.com/gem/oq-engine/pull/3291

PR #3276, opened 2017-11-22
Added a check on poe_agg

micheles:
The aggregation of the probabilities in a disaggregation output should be close (apart from numerical errors) to the probabilities in poes_disagg. However if there are few intensity measure levels this is not necessarily true. We are printing a warning in this case. More precisely I am printing it if poe_agg is not within 10% from the expected poe. Notice that the disaggregation demo was printing the warning, so I changed the levels to make the warning disappear. The warning is kept in case_master, to exercise that part of the code.

Thanks for Robin for help in investigating the issue.

NB: the hazard curve interpolation in the disaggregation calculator, which is the same interpolation we do when computing the hazard maps, seems to be numerically very delicate. The output values in case_master are completely different having changed only the number of levels. The lesson is that disaggregation outputs should be taken with a grain of salt.


rcgee:
LGTM

PR #3284, opened 2017-11-25
Sort taxonomies before comparison in the BCR calculator

### LINKED ISSUE #3283: Incorrect comparison of original/retrofitted taxonomies in the BCR calculator
For the BCR calculator, [the check](https://github.com/gem/oq-engine/blob/engine-2.7/openquake/risklib/riskinput.py#L110-L111) comparing taxonomies from the original and retrofitted vulnerabilities (see snippet below) may fail because python doesn't iterate through the dictionary keys in sorted order, even when both vulnerability models have exactly the same imts and taxonomies.

```python
# classical_bcr calculator
for (taxonomy, vf_orig), (taxonomy_, vf_retro) in \
        zip(rmdict.items(), retrodict.items()):
    assert taxonomy == taxonomy_  # same imt and taxonomy
    self._riskmodels[taxonomy] = riskmodels.get_riskmodel(
        taxonomy, oqparam,
        vulnerability_functions_orig=vf_orig,
        vulnerability_functions_retro=vf_retro)
```

raoanirudh:
Resolves #3283

The vulnerability model for the retrofitted assets in the BCR demo has been updated, so that the demo can act as a test for this case.


micheles:
LGTM, but please add a line to the changelog

PR #3290, opened 2017-11-27
Raise an error if both `iml_disagg` and `poes_disagg` are given

micheles:
Those are exclusive modalities of the disaggregation calculator. Before `poes_disagg` was accepted but ignored in the presence of `iml_disagg`. Now there is less room for confusion.

PS: while at it, I have also replaced several ValueError with InvalidFile errors.


rcgee:
LGTM

PR #3291, opened 2017-11-28
Refactored the classical calculator

micheles:
I am restore the task `classical` which in turns calls `pmap_from_grp` or `pmap_from_trt`. The speed is the same as before.


openquake/commonlib/oqvalidation.py
```diff
@@ -98,6 +98,7 @@ class OqParam(valid.ParamSet):
     num_epsilon_bins = valid.Param(valid.positiveint)
     poes = valid.Param(valid.probabilities, [])
     poes_disagg = valid.Param(valid.probabilities, [])
+    prefilter = valid.Param(valid.boolean, True)
```
ptormene: Please check if this is needed

PR #3293, opened 2017-11-29
Fixed tectonic region types in the magnitude-distance filter example

micheles:
Ruptures in Stable Continental Crust have longer integration distances.


rcgee:
LGTM

Issue #3296, opened 2017-12-01
See if we can improve the (pre)filtering


micheles:
Right now the prefiltering gives very rough information. For instance for a source in a SHARE calculation
one can get `<AreaSource 291> <FilteredSiteCollection with 4123 of 23786 sites>` and  10387 ruptures, but the fine filtering says that there are only 1807 sites affected and 5444 ruptures. This is why the task weights are all wrong and thus there are slow tasks.

micheles:
I improved with I could with the 5 pull requests before. The problem at its root is still unsolved, though :-(

PR #3297, opened 2017-12-01
Some refactoring of hazardlib.calc.hazard_curve

micheles:
This is a pure refactoring without any performance effect, but it will be useful for future progress on the filtering mechanism. Some functions originally in calc_hazard_curve are now methods of gsim.base.ContextMaker. Also, now we are monitoring the total time spent in `get_poes` and not the individual times for each GSIM (it was too much info). This is a first step towards https://github.com/gem/oq-engine/issues/3296.


openquake/hazardlib/gsim/base.py
```diff
@@ -267,6 +271,98 @@ def make_contexts(self, site_collection, rupture, filter=True):
...
+            etype, err, tb = sys.exc_info()
+            msg = '%s (source id=%s)' % (str(err), src.source_id)
+            raise_(etype, msg, tb)
+        return pmap
+
+    # NB: it is important for this to be fast since it is inside an inner loop
+    def _make_pnes(self, rupture, imtls, trunclevel):
+        """
+        :param rupture: a Rupture instance
+        :param sctx: the corresponding SiteContext instance
```
ptormene: Please fix the docstring

PR #3300, opened 2017-12-02
Moved split_source in hazardlib.source

micheles:
The source splitting logic was half in hazardlib and half in commonlib. Now it is all in hazardlib. The performance is only marginally better in models with large logic trees, but this helps towards https://github.com/gem/oq-engine/issues/3296.

NB: the reason why the numbers change slightly is the line `srcs = sum([split_source(src) for src in sources], [])`: now the sources are split in the workers before filtering them. That improves the filtering operation and `make_contexts` becomes faster.


openquake/hazardlib/gsim/base.py
```diff
@@ -273,7 +273,7 @@ def make_contexts(self, site_collection, rupture, filter=True):
 
     def filter_ruptures(self, src, sites):
         """
-        :param src: a source object
+        :param src: a source object, already filtered and split
         :param sites: a FilteredSiteCollection
         :param mon: a Monitor instance
```
ptormene: Please remove the param `mon` from the docstring

micheles: Done

PR #3305, opened 2017-12-06
 Add macOS 10.13 High Sierra to list of supported OSes [skip CI]

raoanirudh:
So far, I haven't noticed any issues installing and running the engine on High Sierra since upgrading in September


micheles:
What about the test that was hanging? The one I suggest to run with zmq?

micheles:
Also, several tests on macOS are skipped because the numbers are different from Linux for various reasons, which we never really investigated.

PR #3307, opened 2017-12-07
Fixed a bug when saving the disaggregation bins

micheles:
This only happens if the number of bins is different among sites. I have added a test for that situation.


openquake/qa_tests_data/disagg/case_4/job.ini
```diff
@@ -1,6 +1,6 @@
 [general]
 
-description = Disaggregation Hazard Demo
+description = Disaggregation with sampling
```
rcgee: "with sampling" refers to what? could you be more specific here

rcgee:
besides the small comment LGTM

openquake/qa_tests_data/disagg/case_4/job.ini
micheles: There is only one kind of sampling, i.e. setting number_of_logic_tree_samples > 0. No room for confusion.

PR #3309, opened 2017-12-07
Called oq-webui the webui process

micheles:
Let's use setproctitle!


micheles:
I thought setproctitle was mandatory now but I was wrong. Fixed.

Issue #3270, opened 2017-11-21
Change the _get_intra_event_std function of ASK14 (GMPE)


g-weatherill:
In the Abrahamson et al (2014) GMPE the `_get_intra_event_std` function raises an error if the `phi_al` term is less than the `phi_amp` term to avoid taking the square root of a negative number (https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/gsim/abrahamson_2014.py#L364). This situation happens for IMTs with periods longer than 5 and for magnitudes smaller than about 4.3 - 4.4. 

The authors of the GMPE have addressed this in their hazard code, and simply set `phi_amp` to be `0.99 * phi_al` in this circumstance (https://github.com/abrahamson/HAZ/blob/master/HAZ%20files/NGAWest2.F#L3098). As raising an error causes the calculation to fail for reasons that might be cryptic to the users, I recommend to adopt the same adjustment here rather than raising an error.

mmpagani:
I’m fine with the suggested adjustment.
m

PR #3316, opened 2017-12-13
Fixes ASK14 GMPE behaviour to remove ValueError

g-weatherill:
Closes https://github.com/gem/oq-engine/issues/3270

When presented with small magnitudes at long periods the Abrahamson et al. (2014) GMPE provides erroneous (i.e. complex) values for intra-event standard deviation. Currently this is caught by raising a ValueError, breaking the calculation. However, as noted in https://github.com/gem/oq-engine/issues/3270 the GMPE authors themselves catch this case by reducing the `phi_amp` term so that it is fractionally smaller than `phi_al`, preventing this from happening. The PR implements this adjustment.


micheles:
LGTM on my part

mmpagani:
LGTM

PR #3329, opened 2017-12-20
Allow more control over figures in HMTK

nackerley:
I have made some of the HMTK plotting routines more flexible, primarily by allowing the user to pass a Axes object into which the data is to be plotted, allowing multiple similar plots to be plotted into the same Figure, for example. 

A few other minor improvements:
- Made colouring of lines in create_stepp_plot() predictable instead of random (@g-weatherill)
- Made filename optional in create_stepp_plot()
- Removed some problematic hardcoding of fontsize
- Made latitude/longitude grid spacing configurable in HMTKBasemap.
- Added simplify() method to Stepp1971 which removes redundant entries from automatically-generated completeness tables, making Stepp plots more readable

It is tricky to test plotting routines - there are no HMTK plot tests to update - so I haven't added any. I have informally tested everything I am pushing, at least. I could add a test for the simplification of completeness tables, if the maintainer thinks that would be helpful.

### DIFF
--- a/openquake/hmtk/plotting/mapping.py
+++ b/openquake/hmtk/plotting/mapping.py
@@ -96,7 +96,8 @@ class HMTKBaseMap(object):
     Class to plot the spatial distribution of events based in the Catalogue
     imported from openquake.hmtk.
     '''
-    def __init__(self, config, title, dpi=300, lat_lon_spacing=2.):
+    def __init__(self, config,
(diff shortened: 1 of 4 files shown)


micheles:
The tests are green, LGTM too. Merry Christmas!

nackerley:
And happy new year!

PR #3336, opened 2018-01-08
On same fs

nastasi-oq:
This PR provide a new server command 'v1/on_same_fs' to be able to verify if a provided filename is accessible and the first 32 bytes has the same checksum of the provided checksum argument.

### DIFF
--- a/doc/web-api.md
+++ b/doc/web-api.md
@@ -221,7 +221,7 @@ a JSON object, containing:
 
 #### POST /v1/on_same_fs
 
-Check if a given filename exists and if first 32 bytes of its content  has the same checksum passed as argument of POST.
+Check if a given filename exists and if the first 32 bytes of its content have the same checksum passed as argument of POST.
 
 Parameters:
 
--- a/doc/web-api.md
+++ b/doc/web-api.md
@@ -219,6 +219,22 @@ a JSON object, containing:
     * error_line: line of the given XML where the error was found (None if no error was found or if it was not a validation error)
 
 
+#### POST /v1/on_same_fs
+
+Check if a given filename exists and if first 32 bytes of its content  has the same checksum passed as argument of POST.
+
+Parameters:
+
+    * filename: filename (with path) of file to be checked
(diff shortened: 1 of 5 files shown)


openquake/server/views.py
```diff
@@ -700,6 +701,34 @@ def web_engine_get_outputs(request, calc_id, **kwargs):
...
+
+    checksum = 0
+    try:
+        data = open(filename, 'rb').read(32)
+        checksum = zlib.adler32(data, checksum) & 0xffffffff
+        if checksum == int(checksum_in):
+            return HttpResponse(content=json.dumps({'success': True}),
+                                content_type=JSON, status=200)
+    except:
+        pass
```
micheles: I guess the expected error here is an IOError when the user passes a file path which does not exist. So I would trap only IOError.

micheles:
Please solve the conflict on the changelog.

openquake/server/tests/tests.py
```diff
@@ -282,3 +286,35 @@ def test_validate_nrml_missing_parameter(self):
         self.assertEqual(resp.status_code, 400)
         self.assertEqual(resp.content,
                          b'Please provide the "xml_text" parameter')
+
+    def test_check_fs_access(self):
+        with tempfile.NamedTemporaryFile(buffering=0, prefix='oq-test_') as f:
+            filename = f.name
```
micheles: Using buffering works only in Python 3. Should we block this until we have Python 3 packages?

doc/web-api.md
```diff
@@ -219,6 +219,22 @@ a JSON object, containing:
     * error_line: line of the given XML where the error was found (None if no error was found or if it was not a validation error)
 
 
+#### POST /v1/on_same_fs
+
+Check if a given filename exists and if first 32 bytes of its content  has the same checksum passed as argument of POST.
```
micheles: `if first 32 bytes of its content  has` -> `if the first 32 bytes of its content have`

Issue #2787, opened 2017-05-10
Ruptures exported in the file ruptures.xml must be usable in the scenario calculator


micheles:
We are missing a test.
NB: this only works for singlePlaneRuptures.

micheles:
Not doing this until this becomes an official feature of the engine.

PR #3339, opened 2018-01-10
Rupture loss table

micheles:
With this PR I store information about the impact of each rupture in an event based risk calculation (the impact is the sum of all the losses generated by the rupture, for all realizations). I am also storing the indices of the ruptures with highest impact for each loss type. The operation is efficient, a `rup_loss_table` of 1 million ruptures can be computed and stored in less than 10 minutes, so we are not slowing down the calculator much. This is preliminary for https://github.com/gem/oq-engine/issues/2787


micheles:
Now the engine produces an output like this:
```
  id | name
 525 | agg_curves-rlzs
 526 | agg_curves-stats
 527 | agg_loss_table
 528 | avg_losses-rlzs
 529 | avg_losses-stats
 530 | fullreport
 531 | loss_maps-rlzs
 532 | loss_maps-stats
 533 | realizations
 534 | rup_loss_table
```
Should I hide the rup_loss_table output and keep it secret?

openquake/commonlib/calc.py
```diff
@@ -618,20 +632,17 @@ class RuptureGetter(object):
     :param grp_id: the group ID of the ruptures (default: all)
     :param rup_id: a specific rupture (default: all)
```
ptormene: Please remove rup_id description from the docstring

Issue #3338, opened 2018-01-10
Honor disagg_outputs


micheles:
As it is now it used only at export time, too late. We should use it to avoid extracting and storing the unwanted disaggregation outputs.

PR #3341, opened 2018-01-10
Use disagg_outputs

micheles:
Closes https://github.com/gem/oq-engine/issues/3338.  Several thousand of lines were removed in `openquake/qa_tests_data/disagg/case_2/expected_output/` because now the line `disagg_outputs = Mag` in `disagg/case_2/job.ini` is honored, i.e. outputs different from `Mag` are not exported anymore.

NB: `disagg_outputs` is not documented in the manual


rcgee:
LGTM

Issue #3211, opened 2017-11-03
ctrl-c does not kill subproc when also dbserver is started by same entrypoint


daniviga:
When `oq webui start` is run and no DbServer is already running, DbServer is started automatically.

In this case, when a `CTRL-C` on the WebUI process is fired, the `oq-job-**` and `oq-workers` processes are left alive and stays in memory until the WebUI is restarted.

When the database server is started a priori everything is working as expected, both with and without `prctl`.

DbServer started manually:
![peek 2017-11-03 09-02](https://user-images.githubusercontent.com/1818657/32364646-5106ba3c-c076-11e7-9403-676e4076f662.gif)

DbServer started by `oq webui start`:
![peek 2017-11-03 09-06](https://user-images.githubusercontent.com/1818657/32364641-4cb14f92-c076-11e7-9303-c64c674b6853.gif)

micheles:
Unfortunately this is true only for small jobs. For large jobs, even if the DbServer is not started by the WebUI, killing the WebUI leaves the jobs running.

micheles:
Jobs started from the WebUI are already killed correctly, at least for Python 3. Job not started by the WebUI are not killed, but I would argue that this is the right behavior. It seems that the change in
https://github.com/gem/oq-engine/pull/3212/ fixed this bug too. Should we close it?

daniviga:
No, unfortunately even using Python 3 I get orphan processes when I `CTRL-C` the WebUI and the DbServer has been started by it.

daniviga:
To solve this issue we need to demonize the dbserver (see https://pypi.python.org/pypi/python-daemon/ as an example)

PR #3346, opened 2018-01-11
Kill all oq processes if the webui is killed

micheles:
Closes https://github.com/gem/oq-engine/issues/3211. Notice that I had to use SIGKILL, SIGTERM is not enough. In our cluster the WebUI is run by the user openquake, so if it is killed only the processes of openquake are killed, which is good.


micheles:
I agree, I do not like the SIGKILL myself. I am closing the PR.

PR #3348, opened 2018-01-12
Added command `oq shell`

micheles:
This is an easy way to drop into an (I)Python console, as suggest by Matteo. Here is an example of usage:

$ oq ~~~python~~~ shell
```python
IPython shell with a global oq object
In [1]: oq.read(-1)
Out[1]: <DataStore 19464, open>
In [2]: oq.  # press TAB here
            get_exposure()        get_site_collection() 
            get_oqparam()         read()               
```

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,5 @@
   [Michele Simionato]
+  * Added a command `oq python` embedding IPython if available
   * Turned the 'realizations' output into a dynamic output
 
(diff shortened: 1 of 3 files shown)


micheles:
I though `shell` was Django specific, but if other projects are using it (like Flask) then it makes sense to use it.

mmpagani:
Very nice! LGTM

PR #3349, opened 2018-01-12
Fixed the HMTK tests for modern numpy

micheles:
Closes https://github.com/gem/oq-engine/issues/2974 for the latest numpy version, 1.14, plus several PEP 8 fixes.
NB: we are NOT upgrading the official numpy version we use and there are no plans to do that without a good reason. It is probably a good idea to upgrade numpy and possibly other dependencies *after* engine 3.0 will be out.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,5 @@
   [Michele Simionato]
+  * Fixed the HMTK tests to work with numpy from 1.11 to 1.14
   * Added a command `oq shell` to open an embedded (I)Python shell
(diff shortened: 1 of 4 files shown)


micheles:
NB: there are several engine tests that break with numpy 1.14 because the string representation of the numbers has changed. It is outside of the scope of this PR to fix them all. It will be done in the future, when we will decide to upgrade the numpy version.

mmpagani:
LGTM. Thanks!

Issue #3347, opened 2018-01-11
Add stop bottons to the WebUI (one per job)


micheles:
To make it easy to kill a job. We will need an API calling `oq abort <job_id>`.
For the moment, we will not implement a multi-kill functionality (it could be implemented with a set of check boxes). The functionality will be used by the QGIS plugin too.

PR #3351, opened 2018-01-12
Added an abort Web API

micheles:
Closes https://github.com/gem/oq-engine/issues/3347.


daniviga:
On Linux it works pretty well. On Windows it does not work at all because setproctitle is unable to set a proper process name (and we should use PIDs). On macOS setproctitile works, but the abort doesn't. I have tu further investigate.

daniviga:
The problem on macOS is that in

```python
358     for p in psutil.process_iter():
359         if p.name() == name:
```

`p.name()` returns `Python`, even if the process has the proper `oq-job-X` name in `ps aux`.

PR #3356, opened 2018-01-15
Specialized pkgs

nastasi-oq:
With this PR we add a couple of optional packages to be able to install properly standalone, master or worker (on a cluster, for example).
Tests are green here:  https://ci.openquake.org/job/zdevel_oq-engine/2639/


debian/control
```diff
@@ -18,3 +18,27 @@ Suggests: python-oq-libs-extra (>= 1.5.0~)
...
+Breaks: python-oq-risklib, python-oq-hazardlib
+Recommends: ${dist:Recommends}
+Suggests: python-oq-libs-extra (>= 1.5.0~)
+Description: computes seismic hazard and physical risk
+ based on the hazard and risk libraries developed by the GEM foundation
+ (http://www.globalquakemodel.org/)
+
+Package: python-oq-engine-worker
+Architecture: all
+Depends: ${shlibs:Depends}, ${misc:Depends}, python-oq-libs (>=1.5.0~), rabbitmq-server (>=2.8.0-2~gem02), supervisor (>=3.0b2-1~), python-oq-engine (= ${binary:Version})
```
daniviga: `python-oq-engine-worker` must not depend on RabbitMQ. I would probably also remove `python-oq-libs` since `master` depends on `engine` which depends already on `libs`. It's ok here to keep `supervisor` (we include an init in this pkg).

debian/control
```diff
@@ -18,3 +18,27 @@ Suggests: python-oq-libs-extra (>= 1.5.0~)
 Description: computes seismic hazard and physical risk
  based on the hazard and risk libraries developed by the GEM foundation
  (http://www.globalquakemodel.org/)
+
+Package: python-oq-engine-master
+Architecture: all
+Depends: ${shlibs:Depends}, ${misc:Depends}, python-oq-libs (>=1.5.0~), rabbitmq-server (>=2.8.0-2~gem02), supervisor (>=3.0b2-1~), python-oq-engine (= ${binary:Version})
```
daniviga: I would probably remove `python-oq-libs` since `master` depends on `engine` which depends already on `libs`. The same is for `supervisor`. Thsi package does not include any service.

debian/control
```diff
@@ -18,3 +18,27 @@ Suggests: python-oq-libs-extra (>= 1.5.0~)
...
+Replaces: python-oq-risklib, python-oq-hazardlib
+# See https://www.debian.org/doc/debian-policy/ch-relationships.html#s-breaks
+Breaks: python-oq-risklib, python-oq-hazardlib
+Recommends: ${dist:Recommends}
+Suggests: python-oq-libs-extra (>= 1.5.0~)
```
daniviga: Should not suggest `python-oq-libs-extra`: in this package there's nothing that relies on it.

debian/control
```diff
@@ -18,3 +18,27 @@ Suggests: python-oq-libs-extra (>= 1.5.0~)
...
+ (http://www.globalquakemodel.org/)
+
+Package: python-oq-engine-worker
+Architecture: all
+Depends: ${shlibs:Depends}, ${misc:Depends}, python-oq-libs (>=1.5.0~), rabbitmq-server (>=2.8.0-2~gem02), supervisor (>=3.0b2-1~), python-oq-engine (= ${binary:Version})
+Replaces: python-oq-risklib, python-oq-hazardlib
+# See https://www.debian.org/doc/debian-policy/ch-relationships.html#s-breaks
+Breaks: python-oq-risklib, python-oq-hazardlib
+Recommends: ${dist:Recommends}
+Suggests: python-oq-libs-extra (>= 1.5.0~)
```
daniviga: Should not suggest `python-oq-libs-extra`: in this package there's nothing that relies on it.

Issue #3190, opened 2017-10-25
Add a csv importer for the exposure


micheles:
The XML exposure could contain a link to a .csv files with the asset information, something like

```xml
<assets> ./assets1.csv ./assets2.csv  </assets>
```

The mandatory fields are `asset_ref, number, taxonomy, lon, lat`.
The optional fields are `area`, `structural-value, structural-deductible, structural-insurance_limit` ... `occupants-day, occupants-night, occupants-transit` and then the tags and `geometry`.

raoanirudh:
It would be useful to support compressed csv files too.

micheles:
Why the field `number` is considered mandatory by IPT but it is not mandatory for the engine?
It is okay to make it mandatory for the engine but then exposure that worked in the past will not work anymore? For instance the exposure in classical_bcr/case_1 has no number. Currently if the number is missing it is considered 1. Currently the number is considered mandatory only in damage calculators, where if missing from the XML and error is raised.

raoanirudh:
@micheles, the relevant section from the manual: 

> The `number` attribute is mandatory for damage calculations. For risk calculations, this attribute must be defined if either the area or any of the costs are provided per structural unit comprising each asset.

micheles:
But in IPT it is always mandatory.

CatalinaYepes:
@micheles we need to update IPT

PR #3365, opened 2018-01-16
Added a .tagvalues attribute to Asset objects

micheles:
And also moved the tagmask concept where it belongs (i.e. only scenario_damage).
This is a preliminary refactoring for https://github.com/gem/oq-engine/issues/3190.


openquake/risklib/asset.py
```diff
@@ -232,26 +253,31 @@ def __repr__(self):
 
 
 class AssetCollection(object):
+    # the information about the assets is store in a numpy array and in a
```
ptormene: `store` -> `stored`

PR #3373, opened 2018-01-19
Add Drouet & Cotton (2015) GMPE including 2017 erratum

stephane-on:
Here is the implementation of the Drouet & Cotton (2015) (https://doi.org/10.1785/0120140240) GMPE including the 2017 erratum.


openquake/hazardlib/gsim/drouet_Alpes_2015.py
```diff
@@ -0,0 +1,681 @@
...
+        <.base.GroundShakingIntensityModel.get_mean_and_stddevs>`
+        for spec of input and result values.
+        """
+        assert all(stddev_type in self.DEFINED_FOR_STANDARD_DEVIATION_TYPES
+                   for stddev_type in stddev_types)
+
+        C = self.COEFFS[imt]
+        mean = self._compute_mean(C, rup.mag, dists.rjb)
+        if isinstance(imt, SA) or isinstance(imt, PGA): # Convert from m/s**2 to g
+            mean=mean-np.log(9.81)
```
g-weatherill: For converting between SI units and fractional g we prefer to use the scipy constants definition of g: `from scipy.constants import g`, to make sure we are not introducing small variations due to differences in rounding g

openquake/hazardlib/gsim/drouet_Alpes_2015.py
```diff
@@ -0,0 +1,681 @@
...
+        """
+        See :meth:`superclass method
+        <.base.GroundShakingIntensityModel.get_mean_and_stddevs>`
+        for spec of input and result values.
+        """
+        assert all(stddev_type in self.DEFINED_FOR_STANDARD_DEVIATION_TYPES
+                   for stddev_type in stddev_types)
+
+        C = self.COEFFS[imt]
+        mean = self._compute_mean(C, rup.mag, dists.rrup)
```
g-weatherill: If you pass the dists object rather than its attribute dists.rjb or dists.rrup then you would only need to modify the `_compute_mean` function to switch from rjb to rrup to rhypo.

openquake/hazardlib/gsim/drouet_Alpes_2015.py
stephane-on: Thanks Graeme, I did the modification.

openquake/hazardlib/gsim/drouet_Alpes_2015.py
stephane-on: I don't really understand how this will help, I will have to pass another argument to the _compute_mean function to condition which distance to use, no?

g-weatherill: Looking again, the idea behind the comment was that if you passed `dists` rather than `dists.rjb` into the call to `_compute_mean` function in the DrouetAlpes2015Rjb, and also passed the `dists` object to the `_compute_term2` function and only inside the `_compute_term2` function indicate:
```python
def _compute_term2(self, C, mag, dists):
    """
    This computes the term f2 equation 8 Drouet & Cotton (2015)
    """
    return (C['c4'] + C['c5'] * mag) * np.log(np.sqrt(dists.rjb **2 + C['c6'] **2)) + C['c7'] * dists.rjb
```
Then in the subclasses you would only need to modify the `_compute_term2` function (and the `REQUIRES_DISTANCES`) in the subclasses and not need to repeat so much of the code of the `get_mean_and_stddevs` method. 
That said however, this is a cosmetic change to reduce the number of lines of code and doesn't really have any impact, so it is not critical.
Note, though, that changing the `REQUIRES_DISTANCES` attribute to reflect the distance metric used in each subclass is critical!

stephane-on: I would still have to modify:
stddevs = self._get_stddevs(C, stddev_types, rup.mag,
                                    dists.rjb.shape[0])
in the get_mean_and_stddevs method. 
The Requires_distance attribute is now defined in each subclass, thanks.

micheles:
I am closing and reopening this PR.

PR #3376, opened 2018-01-22
Added a script to convert from XML to CSV

micheles:
This is useful for https://github.com/gem/oq-engine/issues/3190, to produce the CSV files for the tests.
NB: also the case of retrofitted values must be managed.


utils/expo2csv.py
```diff
@@ -0,0 +1,80 @@
+#!/user/bin/env python
```
ptormene: Prova

PR #3379, opened 2018-01-23
Implements the GMPEs of Bommer et al. (2009) and Afshari & Stewart (2016)

g-weatherill:
Implements two GMPEs for relative significant duration:
1. Bommer et al. (2009)
2. Afshari & Stewart (2016)

Adds `RSD575`, `RSD595` and `RSD2080` to the support IMTs


micheles:
LGTM from the IT point of view. The tests are green on Jenkins: https://ci.openquake.org/job/zdevel_oq-engine/2645/

openquake/hazardlib/gsim/afshari_stewart_2016.py
```diff
@@ -0,0 +1,226 @@
...
+    def get_magnitude_term(self, C, rup):
+        """
+        Returns the magnitude scaling term in equation 3
+        """
+        b0, stress_drop = self._get_sof_terms(C, rup.rake)
+        if rup.mag <= C["m1"]:
+            return b0
+        else:
+            # Calculate moment (equation 5)
+            m_0 = 10.0 ** (1.5 * rup.mag + 16.05)
```
mmpagani: In order to have numbers completely consistent, we should move conversion equations like this one in a utils module.

openquake/hazardlib/gsim/bommer_2009.py
```diff
@@ -0,0 +1,127 @@
...
+import numpy as np
+
+from openquake.hazardlib.gsim.base import CoeffsTable, GMPE
+from openquake.hazardlib import const
+from openquake.hazardlib.imt import RSD595, RSD575
+
+
+class BommerEtAl2009RSD(GMPE):
+    """
+    Implements the GMPE of Bommer et al. (2009) for significant duraction with
```
mmpagani: duraction -> duration

openquake/hazardlib/gsim/afshari_stewart_2016.py
g-weatherill: This makes sense to me. There are about 3 - 4 different utils.py files in different places around the engine (one in risklib, one in hmtk/strain, one in hmtk/seismicity) so looking at consolidation (at least of the more generic formulae such as these) is not a bad idea.

Issue #3388, opened 2018-01-26
Add a command/utility to dump a single calculation


daniviga:
`oq dump` is already able to dump the entire `oqdata`, we should add the ability to dump a single calculation.

Probably, as suggested by @micheles, it's enough just to recreate a record n the destination db having the hdf5 (which contains all the needed info). In this case only the import phase needs to be changed.

In this way it's possible to run a calculation on a foreign machine and then use such calculation with a local engine server instance.

Suggested by @klunk386

PR #3389, opened 2018-01-26
Added the ability to dump a specific calculation

micheles:
Closes https://github.com/gem/oq-engine/issues/3388.


daniviga:
It works, but as it is now it dumps the _full db_ which is not applicable for the case of #3388

Issue #3404, opened 2018-02-01
The site_model association is done when it is not needed


micheles:
Anirudh discovered that it is done twice in the California calculation, see the log for calculation 17320 in cluster2.

raoanirudh:
In the `gmf_ebrisk` calculator, the site model association is done once before reading the exposure, and once again after the assets have been read. However, if the `gmf_ebrisk` calculator is being used, it implies that the ground motion fields have already been calculated and saved in a previous calculation. There is thus no need to do the site model association in the risk job.


```sh
[2018-02-01 11:15:47,956 #245 INFO] Using engine version 2.9.0-gitee42180
[2018-02-01 11:15:48,054 #245 WARNING] GmfEbRiskCalculator is still experimental
[2018-02-01 11:15:50,168 #245 WARNING] The site parameter associated to <Latitude=41.464110, Longitude=-124.034620, Depth=0.0000> came from a distance of 12 km!
[2018-02-01 11:15:50,280 #245 WARNING] The site parameter associated to <Latitude=40.650870, Longitude=-122.947670, Depth=0.0000> came from a distance of 28 km!
[2018-02-01 11:15:50,463 #245 WARNING] The site parameter associated to <Latitude=38.971280, Longitude=-122.432310, Depth=0.0000> came from a distance of 13 km!
[2018-02-01 11:15:50,747 #245 WARNING] The site parameter associated to <Latitude=36.819600, Longitude=-122.051880, Depth=0.0000> came from a distance of 11 km!
[2018-02-01 11:15:50,922 #245 WARNING] The site parameter associated to <Latitude=38.679600, Longitude=-121.902440, Depth=0.0000> came from a distance of 9 km!
[2018-02-01 11:15:51,778 #245 WARNING] The site parameter associated to <Latitude=38.862700, Longitude=-120.979570, Depth=0.0000> came from a distance of 5 km!
[2018-02-01 11:15:52,074 #245 WARNING] The site parameter associated to <Latitude=33.505600, Longitude=-119.117170, Depth=0.0000> came from a distance of 45 km!
[2018-02-01 11:15:53,234 #245 WARNING] The site parameter associated to <Latitude=35.123210, Longitude=-117.202600, Depth=0.0000> came from a distance of 11 km!
[2018-02-01 11:15:53,498 #245 WARNING] The site parameter associated to <Latitude=35.463440, Longitude=-116.614240, Depth=0.0000> came from a distance of 7 km!
[...]

PR #3405, opened 2018-02-02
Forbid the site_model_file when it makes no sense

micheles:
When the hazard is precomputed, it is useless to read the site_model_file and to associate the site parameters to the hazard sites. Closes https://github.com/gem/oq-engine/issues/3404

PS: while at it, I am doing some cleanup of the code.


micheles:
@raoanirudh please try now

PR #3408, opened 2018-02-02
Added an /extract/ API for the hazard outputs

micheles:
and removed the old qgis-hcurves, qgis-hmaps, qgis-uhs API, which despite the name were not compatible with the QGIS plugin.

PS: I am also extending `oq extract` so that it can call the web API too.


openquake/calculators/extract.py
```diff
@@ -153,29 +157,80 @@ def extract_hazard(dstore, what):
...
+    return mesh
+
+
+def hazard_items(dic, mesh, *extras, **kw):
+    """
+    :param dic: dictionary of arrays of the same shape
+    :param mesh: a mesh array with lon, lat fields of the same length
+    :param extras: optional triples (field, dtype, values)
+    :param kw: dictionary of parameters (like investigation_time)
+    :returns: a list of pairs (key, value) suitable for storage in .nzp format
```
ptormene: Typo `nzp` -> `npz`

PR #3410, opened 2018-02-03
Improved the task distribution

micheles:
There are two common problems in classical calculations with not so many sites: there could be not enough tasks (therefore wasting cores) and there can be slow tasks dominating the calculation (this latter problem is ubiquitous). Here is an example for around 1000 sites in Europe using the SHARE model on cluster2:
```
michele@wilson:~/SHAREnew$ oq2 show performance 17327
============================== ======== ========= =======
============================== ======== ========= =======
total classical                13,889   76        134    
get_poes                       9,260    0.0       82,366 
make_contexts                  4,275    0.0       111,780
total runtime                  1,744    283       1      
reading composite source model 256      140       1      
managing sources               107      133       1      
prefiltering                   39       156       1      
================== ==== ====== ===== ===== =========
operation-duration mean stddev min   max   num_tasks
classical          103  163    1.567 1,420 134      
================== ==== ====== ===== ===== =========
classical.received    tot 58.15 MB
```
We have 160 cores, but only 134 tasks are generated, 26 cores are wasted. Also, the slowest task of 1420 seconds is dominating the total runtime of 1,744  seconds. To improve the situation the user has to manually introduce a concurrent_tasks parameter in the job.ini to increase the number of generated tasks.
With this branch the situation is much better:
```
michele@wilson:~/SHAREnew$ oq2 show performance 17326
============================== ======== ========= =======
============================== ======== ========= =======
total classical                14,415   41        183    
get_poes                       9,358    0.0       82,416 
make_contexts                  4,718    0.0       111,834
total runtime                  1,144    370       1      
reading composite source model 253      140       1      
managing sources               227      217       1      
prefiltering                   196      446       1      
================== ==== ====== ===== === =========
operation-duration mean stddev min   max num_tasks
classical          78   113    1.326 664 183      
================== ==== ====== ===== === =========
classical.received    tot 120.99 MB
```
Here the slowest task takes only 664 seconds (less than half the time than before) and more tasks are generated (183 > 134). The relative standard deviation in the task duration distribution goes down from 158% to 145%, i.e. the task distribution improves.
The cost is a longer time in prefiltering and more data transfer, but it is a good tradeoff: notice that the computation time went down from 29 minutes to 19 minutes.

Technically the improvement comes from splitting the area sources before the filtering: this makes the weighing algorithm more accurate. This is consistent with what we did already for the multipoint sources.
Fortunately, splitting the area sources is fast.

PS: a further improvement can be obtained by tweaking the weighting algorithm. The calculation time is less than linear with the number of sites (say that if you have 100 times more sites the calculation is only 10 times slower). What the true relation is difficult to assess but a `sqrt(nsites)` is closer to the reality than `nsites` so heuristically I am using that. Here are the figures with the new weights:
```
michele@wilson:~/yufang$ oq2 show performance 17356
============================== ======== ========= =======
============================== ======== ========= =======
total classical                16,714   29        205    
get_poes                       10,872   0.0       82,416 
make_contexts                  5,463    0.0       111,834
total runtime                  805      371       1      
reading composite source model 254      140       1      
managing sources               228      230       1      
prefiltering                   197      446       1      
================== ==== ====== ===== === =========
operation-duration mean stddev min   max num_tasks
classical          81   76     1.203 326 205      
================== ==== ====== ===== === =========
classical.received    tot 123.44 MB

The total runtime goes down to 13 minutes and the stddev to 94%.
```


mmpagani:
LGTM

openquake/hazardlib/calc/hazard_curve.py
```diff
@@ -134,7 +135,7 @@ def pmap_from_trt(sources, src_filter, gsims, param, monitor=Monitor()):
         poe_mon = monitor('get_poes', measuremem=False)
         pmap = AccumDict({grp_id: ProbabilityMap(len(imtls.array), len(gsims))
                           for grp_id in grp_ids})
-        pmap.calc_times = []  # pairs (src_id, delta_t)
+        pmap.calc_times = AccumDict(accum=numpy.zeros(4))
```
ptormene: Please add a comment about the number 4

micheles: Done

openquake/hazardlib/source/base.py
```diff
@@ -56,7 +57,7 @@ def weight(self):
         if not self.num_ruptures:
             self.num_ruptures = self.count_ruptures()
         return (self.num_ruptures * self.RUPTURE_WEIGHT *
-                self.nsites * self.ngsims)
+                math.sqrt(self.nsites) * self.ngsims)
```
ptormene: Please add a comment about the choice of using math.sqrt

PR #3412, opened 2018-02-05
Upgrade max dependency limits in setup.py

daniviga:
For just a bunch of 'safe' libraries (not numpy, scipy...).

I added also an option to Travis to run tests against packages from upstream instead of our wheels, so in such case latest possible versions will be used.


micheles:
I like the [skip_wheels] option

PR #3413, opened 2018-02-05
Added parameter `source_model_file` to the job.ini

micheles:
Normally the source model files are list inside the logic tree file. However, if the logic tree is trivial, it is more convenient to skip the logic tree file and link  directly to the source model file.


mmpagani:
@micheles I don't know exactly from where this request is coming from but I don't consider it a good idea. I prefer to have a hazard model with a fixed basic structure.

micheles:
It is convenient for me for debugging purposes, I would keep this feature secret.

PR #3414, opened 2018-02-06
Supported retrofitted field in the CSV exposure

micheles:
As of now, it is possible to use CSV exposures but not all fields are supported. In particular, before this PR, the `retrofitted` field was not supported. Also, before this PR there was no check that `retrofittedType` and `retrofittedUnit` were consistent with the ones of the structural type, thus opening the possibility for errors. 
In this PR I am fixing/extending the utility that converts old exposures from XML to CSV and I am changing the classical_bcr calculator so that it only works for structural loss type, the one that has a retrofitting cost.

The field deductible and insurance_limit are still not supported. The CSV format is similar to the one generated by IPT but not yet the same. These enhancements are left to future pull requests.

PS: this PR contains also a minor refactoring which is a huge performance improvement for calculations with a large number of realizations. By vectoring the returned dictionary on the realizations even really large calculations (i.e. Canada with 2187 realizations, 118,565 assets) become manageable.


openquake/commonlib/readinput.py
```diff
@@ -728,9 +748,11 @@ def read(cls, fname, calculation_mode='', insured_losses=False,
         param['ignore_missing_costs'] = set(ignore_missing_costs)
         exposure, assets = _get_exposure(
             param['fname'], param['all_cost_types'])
-        nodes = assets if assets else exposure._read_csv(
-            ~assets, os.path.dirname(param['fname']))
-        exposure._populate_from(nodes, param)
+        _nodes = assets if assets else exposure._read_csv(
+            assets.text, os.path.dirname(param['fname']))
+        if nodes:  # this is useful for Paul Henshaw
+            return nodes
```
pslh: I think you mean return _nodes here.
Also I suggest changing the comment to "This is useful for the GED4ALL import script"

micheles: Done. BTW I have changed the signature of the call, you should use `Exposure.read(fname, asset_nodes=True)` which I believe is clearer.

PR #3423, opened 2018-02-08
Disaggregation by source

micheles:
Computing and displaying the results of a disaggregation-by-source calculation is in general very difficult.
However if we restrict ourselves to the case of a trivial logic tree (i.e. 1 realization) then the problem becomes manageable. In particular, for each site, poe_disagg and imt we can export a csv with header
(poe, source_id) returning the contributions by each source, in order from the greatest contribution to the smallest one. Sources giving zero contribution can be skipped.


rcgee:
I like that you added the source names to the output in addition to the ids. If there is no source name, the word "point" is written. Perhaps something like "None" would be better?

micheles:
You have an example with sources named "point". For what concerns the naming convention I propose a Src suffix, i.e. "poe-0.1-PGA-44.4991-40.1792_Src_17821.csv"

rcgee:
I agree with using the suffix "Src"

rcgee:
LGTM

mmpagani:
better all lowercase to keep things simple

micheles:
But the other outputs have names like "Lon", "Lat", "Mag" and we are using mixed case anyway because we have things like "PGA" and "SA".

PR #3426, opened 2018-02-09
Allow sysadmins to configure a custom path for TMPDIR

### LINKED ISSUE #3425: Add a settings in openquake.cfg for temp data path
Currently export operations are performed via `/tmp`: on cloud instances space available to `/tmp` is fairly limited, and a second volume should be used. This can be done via a `TEMPDIR` override in the `.bashrc` (https://docs.python.org/3.5/library/tempfile.html#tempfile.gettempdir), but it must be done for every user and is not understood by the WebUI (unless the `init` file is changed, which could cause trouble during upgrades).

My proposal is to add a new key to the `openquake.cfg` file that overrides, internally, `TEMDIR`. This will be used both by the cli and the api server.

This actually prevents the Engine to make serious calculations on Azure.

daniviga:
Closes #3425

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -5,6 +5,7 @@
     and by improving the weighting algorithm
 
   [Daniele Viganò]
+  * TMPDIR can be customized via the `openquake.cfg`
   * Updated dependencies compatibility in setup.py
 
   [Michele Simionato]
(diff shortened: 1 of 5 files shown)


micheles:
Please add a line in the changelog. Other than that, LGTM

PR #3422, opened 2018-02-08
Manual 2.9: Document the new output display names

raoanirudh:
Resolves item 3 in both the hazard and risk sections of #3216

Issue #2887, opened 2017-06-14
Automatic manual version


daniviga:
The OpenQuake version in the manual (cover), the citation and DOI (page 3) should be automatically updated by `build.sh` (sed?)

raoanirudh:
And if possible, also the total number of pages on page 3

daniviga:
Closed via #3427 and #3476 

DOI is still manually populated, but this requires a separated issue.

daniviga:
DOI has been fixed here https://github.com/gem/oq-engine/pull/3588

PR #3427, opened 2018-02-09
Automatically update version number on manual cover page

raoanirudh:
Addresses the easy part of #2887

The citation and DOI require far more regex wizardry. eg: https://github.com/gem/oq-engine/pull/3422/files#diff-5bc4e11693b47d7e445cc1f5754a4074

### DIFF
--- a/doc/manual/build.sh
+++ b/doc/manual/build.sh
@@ -5,6 +5,9 @@ if [ $GEM_SET_DEBUG ]; then
 fi
 set -e
 
+version="$(cat openquake/baselib/__init__.py | sed -n "s/^__version__[  ]*=[    ]*['\"]\([^'\"]\+\)['\"].*/\1/gp")"
+sed -i "s/Version [0-9]\.[0-9]\.[0-9]/Version $version/" oq_manual_cover.svg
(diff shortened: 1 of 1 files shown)


doc/manual/build.sh
```diff
@@ -5,6 +5,9 @@ if [ $GEM_SET_DEBUG ]; then
 fi
 set -e
 
+version="$(cat openquake/baselib/__init__.py | sed -n "s/^__version__[  ]*=[    ]*['\"]\([^'\"]\+\)['\"].*/\1/gp")"
```
daniviga: This would make this script work only if called from the project root, otherwise this path would not be found.

I would: save actual path of the script and call `$CURPATH/../../openquake/baselib/__init__.py` this will make the script 'portable'.

daniviga: You can use something like 
`$CURPATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"`

to get the dir where the script is.

raoanirudh: Thanks @daniviga, done!

Issue #3046, opened 2017-09-24
Remove the dependency from concurrent.futures


micheles:
A dependency on `multiprocessing.Pool.imap_unordered` is enough for the kind of concurrency used in the engine. multiprocessing is simpler and has features missing in concurrent.futures (like the ability to run initialization code at pool startup). Also, it will become easier to kill the pool properly.

Issue #3430, opened 2018-02-12
Add distribution mode celery_zmq


micheles:
The idea is to send the tasks via celery and to return the results via zmq. Notice that in this approach the revoke functionality is already implemented so we can close https://github.com/gem/oq-engine/pull/3199.

PR #3431, opened 2018-02-12
Substantial simplification of the parallelization framework

micheles:
Closes https://github.com/gem/oq-engine/issues/3046. Code in the state experimental/proof-of-concept has been removed: in particular the support to ipython and the support to SGE. As it is now, they are not used and still a significant maintenance cost. They can easily be restored if needed (i.e. if a sponsor is willing to pay for that).
There is also a lot of work on the zmq code, in preparation for https://github.com/gem/oq-engine/issues/3430.

The tests with python 2.7 are running here: https://ci.openquake.org/job/zdevel_oq-engine/2664


daniviga:
The `ctrl-c` isn't working properly (again). A single `ctrl-c` does not stop the calculation properly.

The process is locked at
screenshot from 2018-02-12 11-08-46

this at least on mac. I'm going to test it on linux an windows.

PR #3436, opened 2018-02-12
20180213 gridded ruptures

mmpagani:
This is a tiny change that fixes the disaggregation in case of gridded ruptures. I think it would be useful to add a small test case of disaggregation with non-parametric gridded ruptures but perhaps this can be addressed in a second PR. Having this fix in master will help in performing more tests on the current developed models.


openquake/hazardlib/tests/geo/surface/gridded_test.py
```diff
@@ -34,12 +34,17 @@ def setUp(self):
...
+        expected = np.array([111.232737])
         np.testing.assert_allclose(dists, expected, rtol=1e-5, atol=0)
 
     def test_get_closest_points(self):
-        self.assertRaises(NotImplementedError, self.surf.get_closest_points,
-                          self.mesh)
+        res = self.surf.get_closest_points(self.mesh)
+        print(self.surf.mesh.lons)
+        print(self.surf.mesh.lats)
+        print(self.surf.mesh.depths)
```
micheles: Please remove the debugging prints

PR #3440, opened 2018-02-13
Fixed HDF5 fork bug (again)

micheles:
The situation is bad. HDF5 files can be read reliably only if they are opened after the fork, not before. However, the engine must read the datastore before the fork, so we must

1) make sure that the datastore have been closed before the fork
2) make sure that nobody have opened the datastore again after we closed it!

I cannot think of a robust solution, so at each refactoring the bug can resurface and the tests cannot cover it, since it is random. Only the tests macOS can give some hint sometimes.

Depression :-(

NB: my ideal solution for this issue would be to remove the fork, i.e. to abandon multiprocessing and to do the concurrency with zmq. That would have the additional benefit of a much better performance on Windows.

### DIFF
--- a/openquake/calculators/base.py
+++ b/openquake/calculators/base.py
@@ -700,6 +700,9 @@ class RiskCalculator(HazardCalculator):
                     # read the hazard data in the controller node
                     logging.info('Reading hazard')
                     getter.init()
+                else:
+                    # the datastore must be closed to avoid the HDF5 fork bug
+                    assert dstore.hdf5 == (), '%s is not closed!' % dstore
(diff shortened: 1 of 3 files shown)


daniviga:
Could this http://docs.h5py.org/en/latest/swmr.html help?

micheles:
No, I tried to use this feature several times in the past and always failed. Perhaps now the situation is better but from past experience I am sure that getting this to work is a significant effort. Also the format is not compatible with the past, and one year ago it was not even readable with the hdfviewer.
It could still be interesting for other reasons, but it is a big project. In that amount of time I could get zmq working properly which IMO is a more useful thing to do.

PR #3441, opened 2018-02-13
Fix hazard outputs in the case of no outputs

micheles:
The issue, discovered by Paolo, is that the WebUI incorrectly exposes hazard curves/maps/uhs even in the case when there are no outputs, so that the QGIS plugin gets empty outputs and fails.
To test it:

1. take a calculation like LogicTreeCase1ClassicalPSHA and change in the job.ini
```
mean_hazard_curves = false
quantile_hazard_curves = 
```
2. run the calculation with QGIS.

There will be no outputs shown.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,6 @@
   [Michele Simionato]
+  * Fixed a bug in the hazard outputs: they were displayed in the WebUI even
+    missing
   * Implemented splitting of nonparametric sources
 
   [Marco Pagani]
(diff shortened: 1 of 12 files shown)


ptormene:
It might be a little misleading for the user, if `hazard_maps = true` and/or `uniform_hazard_spectra = true` and also those outputs are not generated when hazard curves are not generated. I guess at least a warning would be useful.

micheles:
Now an error is raised if hazard_maps = true and/or uniform_hazard_spectra = true but mean_hazard_curves=false, before starting the calculation.

PR #3452, opened 2018-02-16
Ubuntu Python 3.5

nastasi-oq:
Move Ubuntu Package to python3.
Pre-refresh tests are green here:
  * https://ci.openquake.org/job/zdevel_oq-engine/2700/ and
  * https://ci.openquake.org/job/zdevel_oq-engine/2701/

Current commit tests are running here:
  * https://ci.openquake.org/job/zdevel_oq-engine/2703/


doc/upgrading/ubuntu.md
```diff
@@ -21,7 +21,7 @@ sudo apt-get update
 
 If you are upgrading from OpenQuake Engine release 1.x, before you can process you have to run this command:
 ```bash
-sudo apt-get remove --purge python-oq.*
```
daniviga: python3-oq.* doesn't have to be here:

>If you are upgrading from OpenQuake Engine release 1.x, before you can process you have to run this command:

We should also change process to proceed

doc/upgrading/ubuntu.md
```diff
@@ -21,7 +21,7 @@ sudo apt-get update
...
+sudo apt-get remove --purge python-oq.* python3-oq.*
```
daniviga: python3-oq.* doesn't have to be here:

>If you are upgrading from OpenQuake Engine release 1.x, before you can process you have to run this command:

We should also change process to proceed

nastasi-oq: You are right ! With transitional packages it is more simple move from 2.8 to 2.9.

Issue #3216, opened 2017-11-06
Updates required for the user manual for v2.9


micheles:
### Hazard
- [x] Document the new display names for hazard outputs introduced in #3357 and subsequent commits. See the dict `DISPLAY_NAME` in [openquake/server/db/actions.py](https://github.com/gem/oq-engine/blob/master/openquake/server/db/actions.py) for the complete list.

### Risk
- [x] Add example for scenario_damage/scenario_risk starting from csv gmf files
- [x] Document the gmf_ebrisk calculator
- [x] Document the new display names for risk outputs introduced in #3357 and subsequent commits. See the dict `DISPLAY_NAME` in [openquake/server/db/actions.py](https://github.com/gem/oq-engine/blob/master/openquake/server/db/actions.py) for the complete list.

### LaTeX
- [x] Fix issue with the [scherbaum2011 citation](https://github.com/gem/oq-engine/blob/manual-2.8/doc/manual/oqum/glossary.tex#L42). This citation is in the glossary, so it doesn't get picked up by bibtex until the glossary is compiled. The build.sh script will have to be modified to resolve this issue.

raoanirudh:
Other persistent warnings:

```sh
Package ctable Warning:  Transparency disabled: incompatible with tikz package
LaTeX Warning: Command \@parboxrestore  has changed.
LaTeX Font Warning: Font shape `T1/mdbch/m/fscn' undefined
Package biblatex Warning: Package option 'sortcites' enabled.
Package caption Warning: \captionsetup{type*=...} or \captionof
Package caption Warning: The option `hypcap=true' will be ignored
LaTeX Warning: Float too large for page by 2.78722pt on input line 45.
LaTeX Warning: Float too large for page by 675.28645pt on input line 141.
LaTeX Warning: Float too large for page by 125.28706pt on input line 182.
```

PR #3459, opened 2018-02-20
Manual: Fix a LaTeX issue with a citation [skip CI]

raoanirudh:
By removing it.

Resolves the last item in #3216


micheles:
I like this smart fixes! LGTM

PR #3462, opened 2018-02-20
Adds Rjb 0.1 km clip to Derras 2014

g-weatherill:
The Derras et al. 2014 GMPE gives erroneous results for the case when Rjb = 0. Due to the normalisation terms the Rjb is now clipped at Rjb = 0.1.


micheles:
Please update the changelog. Other than that, LGTM.

PR #3463, opened 2018-02-21
Splitting all the sources upfront

micheles:
Splitting the MultiPointSources upfront improved the task distribution. Splitting the AreaSources upfront improved it even more. Here I am splitting all splittable source typologies upfront and the task distribution is a bit better. The cost is a longer startup time of the tasks, but it is minor. All in all, the improvement is not at all spectacular, but it is a simplification of the current approach, 40 lines of code were removed. 
Since the distribution is so good, I am increasing the default size of the tiles from 20,000 to 30,000 sites, to produce a bit less tasks.


micheles:
Superseded by https://github.com/gem/oq-engine/pull/3491

Issue #3466, opened 2018-02-22
 Make it possible to run calculations from a pre-imported exposure


micheles:
This is useful in a lot of situations, and also for the Aristotle project. We need a method to extract
a reduced AssetCollection from the full exposure, given a region_constraint.

mmpagani:
rtree is your friend :)

micheles:
Actually I am convinced that `shapely.geometry.Point.within(region)` (the one we are already using) is fast enough.

PR #3467, opened 2018-02-22
Moved the asset_refs inside the AssetCollection

micheles:
This is a minor refactoring, but it is helpful as a step towards https://github.com/gem/oq-engine/issues/3466.


openquake/commonlib/util.py
```diff
@@ -125,11 +125,11 @@ def get_assets(dstore):
     :returns: an ordered array of records (asset_ref, taxonomy, lon, lat)
     """
     assetcol = dstore['assetcol']
-    asset_refs = dstore['asset_refs'].value
+    aref = assetcol.asset_refs
```
ptormene: Perhaps `arefs` (plural) would be more intuitive

ptormene:
If it's not too internal stuff, you might also update che changelog

PR #3468, opened 2018-02-22
Manual: Add downloadable examples for sites, gmfs csvs for scenarios

raoanirudh:


micheles:
Now it builds correctly :-)

PR #3473, opened 2018-02-23
Added a distribution mode celery_zmq

micheles:
In this distribution mode the results are returned via zmq, thus reducing significantly the load on celery.
Here are the demos: https://travis-ci.org/gem/oq-engine-celery/builds/345543317.
The new approach is a big success, and I was able to compute on cluster2 the ruptures for Guatemala with 250,000 years (https://oq2.wilson.openquake.org/engine/17614/outputs), something impossible to do with pure celery because rabbitmq was dying (WorkerLostError). Also, the South Africa calculation for the first time ran to completion. The SHARE computation is slightly faster (8532 s for job=17615 vs 8963 s for job=17645, i.e. 7 minutes less on a data transfer of 274 GB).  The revoke works well.
I tried to stress as much as possible the cluster by running multiple big calculations simultaneously and it is rock solid.
If there are no ports available in the allowed range it raises a ZMQBindError. The only issue is that
if the OEMKiller kills the master, the workers keep running. I guess the reason is that the OEMKiller sends a SIGKILL which is not trappable, so the revoke message is not sent to celery.

The Tanzania calculation that was mysteriously hanging is still hanging :-(


openquake/baselib/parallel.py
```diff
@@ -554,13 +622,9 @@ def __iter__(self):
...
         self.progress('Executing "%s" in process', self.name)
-        allargs = []
-        for task_no, args in enumerate(self.task_args):
-            if isinstance(args[-1], Monitor):  # not the case for wakeup
-                args[-1].task_no = task_no
-            allargs.append(args)
+        allargs = list(self._genargs(pickle=False))
         yield len(allargs)
-        for args in allargs:
+        for task_no, args in enumerate(allargs):
```
ptormene: Please remove the unused task_no

PR #3480, opened 2018-02-27
A Simple writer for HMTK catalogue

klunk386:
Included a method to export data directly from the Catalogue() object in standard HMTK format.


openquake/hmtk/tests/seismicity/catalogue_test.py
```diff
@@ -82,6 +84,15 @@ def setUp(self):
             [1960, 5.0],
         ])
 
+    def write_catalogue(self):
+        # Test the export of a catalogue in csv format
+        inp_file = 'data/test_write_input.csv'
+        out_file = 'data/test_write_output.csv'
```
micheles: There is a trick to use. As it is, the test will break if you are not in the right directory. What you can do is to define

CURRENT_DIR = os.path.dirname(__file__)
and then 

￼        inp_file = os.path.join(CURRENT_DIR, 'data/test_write_input.csv')
￼        out_file = os.path.join(CURRENT_DIR, 'data/test_write_output.csv')

openquake/hmtk/seismicity/catalogue.py
```diff
@@ -107,8 +115,38 @@ def get_number_events(self):
...
+    def write_catalogue(self, output_file, key_list=SORTED_ATTRIBUTE_LIST):
+        """
+        Writes the catalogue to file using HTMK format (CSV).
+
+        :param output_file:
+            Name of the output file
+        :param key_list:
+            Optional list of attribute keys to be exported
+        """
+
```
micheles: It is better to write this as a with statement, `with open(output_file, 'wt') as outfile`.

openquake/hmtk/seismicity/catalogue.py
```diff
@@ -107,8 +115,38 @@ def get_number_events(self):
...
+        outfile = open(output_file, 'wt')
+        writer = csv.DictWriter(outfile, fieldnames=key_list)
+
+        writer.writeheader()
+        for i in range(0, self.get_number_events()):
```
micheles: You can just write `range(self.get_number_events())`

openquake/hmtk/tests/seismicity/catalogue_test.py
```diff
@@ -82,6 +87,15 @@ def setUp(self):
             [1960, 5.0],
         ])
 
+    def write_catalogue(self):
+        # Test the export of a catalogue in csv format
+        inp_file = os.path.join(CURRENT_DIR, 'data/test_write_input.csv')
+        out_file = os.path.join(CURRENT_DIR, 'data/test_write_output.csv')
```
micheles: This is not point in committing the output file, since it is generated. Also, in the test the output file should be written on a temporary location (write permissions on the source directory cannot be assumed) and removed at the end of the test, after the comparison.

openquake/hmtk/tests/seismicity/catalogue_test.py
```diff
@@ -82,6 +88,16 @@ def setUp(self):
             [1960, 5.0],
         ])
 
+    def write_catalogue(self):
+        # Test the export of a catalogue in csv format
+        fi = os.path.join(CURRENT_DIR, 'data/simple.csv')
+        parser = CsvCatalogueParser(fi)
+        cat = parser.read_file()
+        fo = tempfile.NamedTemporaryFile()
+        cat.write_catalogue(fo.name)
```
micheles: Again, better to use the with statement for the TemporaryFile

openquake/hmtk/seismicity/catalogue.py
```diff
@@ -107,8 +115,37 @@ def get_number_events(self):
...
+        Writes the catalogue to file using HTMK format (CSV).
+
+        :param output_file:
+            Name of the output file
+        :param key_list:
+            Optional list of attribute keys to be exported
+        """
+
+        with open(output_file, 'wt') as of:
+            writer = csv.DictWriter(of, fieldnames=key_list)
```
micheles: BTW, since 'rt' and 'wt' are the same as 'r' and 'w', nobody uses the long form.

openquake/hmtk/tests/seismicity/catalogue_test.py
```diff
@@ -82,6 +88,16 @@ def setUp(self):
...
 
+    def write_catalogue(self):
+        # Test the export of a catalogue in csv format
+        fi = os.path.join(CURRENT_DIR, 'data/simple.csv')
+        parser = CsvCatalogueParser(fi)
+        cat = parser.read_file()
+        with tempfile.NamedTemporaryFile() as fo:
+            cat.write_catalogue(fo.name)
+            self.assertTrue(filecmp.cmp(fi, fo.name))
+            fo.close()
```
micheles: You do not need the close(). The with will take care of it.

openquake/hmtk/seismicity/catalogue.py
```diff
@@ -107,8 +115,37 @@ def get_number_events(self):
...
+                            else:
+                                data = int(data)
+                        if key in self.FLOAT_ATTRIBUTE_LIST:
+                            if np.isnan(data):
+                                data = ''
+                            else:
+                                data = float(data)
+                    row_dict[key] = data
+                writer.writerow(row_dict)
+            of.close()
```
micheles: No need for the close.

micheles:
Please refresh the branch and add a line in the changelog

micheles:
Congratulations @klunk386 !

PR #3488, opened 2018-03-01
Bug fix: we were reading the source model twice in disaggregation

micheles:
I discovered this by looking at the logs of Julio's computations, that were taking a long time parsing the source models. @mmpagani, we need to merge the pull requests on disaggregation, that have been open for two months. If you have no time to tests the new features, let's not document them and keep them in experimental status, but having pull requests open two months is no good, then errors like this one that I thought I had fixed a long time ago will stay in master and will affect the official release (engine 2.9 has this bug).

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,5 @@
   [Michele Simionato]
+  * Fixed a small bug: we were reading the source model twice in disaggregation
   * Added a check to discard results coming from the wrong calculation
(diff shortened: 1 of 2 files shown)


mmpagani:
@micheles I agree to merge the PR left open and to leave the new features in an experimental status. Perhaps we can ask to our users to test it and provide feedback.

PR #3490, opened 2018-03-01
Renumber source IDs

micheles:
This script allows to renumber all the sources in a composite source model. It is useful in models (i.e. CCARA) containing duplicated source IDs. In another PR I am adding a warning on duplicated source IDs, so that it is possible to recognize this situation.

### DIFF
--- a/openquake/commands/renumber_sources.py
+++ b/openquake/commands/renumber_sources.py
@@ -15,7 +15,6 @@
 #
 # You should have received a copy of the GNU Affero General Public License
 # along with OpenQuake. If not, see <http://www.gnu.org/licenses/>.
-import shutil
 import logging
 from openquake.baselib import sap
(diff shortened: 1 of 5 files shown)


openquake/commands/renumber_sources.py
```diff
@@ -0,0 +1,62 @@
...
+                src['id'] = str(number)
+                number += 1
+        with open(path, 'wb') as f:
+            nrml.write(root, f)
+
+
+@sap.Script
+def renumber_sources(smlt_file):
+    """
+    Renumber the sources belowing to the same source model, even if split
```
ptormene: I guess you meant "belonging"

julgp:
ok, with this PR, but the most efficient solution is to modify in the oq-mbt the way we are creating the Ids. Yesterday I did, today I'll do the PR

PR #3491, opened 2018-03-02
Splitting the sources upfront

micheles:
Splitting the MultiPointSources upfront improved the task distribution. Splitting the AreaSources upfront improved it even more. Here I am splitting all splittable source typologies upfront and the task distribution is a bit better. The cost is a longer startup time of the tasks, but it is minor. All in all, the improvement is not at all spectacular, but it is a simplification of the current approach.
Since the distribution is so good, I am increasing the default size of the tiles from 20,000 to 30,000 sites, to produce a bit less tasks.


openquake/commonlib/oqvalidation.py
```diff
@@ -53,7 +53,7 @@ class OqParam(valid.ParamSet):
     coordinate_bin_width = valid.Param(valid.positivefloat)
     compare_with_classical = valid.Param(valid.boolean, False)
     concurrent_tasks = valid.Param(
-        valid.positiveint, multiprocessing.cpu_count())
+        valid.positiveint, multiprocessing.cpu_count() * 3)
```
ptormene: Please add a comment

openquake/commonlib/oqvalidation.py
```diff
@@ -128,7 +128,7 @@ class OqParam(valid.ParamSet):
     max_site_model_distance = valid.Param(valid.positivefloat, 5)  # by Graeme
     sites = valid.Param(valid.NoneOr(valid.coordinates), None)
     sites_disagg = valid.Param(valid.NoneOr(valid.coordinates), [])
-    sites_per_tile = valid.Param(valid.positiveint, 20000)
+    sites_per_tile = valid.Param(valid.positiveint, 30000)
```
ptormene: Please add a comment

openquake/commonlib/source.py
```diff
@@ -845,7 +884,7 @@ def init_serials(self):
         start = 0
         for sg in self.src_groups:
             for src in sg:
-                nr = src.num_ruptures
+                nr = src.count_ruptures()
```
ptormene: Please check if you can reuse `num_ruptures` instead

openquake/commonlib/oqvalidation.py
micheles: Done

openquake/commonlib/oqvalidation.py
micheles: Done

openquake/commonlib/source.py
micheles: Restored `num_ruptures`

openquake/commonlib/source.py
```diff
@@ -48,6 +48,33 @@
                       ('gsims', 'S100'), ('weight', F32)])
 
 
+def split_sources(srcs):
+    """
+    :param srcs: sources
+    :param mon: a Monitor instance
+    :returns: a list of split sources
```
ptormene: Please add that it also returns the split_time

PR #3495, opened 2018-03-03
Fixed calculation of statistics in classical_risk

micheles:
This is the bug discovered in the calculation for the Cariboo region in Western Canada. I will add a test sensitive to the bug later on.


micheles:
In the same file, few lines above, there are the definitions
```python
        self.R = len(rlzs)
        self.A = len(self.assetcol)
        self.L = len(self.riskmodel.loss_types)
        self.I = oq.insured_losses + 1
        self.S = len(oq.risk_stats())
```
which I think are clearer than any comment I could write.

Issue #3500, opened 2018-03-06
Fix the test scenario_damage/case_6


micheles:
It is all wrong. Every asset goes in the no_damage state and due to numerical errors all other states are populate very very small numbers, possibly negative. Ask Anirudh.

PR #3504, opened 2018-03-06
Fixes negative damages in scenario_damage/test_case_6

micheles:
Closes https://github.com/gem/oq-engine/issues/3500. Removed also extra-small damages, since they were numerical artifacts, as suggested by @raoanirudh.


openquake/risklib/riskmodels.py
```diff
@@ -500,6 +500,7 @@ def __call__(self, loss_type, assets, gmvs_eids, _eps=None):
         """
         ffs = self.risk_functions[loss_type]
         damages = scientific.scenario_damage(ffs, gmvs_eids[0])  # shape (D, E)
+        damages[damages < 0] = 0  # sanity check
```
raoanirudh: Perhaps we could extend this to  ```damages[damages < tol] = 0``` where, say ```tol = 1E-3 * number```

The other really small numbers seen here, albeit positive, are also really artefacts. eg:
```1.952397E-22, 5.210657E-22, 4.900346E-24, 1.311654E-23, 3.506161E-40, 1.041459E-39``` for the last asset in dmg_by_asset.csv

raoanirudh: Another option could be to print the numbers after rounding them to three decimal places, which will result in all of these very small numbers being rounded to 0.000.

PR #3505, opened 2018-03-06
Made it possible to use a pre-imported risk model

micheles:
This is needed for the Aristotle project.


openquake/commonlib/oqvalidation.py
```diff
@@ -618,10 +618,18 @@ def is_valid_inputs(self):
         Invalid calculation_mode="{calculation_mode}" or missing
         fragility_file/vulnerability_file in the .ini file.
         """
+        if self.hazard_calculation_id:
+            parent = list(datastore.read(self.hazard_calculation_id))
```
ptormene: Parent can be of different types. I suggest to use only one type for it

ptormene: And you might change `parent` into `parent_datasets`, so it becomes clear that it can contain multiple elements

micheles: Done

openquake/risklib/scientific.py
```diff
@@ -563,6 +563,7 @@ class FragilityFunctionList(list):
     A list of fragility functions with common attributes; there is a
     function for each limit state.
     """
+    # NB: the list if populated after instantiation by .append calls
```
ptormene: typo `if` ==> `is`

Issue #3502, opened 2018-03-06
Make it possible to import a calculation from one database to another


micheles:
For the moment, import the calculation if the calculation ID is free, otherwise raise an error.
We can also extend the command to import a bunch of calculation IDs. The command should download the datastore from the WebUI.

PR #3516, opened 2018-03-08
Made it possible to import remote calculations

micheles:
Closes https://github.com/gem/oq-engine/issues/3502.
This only works for calculations without parent, supporting also a parent is left for future work.
The syntax is

```bash
$ oq importcalc https://oq2.wilson.openquake.org/v1/calc/XXX user password
```


openquake/commands/importcalc.py
```diff
@@ -0,0 +1,74 @@
...
+            f.write(chunk)
+    logs.dbcmd('import_job', calc_id, json['calculation_mode'],
+               json['description'], json['owner'], json['status'], datadir)
+    with datastore.read(calc_id) as dstore:
+        engine.expose_outputs(dstore)
+    logging.info('Imported calculation %d successfully', calc_id)
+
+importcalc.arg('calc_url', 'calculation URL')
+importcalc.arg('username', 'user name')
+importcalc.arg('password', 'user password')
```
daniviga: I would like to have the possibility to pass user and password via env variables. This is very useful for example when a script that uses this function is committed in the repo but we cannot disclose username and password, so them are stored in the runner configuration (like Jenkins or Travis). We already had a similar issue with @ptormene on QGIS side.

micheles:
In the description of this PR I wrote:
```This only works for calculations without parent, supporting also a parent is left for future work.```
The calculation you tried has a parent. I meant to give a better error message anyway.
 https://oq2.wilson.openquake.org/v1/calc/17583 which has no parent works.

micheles:
I have changed the WebAPI to return information about the parent calculation ID. I have updated cluster2 with this branch, so you can download from there during your tests.

Issue #3534, opened 2018-03-14
griddedRuptures are not exportable


micheles:
Here is the error:
```python
ebr.export(mesh)
  File "/home/michele/oqcode/oq-engine/openquake/hazardlib/source/rupture.py", line 629, in export
    rupture.surface, iffs, ims)
  File "/home/michele/oqcode/oq-engine/openquake/hazardlib/source/rupture.py", line 536, in get_geom
    for i, corner in enumerate((surface.top_left,
AttributeError: 'GriddedSurface' object has no attribute 'top_left'
```

Moreover `GriddedSurface.get_strike` and `GriddedSurface.get_dip` and `GriddedSurface.get_surface_boundaries` are not implemented.

PR #3539, opened 2018-03-16
Gridded rupture are not exportable

micheles:
Closes https://github.com/gem/oq-engine/issues/3534.


micheles:
This is ready for review now.

Issue #3544, opened 2018-03-20
Port the GALE code inside the engine


micheles:
This is going to be long and difficult. We need

[x] Add an utility to download and parse USGS shakemaps (both grid.xml and uncertainty.xml)
[x] Add an utility to fillter a given shakemap against a pre-imported exposure
[ ] Add an utility to convert a shakemap into a GMF array in the engine format
[ ] Fix the bugs in the current GALE code
[ ] Perform a risk calculation with the given GMF array and pre-imported risk functions

PR #3545, opened 2018-03-20
Shakemap to GMFs

micheles:
Part of https://github.com/gem/oq-engine/issues/3544.


openquake/hazardlib/shakemap.py
```diff
@@ -0,0 +1,237 @@
...
+
+def spatial_length_scale(imt, vs30clustered):
+    """
+    :param imt: intensity measure type
+    :param vs30clustered: boolean
+    :returns: the `b` parameter in the correlation matrix
+    """
+    T = imt.period
+    if T < 1:
+        b = 40.7 - 15.0 * T if vs30clustered else 8.5 + 17.2 * T
```
ptormene: Please add a reference to a scientific paper that explains those numbers

micheles: Vitor says that the article is still in the publishing process, accepted but who knows when it will appear.

g-weatherill: Isn't this the same model that is already implemented here https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/correlation.py#L126 ? If that is the case the scientific paper is Jayaram & Baker (2009) as cited in the doc string of that class.

Any way to get around having two different implementations of the same formula?

micheles:
You are right, Graeme. Now I remember that I wanted to check the correlation matrix in hazardlib but then, busy with the 3.0 release, I forgot to do that. It makes sense to reuse the formula in hazardlib.

PR #3547, opened 2018-03-21
Extracting the occupancy_periods from the first asset and comparing with the ones declared in the header

micheles:
Now the user gets an error if the declared periods are different from the actual periods. Moreover now all occupancy periods are stored, not only the one specified in the job.ini, thus making possible to reuse exposures with occupants. Closes https://github.com/gem/oq-engine/issues/3466.


openquake/risklib/asset.py
```diff
@@ -505,78 +511,79 @@ def __fromh5__(self, dic, attrs):
...
+            first_asset = assets[0]
+            break
+    else:  # no break
+        raise ValueError('There are no assets!')
+    loss_types = []
+    occupancy_periods = []
+    for name in sorted(first_asset.values):
+        if name.startswith('occupants_'):
+            period = name.split('_', 1)[1]
+            if period != 'None':
```
ptormene: Please check if it can be None

micheles: Yes, the check is still needed for the scenario_risk tests

PR #3548, opened 2018-03-21
flattens lats and lons when calling get_spherical_bounding_box

kejohnso:


micheles:
Would it work to change `geo_utils.get_spherical_bounding_box` and put the `.flatten()` there, so that there is a single change?

mmpagani:
Hi Kendra, can you pls provide a short description of why you think this change is necessary?

kejohnso:
Okay - confirmed that this works with the full source set.

Thanks,
Kendra

micheles:
We need a test exhibiting the issue to be added into openquake/hazardlib/tests/geo/utils_test.py

kejohnso:
Okay, I have added this. I'll open a new pull request.

On Thu, Mar 22, 2018 at 8:19 AM, Michele Simionato <notifications@github.com

Issue #3485, opened 2018-02-28
Groups are not validated when downloading datastores


daniviga:
User groups are not evaluated when a datastore is downloaded via the API

PR #3549, opened 2018-03-21
Fix authentication on API

daniviga:
This PR addresses the following issues:

- The logic behind users access validation was repeated many times in the code, it has been now moved entirely in `utils.py`
- Some endpoints were not honoring groups ACLs
- Some endpoints were not protected by ACLs at all

Now, ACLs can be validated against a resource (like a `job`) with the function

```python
utils.user_has_permission(request, res_owner)
```
which returns `True` if the current logged user has permission to see a resource owned by `res_owner`

Another two helpers are provided:

- `utils.get_user(request)` returns a username following a logic based on the fact that the authentication mechanism is on or off
- `utils.get_valid_users(request)` returns a list of usernames based on groups membership

Closes https://github.com/gem/oq-engine/issues/3485

### DIFF
--- a/openquake/server/utils.py
+++ b/openquake/server/utils.py
@@ -39,7 +39,7 @@ def get_user(request):
         if request.user.is_authenticated():
             user = request.user.username
         else:
-            # This may happens with crafted requests
+            # This may happen with crafted requests
             user = ''
     else:
         user = (settings.DEFAULT_USER if
@@ -61,7 +61,7 @@ def get_valid_users(request):
                 users = list(User.objects.filter(groups__name=groups)
                              .values_list('username', flat=True))
         else:
-            # This may happens with crafted requests
+            # This may happen with crafted requests
             users = []
     return users
 
--- a/openquake/server/utils.py
+++ b/openquake/server/utils.py
@@ -31,7 +31,10 @@ if settings.LOCKDOWN:
 
 
 def get_user(request):
-
+    """
+    Returns the users from `request` if authentication is enabled, otherwise
+    returns the default user (from settings, or as reported by the OS).
+    """
     if settings.LOCKDOWN and hasattr(request, 'user'):
         user = request.user.username
     else:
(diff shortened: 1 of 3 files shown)


openquake/server/views.py
```diff
@@ -298,6 +298,8 @@ def calc_info(request, calc_id):
     """
     try:
         info = logs.dbcmd('calc_info', calc_id)
+        if not utils.user_has_permission(request, info['user_name']):
+            return HttpResponseNotFound()
```
micheles: I would have expect an Unauthorized exception 401 here

openquake/server/views.py
```diff
@@ -540,14 +539,11 @@ def calc_results(request, calc_id):
...
-    user = utils.get_user_data(request)
-
     # If the specified calculation doesn't exist OR is not yet complete,
     # throw back a 404.
     try:
         info = logs.dbcmd('calc_info', calc_id)
-        allowed_users = user['group_members'] or [user['name']]
-        if user['acl_on'] and info['user_name'] not in allowed_users:
+        if not utils.user_has_permission(request, info['user_name']):
             return HttpResponseNotFound()
```
micheles: As before: I would have expect an Unauthorized exception 401 here.

openquake/server/views.py
```diff
@@ -620,8 +616,10 @@ def get_result(request, result_id):
     # the job which it is related too is not complete,
     # throw back a 404.
     try:
-        job_id, job_status, datadir, ds_key = logs.dbcmd(
+        job_id, job_status, job_user, datadir, ds_key = logs.dbcmd(
             'get_result', result_id)
+        if not utils.user_has_permission(request, job_user):
+            return HttpResponseNotFound()
```
micheles: Again

openquake/server/views.py
```diff
@@ -717,10 +713,8 @@ def get_datastore(request, job_id):
         A `django.http.HttpResponse` containing the content
         of the requested artifact, if present, else throws a 404
     """
-    user = utils.get_user_data(request)
-    username = user['name'] if user['acl_on'] else None
-    job = logs.dbcmd('get_job', int(job_id), username)
-    if job is None:
+    job = logs.dbcmd('get_job', int(job_id))
+    if job is None or not utils.user_has_permission(request, job.user_name):
         return HttpResponseNotFound()
```
micheles: Should we distinguish NotFound from Unauthorized?

openquake/server/views.py
```diff
@@ -737,11 +731,10 @@ def get_oqparam(request, job_id):
     """
     Return the calculation parameters as a JSON
     """
-    user = utils.get_user_data(request)
-    username = user['name'] if user['acl_on'] else None
-    job = logs.dbcmd('get_job', int(job_id), username)
-    if job is None:
```
micheles: As before

openquake/server/utils.py
```diff
@@ -39,7 +39,7 @@ def get_user(request):
         if request.user.is_authenticated():
             user = request.user.username
         else:
-            # Should never happens
+            # This may happens with crafted requests
             user = ''
```
micheles: No "s" in "happens" ;-)

openquake/server/utils.py
```diff
@@ -30,28 +30,60 @@
...
         if request.user.is_authenticated():
-            name = request.user.username
+            user = request.user.username
+        else:
+            # This may happen with crafted requests
+            user = ''
+    else:
+        user = (settings.DEFAULT_USER if
+                hasattr(settings, 'DEFAULT_USER') else getpass.getuser())
+
```
micheles: You can do that in one line
```python
user = getattr(settings, 'DEFAULT_USER', getpass.getuser())
```

openquake/server/utils.py
```diff
@@ -30,28 +30,60 @@
...
-    else:
-        name = (settings.DEFAULT_USER if
-                hasattr(settings, 'DEFAULT_USER') else getpass.getuser())
+    return acl_on
+
 
-    return {'name': name, 'group_members': group_members, 'acl_on': acl_on}
+def user_has_permission(request, owner):
+    """
+    Returns `true` if user coming from the request has the permission
```
micheles: Use `True` and `False` in the docstring, please.

PR #3552, opened 2018-03-22
Fixed an issue with GriddedSurfaces

micheles:
This is the minimal change unblocking Kendra. We will keep working on https://github.com/gem/oq-engine/pull/3548 later on.


kejohnso:
okay

Issue #3499, opened 2018-03-05
Using a filtered `region_grid` to compute the hazard for highly distributed exposures


raoanirudh:
For risk calculations involving exposures with assets distributed at a large number of sites, computing the hazard (gmfs / hazard curves) at each exposure site can be computationally expensive and indeed profligate.

A risk analyst can already actively address this issue by precomputing the hazard at a reduced set of sites by providing either: (1) a sites csv file with the list of reduced sites, or (2) a `region` and `region_grid_spacing`. 

The second option could be simplified by inferring the `region` directly from the exposure, and thus requiring the user to specify only the `region_grid_spacing`.

This option could also be enhanced by further discarding sites on the grid that are far away from any assets in the exposure.

The basic sequence of steps to achieve the above could be as follows — if the job configuration file specifies an exposure model and a `region_grid_spacing` but does not specify a `region`:
1. Compute the [convex hull](https://github.com/gem/oq-engine/blob/engine-2.9/openquake/hazardlib/geo/mesh.py#L343-L366) of all of the sites in the exposure, and use this set of points as the `region`
2. Use the `region_grid_spacing` to discretize the `region` into a grid
3. Discard sites from the grid that are farther away than a threshold value equal to ~`region_grid_spacing` / `sqrt(2)`~ ~`asset_hazard_distance`~ `region_grid_spacing` from any asset in the exposure
4. Compute the hazard only on the sites on the grid that weren't discarded
5. Compute the risk for each asset using the hazard results from the nearest available site (nearest-neighbor interpolation).

micheles:
Implemented https://github.com/gem/oq-engine/pull/3553 except the part about the `asset_hazard_distance`, which is left as it was.

raoanirudh:
@micheles Updated the specs above to reflect your comment:
> Implemented #3553 except the part about the asset_hazard_distance, which is left as it was.

raoanirudh:
Updated the Step 3 in the specs again, after discussions with @micheles and @CatalinaYepes

micheles:
There is a problem with exposures with a site model. Currently the site mesh is extracted from the site model in that case; instead, it should be extracted from the exposure. This is what was causing `NaN` site parameters in the scenario_risk for Maule by Catalina. I am fixing the issue in https://github.com/gem/oq-engine/pull/3565.

PR #3553, opened 2018-03-23
Automatic gridding of the exposure

micheles:
Closes https://github.com/gem/oq-engine/issues/3499. A test for Anirudh could be running a calculation with different grid spacing and checking the total losses.


CatalinaYepes:
I have run a test in GemControl (calculation did not finish, Calc_ID 2960) and in my own laptop (116 secs) for an exposure with the following characteristics:
- Number of assets: 72511
- Number of sites: 13038
- Sites for hazard calculations (using 5km grid): 110

Comments:

1. The site grid created for hazard calculation does not include points for the outer most locations in the exposure. Therefore, some exposure points have hazard locations farther away than the desired distance (`region_grid_spacing`).
2. The hazard sites that do not have associated an exposure location closer than the `region_grid_spacing` should be removed. This is specially important for exposure models that have exposure locations with a significant distance between points.

debian/changelog
```diff
@@ -1,4 +1,7 @@
   [Michele Simionato]
+  * Now the engine automatically associate the exposure to a grid if
```
ptormene: associates

PR #3554, opened 2018-03-23
Call `Starmap.init/Starmap.shutdown` in the proper places

micheles:
When using a process pool it is imperative to fork before reading the source model, to avoid using a lot of memory.  This is done by calling `Starmap.init` in the `.run()` method of a calculator before reading the inputs. With this fix, on a calculation by Kendra, on my workstation the memory occupations went down from 50 GB to 5 GB.

While at it, I changed the name of distribution mechanism: what was called `futures` is now called `processpool`. We are not using futures anymore, so I am also changing the requirements (actually the requirements file for Python 2.7 should be remove, but I leave that task to Daniele).

### DIFF
--- a/doc/requirements.md
+++ b/doc/requirements.md
@@ -3,7 +3,6 @@
 ### Main dependencies
  
 * Python - Implementation language
-* concurrent.futures - Standard package for concurrence in Python
 * Django - Used by the API server and the WebUI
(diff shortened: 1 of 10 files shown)


daniviga:
I would change also the `openquake.cfg` and update relative patch in debian

micheles:
Done

PR #3558, opened 2018-03-26
Added a check for wrong hosts in importcalc

micheles:
This should make it more difficult to use the wrong URL.


openquake/commands/importcalc.py
```diff
@@ -23,7 +24,7 @@
 from openquake.server import dbserver
 from requests import Session
 
-CHUNKSIZE = 1024**2  # 1 MB
+CHUNKSIZE = 5*1024**2  # 5 MB
```
daniviga: I would use a multiple of 2 like 4 MB or 8 MB

PR #3561, opened 2018-03-26
20180226 decimal

mmpagani:
This removes the use of decimal in testing that probabilities sum to 1. The test now uses a precision of 1e-12.


micheles:
LGTM if the tests are green

PR #3564, opened 2018-03-27
Change the API to return a proper error code with auth on

daniviga:
With the following PR we

- return a proper HTTP code if an API endpoint under `/v1/` is called without being authenticated (when `LOCKDOWN = True`). This helps API consumers to properly validate the response
- moved `engine_version` and `engine_latest_version` under `/v1/` for the above reason


micheles:
LGTM if the tests are green

PR #3569, opened 2018-03-28
Minor improvements to openquake.hmtk.plotting

nackerley:
The impetus here was a bug introduced by my last pull request whereby the call to _save_image() in plot_recurrence_models() was not getting the wrong arguments and failing. The usage is now consistent wherever it is called; it would be better implemented as a decorator, however.

The commit messages summarize the changes in some detail. The main goal is flexibility, for example if you want to do these plots for a number of zones, sending the results to different axes of plt.subplots().

### DIFF
--- a/openquake/hmtk/plotting/seismicity/catalogue_plots.py
+++ b/openquake/hmtk/plotting/seismicity/catalogue_plots.py
@@ -12,9 +12,6 @@ from matplotlib.colors import LogNorm, Normalize
 
 from openquake.hmtk.seismicity.occurrence.utils import get_completeness_counts
 
-# Default the figure size
-DEFAULT_SIZE = (8., 6.)
-
 
 def build_filename(filename, filetype='png', resolution=300):
     """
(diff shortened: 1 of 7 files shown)


openquake/hmtk/plotting/seismicity/catalogue_plots.py
```diff
@@ -90,11 +88,11 @@ def plot_depth_histogram(
         To sample depth uncertainty choose number of samples
     """
     if ax is None:
-        fig, ax = plt.subplots(figsize=DEFAULT_SIZE)
+        fig, ax = plt.subplots(figsize=figure_size)
     else:
         fig = ax.get_figure()
     # Create depth range
-    if len(catalogue.data['depth']) == 0:
+    if len(catalogue.data['depth']) == 0:  # pylint: disable=len-as-condition
```
g-weatherill: You can just switch this to `if not len(...)`, there is no need to disable pylint's check for this

micheles:
@nackerley, please update the changelog file.

micheles:
LGTM

PR #3572, opened 2018-04-03
Removed duplicated files in the classical_damage tests

micheles:
Over 1,200 lines of duplication in the XML were removed.


micheles:
The classical_damage tests here are just copied from your test cases, Anirudh. There is no test with a nontrivial logic tree. There is a single test with a GMPE logic tree containing 2 GSIMs.

raoanirudh:
Are we following the same numbering of the cases as in oq-risk-tests? [case_4a](https://github.com/gem/oq-risk-tests/tree/master/test/classical_damage/inputs/case_4a) in oq-risk-tests has a GMPE logic tree containing two GSIMs, and [case_master](https://github.com/gem/oq-risk-tests/tree/master/test/classical_damage/inputs/case_master) has two source models (and also two GSIMs).

micheles:
Uhm... I see that we have an issue. `case_master` is there but it is not run by the classical_damage tests!

PR #3578, opened 2018-04-03
Add information about the number of events generated by each source

micheles:
This was requested by @mmpagani. It would be nicer to store which events were generated by each source, but it is more difficult and it is left for the future.


micheles:
The number of events is stored inside the source_info dataset.

PR #3584, opened 2018-04-04
Fixed set_random_years with nonparametric sources

micheles:
Nonparametric sources are used for time-dependent models. In such cases each stochastic event set is independent, so we cannot do what we do usually, i.e. cumulate the years in the `events` dataset.
/For instance for the test event based case_8 there are 2 SES and 4 events happening at the years
(15, 29) for SES1 and (39, 43) for SES2. Without this fix the years would be (15, 29) for SES1 and (89, 93) for SES 2.


micheles:
I agree @raoanirudh and I have changed the code accordingly. Notice that since the ordering of the event loss table is by year, the ordering of the exported CSV files is not different.

PR #3613, opened 2018-04-13
Added scenario-from-shakemap documentation

micheles:
See https://github.com/gem/oq-engine/blob/scenario_from_shakemap/doc/scenario-from-shakemap.md


micheles:
Waiting for Catalina that should give me the full exposure for South America.

Issue #3614, opened 2018-04-13
Check if we can remove some ugly hacks


daniviga:
in particular the one that changes the `PYTHONPATH`. It should not be required anymore:

```python
# this is only executed when openquake is installed via linux binary packages
if os.path.realpath(__file__) == '/opt/openquake/bin/oq':
    # custom pythonpath is pushed in position 1 instead of 0 because
    # sys.path[0] is populated at runtime by python itself. See:
    # https://docs.python.org/2/library/sys.html#sys.path
    sys.path.insert(1, '/opt/openquake/lib/python%d.%d/site-packages' %
                    sys.version_info[:2])
os.environ['PYTHONPATH'] = ":".join(sys.path)
```

https://github.com/gem/oq-engine/blob/master/bin/oq#L23 and https://github.com/gem/oq-engine/blob/master/utils/celery-status#L23

PR #3615, opened 2018-04-13
Convert celery-status in an oq command

daniviga:
This PR converts `celery-status` script into an `oq` command. In this way we can avoid ugly hacks to make the script working when packages are used (common case on multi-node setups).

Will reduce also maintenance costs on our clusters.

```bash
$ oq celery status
==========
Host: celery@gemlatitude5450
Status: Online
Worker processes: 4
Active tasks: 0
==========

Total workers:       4
Active tasks:        0
Cluster utilization: 0.00%
```

Part of #3614

### DIFF
--- a/openquake/commands/celery.py
+++ b/openquake/commands/celery.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 # vim: tabstop=4 shiftwidth=4 softtabstop=4
 #
-# Copyright (C) 2017-2018 GEM Foundation
+# Copyright (C) 2018 GEM Foundation
 #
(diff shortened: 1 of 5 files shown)


openquake/commands/celery.py
```diff
@@ -0,0 +1,78 @@
...
+        ins = app.control.inspect()
+
+        total_workers = 0
+        num_active_tasks = 0
+
+        all_stats = ins.stats()
+        if all_stats is None:
+            print("No active workers")
+            sys.exit(0)
+
```
micheles: I like more `sys.exit("No active workers")` with no print.

PR #3618, opened 2018-04-13
Remove sys.path mangling from 'oq'

### LINKED ISSUE #3614: Check if we can remove some ugly hacks

daniviga:
Finally we can get rid of these lines thanks to the new packaging strategy introduced with OQ 2.9.

Were used only in packages: https://ci.openquake.org/job/zdevel_oq-engine/2738/

Closes #3614


micheles:
LGTM if the packager.sh is happy

PR #3632, opened 2018-04-18
Added `dmg_by_event` and `losses_by_event` outputs

micheles:
Closes https://github.com/gem/oq-engine/issues/3001.


CatalinaYepes:
The output name should be modify from `dmg_by_event` to `Aggregate Event Damages`

CatalinaYepes:
@ptormene should include in the plugin the option to visualize these outputs as csv (the attribute table)

ptormene:
If the new outputs have to be loaded in QGIS exactly as for realizations, with no additional logic, it should require just a very simple change in the plugin.

Issue #3637, opened 2018-04-21
Support for the Aristotle project


micheles:
- [x] run calculations for the full South America to refine the performance
- [x] transfer only the assets within the shakemap region
- [x]  make it possible to set the hazard_calculation_id in the job.ini
- [x] allow to manage both hazard grids and risk grids
- [x] make it possible to plot the distribution of the losses
- [x] manage also the case of truncation_level=None

PR #3644, opened 2018-04-23
Fixed shakemap units

micheles:
The shakemap units are PCTG (percent of g) so we must divide by 100 the generated ground motion fields. Before we were dividing by 100 the intensity value but not the standard deviation, which is inconsistent. Needed for https://github.com/gem/oq-engine/issues/3637.


VSilva:
LGTM

PR #3653, opened 2018-04-26
Removed prefiltering on the workers

micheles:
Currently the engine performs two kinds of prefiltering:

1. a fast prefiltering using rtree on the controller node
2. a slow prefiltering using hazardlib on the workers

The idea of using a single prefiltering only on the controller node was rejected in the past because of the implications on the data transfer. If during the first prefiltering one stores for each source an array of indices with the sites affected by that source, the time to send that array and to unpickle it in the workers can be longer that the time to repeat the filtering. Moreover, in the past the data transfer was happening all at once and increasing it could easily cause a crash of rabbitmq. Nowadays things are different.

First of all, the sending of the sources in done incrementally during prefiltering and we can easily afford to increase the data transfer without crashing rabbitmq.
Secondly, now we have nonparametric sources with GriddedSurfaces. The hazardlib prefiltering of GriddedSurfaces is ultra-slow, slower than anything else we ever found.
Using rtree to prefilter also in the workers is a possibility, but it was rejected because

1. rtree indices are not pickleable, we would have to rebuild them for each task, which is slow
2. storing the indices on a file is not possible because we have the request that calculations
must be possible even on clusters without a shared_dir.

Since the sources have been already prefiltered in the controller node, it is clear that transferring that information is the most economic approach conceptually, and now that rabbitmq is less stressed than in the past (also thanks to the usage of zmq to return back the results, introduced in engine 3.0) this is the most promising possibility. It is also the one that requires the smallest change in the code base.

Here is a test performed on a single nonparametric source from the CCARA model, using around ~30,000 sites, i.e. the shape of a normal tile:

Data transfer master: sitecol=33.47 MB, sources=3.9 MB 
Data transfer this branch: sitecol=33.47 MB sources=6.25 MB

As you see, actually the data transfer is dominated by the site collection, so an increase in the source data is negligible. On the other hand, the improvement on the performance is spectacular:
```
   ============================== ======== ========= ======
   ============================== ======== ========= ======
   total classical (filtering)        17       6.66797   22    
   total classical (master)       1,964    8.88281   22    
```
The speedup is of a factor 115!
I ran a SHARE calculation with the cluster to run an example of calculation without GriddedRuptures and verified that 1) the increase in data transfer is negligible 2) the effect in performance is negligible.

NB: removing the prefiltering on the workers may have an effect on event based calculations. The reason is that the rtree filtering is not the same as the hazardlib prefiltering. This happens in the test case_miriam. Here there is a rupture which

1) according to hazardlib has zero sites with distances < 200 km
2) according to rtree has 11 sites with distances < 200, the distances being:

```python
   array([199.84863308, 199.94852645, 199.85327669, 199.95316774,
          199.85797065, 199.95785936, 199.86271495, 199.96260129,
          199.8675096 , 199.96739354, 199.87235458])
```
As you see, all the distances are very close to 200 km. This is a corner cases. The effect however
can be huge on a test with a small number of events (this happens because the test has only 2 SES
of 50 years each). Since the seeds are generated along with the ruptures, even if a single rupture is not discarded the seeds are completely different and the results can be substantially different. The tests are numerically unstable and this is on purpose, to see what can go wrong (and also to make them fast).

UPDATE: here are the figures for a full CCARA calculation (i.e. not just one nonparametric source):

- Data transfer 1.45 GB -> 5.03 GB: 
  the data transfer in the sources increases by a factor of 30 and becomes dominating compared to the site collection data transfer; still, it is far from being problematic

- Comulative computation time 1,715,227s -> 1,334,414s:
   22% better, time saved in the filtering
- Slowest task 25,930s -> 6,178s:
  4.2 times better, this brings down the computation time from 8 hours to 2 hours!


openquake/hazardlib/source/base.py
```diff
@@ -62,6 +61,16 @@ def weight(self):
...
 
+    @property
+    def nsites(self):
+        """
+        :returns: the number of sites affected by this source
+        """
+        try:
+            return len(self.indices)
+        except AttributeError:
+            return 1
```
ptormene: Please add a comment

PR #3662, opened 2018-04-30
Changed filter_sources_with_rtree->prefilter_sources

micheles:
The parameter `prefilter_sources` now accepts three values:

"rtree": prefilter with rtree
"numpy": prefilter with numpy
"no": do not prefilter at all

 `prefilter_sources=no` is useful when debugging issues with the prefiltering mechanism.


kejohnso:
LGTM

PR #3664, opened 2018-05-02
Fixed the latest IDL bug

micheles:
The cause of the bug was the method `Polygon.get_bbox` which was not considering the International Date Line. I have added a few new tests and performed some refactoring. Also, an IDL bug in the plotting utility was fixed.


kejohnso:
LGTM!

PR #3674, opened 2018-05-04
Rigenerated expected files

micheles:
Fixes https://ci.openquake.org/job/master_oq-engine/4433/
The problem was caused by the recent change in https://github.com/gem/oq-engine/pull/3672: since the test `case_master` was skipped in systems != Ubuntu 16.04 the test was green even if I had forgot to regenerated the expected files. I am unskipping the test to avoid this to happen in the future.


daniviga:
It looks like one rlz is missing in the expected files:

```python
FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/gem/oq-engine/openquake/qa_tests_data/event_based_risk/case_master/expected/loss_maps-rlz-008.csv'
```

PR #3684, opened 2018-05-09
Removed validations on the SourceModel class

micheles:
This was requested by @mmpagani, because for the tools it is easier to pass the investigation time as a float rather than a string. I have also renamed one of the two SourceModel classes as LtSourceModel, to avoid confusions.


mmpagani:
LGTM! Grazie

PR #3688, opened 2018-05-14
Parallelized the source prefiltering

micheles:
The prefiltering is sequential and thus not as fast as it could be. The idea is to parallelize it in the master node: it is enough to store the Rtree index on the file system and then read it from the workers.


openquake/calculators/event_based.py
```diff
@@ -158,12 +158,11 @@ def gen_args(self, csm, monitor):
...
         """
         oq = self.oqparam
-        src_filter = SourceFilter(self.sitecol.complete, oq.maximum_distance,
-                                  oq.prefilter_sources)
+        src_filter = SourceFilter(self.sitecol.complete, oq.maximum_distance)
 
         def weight(src):
             return src.num_ruptures * src.RUPTURE_WEIGHT
-        csm = self.csm.filter(src_filter)
+        csm = self.csm.filter(src_filter, self.monitor('prefilter'))
```
ptormene: Please add case using rtree

Issue #2727, opened 2017-04-21
Check tests numbers on macOS


daniviga:
This must be planned for next sprint, we have at least two tests skipped: see https://github.com/gem/oq-engine/pull/2726 and https://github.com/gem/oq-engine/pull/2589. Also the test gmf_ebrisk case_3 is skipped.

For example the generation of loss maps stats is terribly wrong, the number of losses do not match...

We'll need help from the scientists too.

daniviga:
We should raise priority of this. Another test is broken: https://ci.openquake.org/job/macos/job/master_mac_oq-engine/566/testReport/junit/openquake.calculators.tests.event_based_risk_test/EventBasedRiskTestCase/test_case_4/

We already have some tests skipped and macOS user base is quite relevant.

daniviga:
The last reported case is relevant to https://github.com/gem/oq-engine/issues/2995 and not strictly related to macOS

raoanirudh:
@daniviga, @micheles: I can check the test numbers on macOS 10.14. Which test cases should we be looking at specifically?

daniviga:
I would say that currently the situation on macOS is stable so far. Closing.

PR #3702, opened 2018-05-18
Regenerated some missing expected files

micheles:
This solves the breaking on Jenkins: https://ci.openquake.org/job/master_oq-engine/4458/
Due to the change in the distance formula entered yesterday (https://github.com/gem/oq-engine/pull/3697) the numbers changed but I did not regenerate all of the expected files. It would still be interesting to check if we can avoid skipping the tests on macOS now that we have the new stack of libraries (https://github.com/gem/oq-engine/issues/2727). I suspect that many tests currently skipped could be unskipped.


micheles:
We can leave that to @raoanirudh ! :-)

daniviga:
It seems macOS is not happy about new expected files: https://ci.openquake.org/job/macos/job/master_macos_engine/label=macos,python=python3.5/724/console

PR #3714, opened 2018-05-21
Added case_master test to classical_damage

micheles:
This tests comes from Anirudh's tests. It was forgotten.


ptormene:
LGTM if tests are green

PR #3715, opened 2018-05-22
Check on the Python version

micheles:
This is is useful when one is in the wrong venv.


openquake/commands/__main__.py
```diff
@@ -25,6 +25,10 @@
 from openquake.commonlib import __version__
 from openquake import commands
 
+# check for Python version
+if sys.version < '3.5':
+    sys.exit('Python 3.5 is required, you are using %s', sys.executable)
```
daniviga: I would say `Python 3.5+ is required`.

PR #3718, opened 2018-05-22
create shapefile for sources with different mfds

rcgee:
Previously, it was only possible to perform "oq to_shapefile" for sources with specific mfd typologies. Now it is also possible with the "YoungsCoppersmithMFD" and "arbitraryMFD" typologies.


micheles:
You neeed to add tests, @rcgee. The right place is the SourceModelShapefileConverterTestCase class in openquake/commands/tests/commands_test.py.

micheles:
Please add a line to the changelog. Apart from that, LGTM

PR #3723, opened 2018-05-23
PID was not saved into the DB when CLI was used 

daniviga:
See title


micheles:
Please update the changelog

PR #3727, opened 2018-05-24
get_executing_jobs now return only jobs with an existing pid

daniviga:
Closes https://github.com/gem/oq-infrastructure/issues/59

I would not update the changelog since this is an internal only feature

### DIFF
--- a/openquake/server/db/actions.py
+++ b/openquake/server/db/actions.py
@@ -690,7 +690,7 @@ SELECT %s FROM job WHERE status='executing' ORDER BY id desc''' % fields)
     for r in rows:
         # if r.pid is 0 it means that such information
         # is not available in the database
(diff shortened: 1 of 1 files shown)


openquake/server/db/actions.py
```diff
@@ -668,17 +669,30 @@ def get_results(db, job_id):
...
+    running._fields = fields.split(',')
+
+    query = ('''-- executing jobs
+SELECT %s FROM job WHERE status='executing' ORDER BY id desc''' % fields)
+    rows = db(query)
+    for r in rows:
+        # if r.pid is 0 it means that such information
+        # is not available in the database
+        if psutil.pid_exists(r.pid) and r.pid != 0:
+            running.append(r)
```
micheles: I like more the style
```
if r.pid and psutil.pid_exists(r.pid)
```

PR #3729, opened 2018-05-24
Removed nbytes information from gmdata

micheles:
For unknown reasons only in large calculations the value of nbytes stored in the `gmdata` array is wrong, below the real value of bytes of the `gmf_data/data` array. After spending some hours chasing the whale I decided to remove the nbytes information since it is always useless and sometimes wrong. The right number of bytes is saved as an attribute of `gmf_data/data` and it is also continuously updated during the calculation, so there is no reason to keep another counter in `gmdata`. It was useful years ago, when we did not have the information handy in the datastore.

NB: the `gmdata` array is there only for debugging purposes, it is not used by the engine.


micheles:
Green now.

PR #3736, opened 2018-05-28
Stored the zipped input files

micheles:
The scientists always try to make my life more difficult, for instance by hiding the input files they have been using. With the change here the input files are store (in zipped form) in the datastore, so that I can easily get them when I have to debug a calculation with `oq export job_zip`.

### DIFF
--- a/openquake/calculators/export/__init__.py
+++ b/openquake/calculators/export/__init__.py
@@ -74,6 +74,6 @@ def export_job_zip(ekey, dstore):
     Export the bytes in the job_zip dataset in a job.zip file
     """
     dest = dstore.export_path('job.zip')
-    zbytes = dstore['job_zip'].value + b'\x00' * 4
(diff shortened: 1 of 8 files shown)


openquake/calculators/export/__init__.py
```diff
@@ -59,8 +60,20 @@ def keyfunc(ekey):
...
 import_all('openquake.calculators.export')
+
+
+@export.add(('job_zip', 'zip'))
+def export_job_zip(ekey, dstore):
+    """
+    Export the bytes in the job_zip dataset in a job.zip file
+    """
+    dest = dstore.export_path('job.zip')
+    zbytes = dstore['job_zip'].value + b'\x00' * 4  # magic trick!
```
ptormene: Perhaps a check like the following could make it more robust?
```python
ar = numpy.array(b'Foobar\x00\x00\x00\x00')
ar.nbytes - len(ar.item())
```

PR #3747, opened 2018-05-31
Renamed job_zip -> input_zip

micheles:
And changed the display name to "Input Files"


daniviga:
(Do we have also a command line to get inputs from the datastore?)

micheles:
Yes, the normal `oq engine --export-output`. It is not different from other outputs.

PR #3749, opened 2018-05-31
Added an utility to combine mean curves coming from two computations

micheles:
This is a starting point. This utility could even be removed from the engine. The usage is as follows:

```bash
$ python combine_mean_curves.py -h
usage: combine_mean_curves.py [-h] calc1 calc2

Combine the hazard curves coming from two different calculations. The result
will be the hazard curves of calc1 with the sites in common with calc2 updated
by combining the PoEs of calc2.

positional arguments:
  calc1       first calculation
  calc2       second calculation

optional arguments:
  -h, --help  show this help message and exit

$ python combine_mean_curves.py 48001 48003
Generated combine_48001_48003.hdf5
```


rcgee:
I tested this script using a hypothetical case (i.e. by combining the results of Classical Hazard QA Test, Case 2 computed twice at one site). The combined hazard curve is what I expect, so LGTM

PR #3752, opened 2018-06-01
Allow to set a server name in the WebUI

daniviga:
A server name can be specified to customize the WebUI in case of multiple installations of the Engine are available.

This helps avoiding confusion between different installations when the WebUI is used:

screenshot from 2018-06-01 11-14-43


openquake/server/settings.py
```diff
@@ -164,6 +164,11 @@
 
 FILE_UPLOAD_MAX_MEMORY_SIZE = 1
 
+# A server name can be specified to customize the WebUI in case of
+# multiple installations of the Engine are available. This helps avoiding
+# confusion between different installations when the WebUI is used
+SERVER_NAME = None
+
```
micheles: What about `SERVER_NAME=socket.gesocket.gethostname()`?

micheles:
Please update the changelog.

PR #3765, opened 2018-06-11
Store the estimated GMFs size at the end of the event_based_rupture calculator

micheles:
We have enough information to do that. Also, I have removed some obsolete code.


debian/changelog
```diff
@@ -1,4 +1,5 @@
   [Michele Simionato]
+  * Stored an attributed `events.estimated_gmf_size`
```
ptormene: Please rename this

Issue #3649, opened 2018-04-24
Delete calc does not work properly when `multi_user = True`


daniviga:
```bash
$ oq engine --dc 5
Are you sure you want to (abort and) delete this calculation and all associated outputs?
This action cannot be undone. (y/n): y
Job 5 is complete
Could not remove /home/daniele/oqdata/calc_5.hdf5: [Errno 13] Permission denied: '/home/daniele/oqdata/calc_5.hdf5'
```

This is probably caused by the fact that the deletion of hdf5 file is performed by the DbServer, thus by the `openquake` user, instead of the actual owner of the file (the user that runs `oq`).

micheles:
Moreover `oq reset` does not clean up the database, only the datastore, as discovered by Catalina, even on a single machine like gemmicro.

PR #3766, opened 2018-06-11
Fixed deletion of calculations

micheles:
Closes https://github.com/gem/oq-engine/issues/3649. I have tested it in cluster2 with `oq2 reset -y` which removed correctly all of my calculations.
I have also removed the script `bin/check_demos`: now the demos are tested by QGIS, so it is redundant (and it was not working anyway).


.travis.yml
```diff
@@ -56,9 +56,7 @@ jobs:
             chmod 600 $TRAVIS_BUILD_DIR/.deploy_rsa &&
             eval $(ssh-agent -s) && ssh-add $TRAVIS_BUILD_DIR/.deploy_rsa &&
             scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null /tmp/oqdata.zip travis@ci.openquake.org:/var/www/artifacts.openquake.org/travis/oqdata-${BRANCH}.zip &&
-            oq workers stop &&
-            oq reset --yes &&
```
daniviga: Why we are not testing `oq reset` anymore?

micheles: Because I forgot to restore the removed lines. Fixed here: https://github.com/gem/oq-engine/commit/e1e21031857d9dec33fa8548655d054d92e4dc6a

PR #3778, opened 2018-06-13
Reverted the swmr change

micheles:
I thought it was working. It was not.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,6 +1,3 @@
-  [Michele Simionato]
-  * Computed the hazard curves and maps by using the SWMR mode of h5py 2.8
-
   [Daniele Viganò]
   * Added the hostname in the WebUI page. It can be customize by the user
     via the `local_settings.py` file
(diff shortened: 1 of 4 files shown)


daniviga:
Could be something related to the fact that we are sharing the hdf5 file via NFS?

daniviga:
> The  HDF5  SWMR feature  was tested  on  Linux  clusters  with GPFS  and  Lustre,  Linux  systems  with ext3/4, FreeBSD  with  UFS2,  OS X  with  HFS+,  and  Windows  with  NTFS  and  GPFS. The  SWMR feature  does  not work  on  network  file  systems (NFS, SMB/Windows  file  shares,  and  so  on)  since  write  ordering  may  not be  preserved  on  those  systems.

It's also true that we are not making any write via NFS, we just read from it

micheles:
No, there was a stupid mistake so that the engine was transferring the data via celery instead of reading them from the datastore; having fixed that mistake, it became clear to me that the SWMR feature does not work as I expected, one keeps reading bogus numbers for the PoEs with possibly random reading errors.  This even on a single machine without NFS.

PR #3781, opened 2018-06-13
Decouple zmq from zworkers and bind/connect IPs

### LINKED ISSUE #3780: Bind and connect (zmq) should have different settings
Take into account this case:

- We want zmq reciver to listen on `0.0.0.0`
- We need celery to send back results on a specific host (like `oq-engine-master`).

Currently this is not possible since you need to know the IP in advance and since the same settings is used for both the `bind` and the `connect` (via backurl) hostnames cannot be used.

This is making debug and testing more complex in virtual clusters with automatic deployment and/or with Docker

daniviga:
Fixes #3780. See the ticket for an explanation and https://github.com/gem/oq-builders/pull/96 for a real-life example.

It also provides better explanation of `openquake.cfg` settings for hardcore deployments.

https://ci.openquake.org/job/zdevel_oq-engine/2775/

### DIFF
--- a/openquake/baselib/zeromq.py
+++ b/openquake/baselib/zeromq.py
@@ -151,6 +151,5 @@ class Socket(object):
             return self.zsocket.recv_pyobj()
 
     def __repr__(self):
-        end_point = getattr(self, self.end_point)
-        return '<%s %s %s %s>' % (self.__class__.__name__, end_point,
-                                  SOCKTYPE[self.socket_type], self.mode)
+        return '<%s %s %s>' % (self.__class__.__name__,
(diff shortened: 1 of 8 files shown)


openquake/baselib/zeromq.py
```diff
@@ -152,6 +151,6 @@ def send(self, obj):
             return self.zsocket.recv_pyobj()
 
     def __repr__(self):
-        end_point = getattr(self, 'backurl', self.end_point)
+        end_point = getattr(self, self.end_point)
         return '<%s %s %s %s>' % (self.__class__.__name__, end_point,
```
micheles: Just remove this

micheles:

PR #3788, opened 2018-06-18
Added a Magnitude Frequency Distribution output to the classical calculator

micheles:
This should help the hazard scientists to double check the hazard models. You will see the new output as in this example:

```
[2018-06-18 06:34:59,821 #910 INFO] Calculation 910 finished correctly in 19 seconds
  id | name
  49 | Full Report
  50 | Hazard Curves
  51 | Hazard Maps
  52 | Input Files
  53 | Magnitude Frequency Distribution
  54 | Seismic Source Groups
```
It is exported as a .csv with header (mag, freq).

NB: the feature is NOT implemented for NonParametricProbabilisticRuptures yet, so I am setting the frequency to 0 to make clear that it is not working.


micheles:
Closing because I want to implement this in a better way, after discussion with @mmpagani

PR #3792, opened 2018-06-19
Added a check on duplicated branch IDs in GsimLogicTree

micheles:
We had a check on duplicated branchSetIDs but non on branchIDs.


klunk386:
Just a question. Does this apply only to IDs of the same branch set or to the whole Logic-Tree? I mean, can some branches in different branch sets still have same ID?
For the rest it LGTM.

PR #3795, opened 2018-06-19
oq check_input

micheles:
It would be nice if we had a procedure to run this command on all published GEM models, to make sure that the input files are correct. Also, IPT should run this command.

```bash
$ oq check_input --help
usage: oq check_input [-h] job_ini_or_zip

positional arguments:
  job_ini_or_zip  Check the input files

optional arguments:
  -h, --help      show this help message and exit
```

NB: parsing large source models/exposures can take minutes, so this command can be slow, even if it does not run any computation.


daniviga:
For IPT we'll need to expose this to the API, sice IPT does not have direct access to the Engine.

PR #3798, opened 2018-06-20
Store the sizes of some outputs in the database

micheles:
Now a call to `/v1/calc/ID/results` will return in the dictionary a key `size_mb` which will be `None` for some outputs and a float (the size in MB) for others. The WebUI and the QGIS plugin could use such information to show the size of the output if available (for instance as a tooltip on a mouse over event). This is not urgent at all, but the API is there, should we need it. Refining it for the hazard outputs will require more work. Here is an example for the  event based outputs:

```
[{'id': 27,
  'name': 'Full Report',
  'outtypes': ['rst'],
  'size_mb': None,
  'type': 'fullreport',
  'url': 'http://127.0.0.1:8800/v1/calc/result/27'},
 {'id': 28,
  'name': 'Ground Motion Fields',
  'outtypes': ['xml', 'csv', 'npz'],
  'size_mb': 0.00884246826171875,
  'type': 'gmf_data',
  'url': 'http://127.0.0.1:8800/v1/calc/result/28'},
 {'id': 29,
  'name': 'Hazard Curves',
  'outtypes': ['csv', 'xml', 'npz'],
  'size_mb': 0.027740478515625,
  'type': 'hcurves',
  'url': 'http://127.0.0.1:8800/v1/calc/result/29'},
 {'id': 30,
  'name': 'Earthquake Ruptures',
  'outtypes': ['xml', 'csv'],
  'size_mb': 0.008056640625,
  'type': 'ruptures',
  'url': 'http://127.0.0.1:8800/v1/calc/result/30'},
 {'id': 31,
  'name': 'Seismic Source Groups',
  'outtypes': ['csv'],
  'size_mb': None,
  'type': 'sourcegroups',
  'url': 'http://127.0.0.1:8800/v1/calc/result/31'}]
```


daniviga:
I added

screenshot from 2018-06-20 09-40-02

PR #3808, opened 2018-06-25
20180625 decimaltime

mmpagani:
This PR adds some tests to the function computing decimal time. In case of earthquakes missing some of the timing parameters, these can be replaced with a default value using a bunch of lines like the ones included in one of the two tests.

### DIFF
--- a/openquake/hmtk/seismicity/utils.py
+++ b/openquake/hmtk/seismicity/utils.py
@@ -157,7 +157,20 @@ def decimal_time(year, month, day, hour, minute, second):
     tho = np.zeros_like(year, dtype=int)
     tmi = np.zeros_like(year, dtype=int)
     tse = np.zeros_like(year, dtype=float)
-
+    #
+    # Checking inputs
+    if any(month < 1) or any(month > 12):
+        raise ValueError('Month must be in [1, 12]')
+    if any(day < 1) or any(day > 31):
+        raise ValueError('Day must be in [1, 31]')
+    if any(hour < 0) or any(hour > 24):
+        raise ValueError('Hour must be in [0, 24]')
+    if any(minute < 0) or any(minute > 60):
+        raise ValueError('Minute must be in [0, 60]')
+    if any(second < 0) or any(second > 60):
+        raise ValueError('Second must be in [0, 60]')
+    #
+    # Initialising values
     if any(month):
         tmo = month
     if any(day):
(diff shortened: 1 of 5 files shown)


kejohnso:
Marco I get the tests, I get this failure message:

  File "/Users/kjohnson/GEM/oq-engine/openquake/hmtk/tests/seismicity/test_seismicity_utils.py", line 129, in test_decimal_time2
    utils.decimal_time(self.year, [], [], [], [], []),
  File "/Users/kjohnson/GEM/oq-engine/openquake/hmtk/seismicity/utils.py", line 162, in decimal_time
    if any(month < 1) or any(month > 12):
TypeError: unorderable types: list() < int()

openquake/hmtk/seismicity/utils.py
```diff
@@ -157,7 +157,20 @@ def decimal_time(year, month, day, hour, minute, second):
...
+        raise ValueError('Month must be in [1, 12]')
+    if any(day < 1) or any(day > 31):
+        raise ValueError('Day must be in [1, 31]')
+    if any(hour < 0) or any(hour > 24):
+        raise ValueError('Hour must be in [0, 24]')
+    if any(minute < 0) or any(minute > 60):
+        raise ValueError('Minute must be in [0, 60]')
+    if any(second < 0) or any(second > 60):
+        raise ValueError('Second must be in [0, 60]')
+    #
```
micheles: You can just use
```
datetime.datetime(year, month, day, hour, minute, second)
```

and Python will do the validation for you.

mmpagani: I know Michele this is the function we have currently in the hmtk. With more time we will need to consider replacing it.

mmpagani:
Thanks Kendra for checking. I'll review the code. Quite strange btw since I've been running tests locally.

openquake/hmtk/tests/seismicity/test_seismicity_utils.py
```diff
@@ -126,7 +126,8 @@ def test_decimal_time2(self):
         '''Tests the function utils.decimal_time'''
         self.year = np.array([1990])
         self.assertTrue(np.allclose(
-            utils.decimal_time(self.year, [], [], [], [], []),
+            utils.decimal_time(self.year, np.array([]), np.array([]),
```
micheles:
```
self.assertTrue(np.allclose
```

is horrible. The right way it to use

```
numpy.testing.assert_allclose
```

openquake/hmtk/seismicity/utils.py
g-weatherill: Numpy's vectorised datetime functionality (https://docs.scipy.org/doc/numpy/reference/arrays.datetime.html) does a better job of handling all of the various date and time corrections and gives quite a large speedup. It wasn't available in Python at the time I wrote these modules. The downside of datetime is that it requires that the year is greater than or equal to 1. It is not totally uncommon for a user to want to run the hmtk on catalogues that contain historical events and instrumental ones, which may contain events Before Common Era (represented by negative numbers). Whilst I don't expect these to flag up aftershocks (not real ones anyway) if you wanted to port the code to datetime then one would need to account for this case.

PR #3830, opened 2018-07-05
Increase region_grid_spacing in 2 demos

ptormene:
Loading those demos in the OQ-Engine/IRMT-QGIS-plugin integration tests was quite slow (see https://github.com/gem/oq-irmt-qgis/pull/476). With this minor change, the QGIS loaders for these demos become about 5 times faster (loading hcurves still requires almost 20 seconds, but it's much better than over 100).
I kindly ask @mmpagani, and the scientific team in general, if it's possible to make faster also any of the other demos listed in https://github.com/gem/oq-irmt-qgis/pull/476


micheles:
LGTM

Issue #3833, opened 2018-07-06
Calculation support for geotechnical hazard


g-weatherill:
A comprehensive analysis of seismic risk for a region should, ideally, take into account losses due to secondary effects (landsliding, liquefaction and co-seismic surface rupture). Methodologies for assessment of seismic hazard and risk from geotechnical failure can be found in the scientific literature, and although the methodologies are not as mature as that of conventional ground shaking hazard there is a growing need for them to be integrated into software for practical application.

A preliminary prototype has been created offline and several steps are needed (which will be addressed through different pull requests) to integrate a geotechnical module. In the initial stage only landsliding and liquefaction are addressed, with co-seismic rupture to follow.

Steps:

- [x] Extension of supported IMTs to include different forms of permanent ground displacement

- [x] Extension of the site model to incorporate parameters relevant for secondary hazards. A survey of methods for practical application indicate the following as potential optional site attributes (all scalar): yield acceleration, slope, compound topographic index (CTI), HAZUS liquefaction susceptibility category, HAZUS landsliding susceptibility category, groundwater depth, distance to coast, distance to river, distance to nearest water body, elevation above nearest water body, annual precipitation

- [ ] Functionality for secondary hazards including i.) probability of occurrence of failure, ii.) probability of exceeding a given displacement, iii.) generation of fields of displacement conditioned upon the ground motion level

- [x] Development of a logic tree functionality for applying different geotechnical models

- [ ] Support for defining multiple correlated intensity measures at a given site by either i. integrating over a conditional distribution of secondary measures for a given primary IM (i.e. Vector-PSHA), and/or ii.) incorporation of spatial cross-correlation in the ground motion models (already raised here: https://github.com/gem/oq-engine/issues/3607)

- [x] Support for calculation of ground motion fields from different GMPEs for different IMTs

- [ ] Core calculators for running secondary hazard analysis

- [ ] Definition of vulnerability functions to define losses given multiple inputs, e.g. multiple forms of displacement (i.e. lateral spread + settlement, liquefaction + slope displacement, displacement + ground motion, etc.)

- [ ] Probabilistic fault displacement hazard assessment (PFDHA) module

PR #3834, opened 2018-07-06
Extends site objects for geotechical properties

g-weatherill:
Addresses points 1 and 2 of https://github.com/gem/oq-engine/issues/3833

Adds the following optional attributes to the site object, including reading and validation from a site model: `yield acceleration`, `slope`, compound topographic index (`cti`), HAZUS liquefaction susceptibility category (`liquefaction_susceptibility`), HAZUS landsliding susceptibility category (`landsliding_susceptibility`), groundwater depth (`dw`), distance to coast (`dc`), distance to river (`dr`), distance to nearest water body (`dwb`), elevation above nearest water body (`hwater`), annual precipitation (`precip`)

Adds IMTs for permanent ground displacement phenomena:  Lateral Spread (`PGDfLatSpread`), Settlement (`PGDfSettle`), slope displacement (`PGDfSlope`) and co-seismic rupture (`PGDfRupture`)

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,6 @@
+  [Graeme Weatherill]
+  * Extends site object to include geotechnical (secondary hazards) parameters
+  
   [Michele Simionato]
   * Added a check on source model logic tree files: the uncertaintyModel
     values cannot be repeated in the same branchset
--- a/openquake/commands/tests/commands_test.py
+++ b/openquake/commands/tests/commands_test.py
@@ -225,7 +225,7 @@ class RunShowExportTestCase(unittest.TestCase):
         with Print.patch() as p:
             show_attrs.func('sitecol', self.calc_id)
         self.assertEqual(
-            '__pyclass__ openquake.hazardlib.site.SiteCollection\nnbytes 54',
+            '__pyclass__ openquake.hazardlib.site.SiteCollection\nnbytes 134',
             str(p))
 
     def test_export_calc(self):
--- a/openquake/commonlib/readinput.py
+++ b/openquake/commonlib/readinput.py
@@ -310,6 +310,17 @@ site_model_dt = numpy.dtype([
     ('z1pt0', numpy.float64),
     ('z2pt5', numpy.float64),
     ('backarc', numpy.bool),
+    ('liquefaction_susceptibility', numpy.int16),
+    ('landsliding_susceptibility', numpy.int16),
+    ('dw', numpy.float64),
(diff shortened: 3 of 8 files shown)


mmpagani:
Since these parameters aren't used in the calculators for ground motion calculation, did you consider  having a geotechnical site object (and corresponding collection). In this way the proper site collection object would be sent to the corresponding calculator.

micheles:
@mmpagani 

I still prefer to have single site collection object, but extensible. It would work as follows:

1. the engine reads the GSIM logic tree and therefore the GSIMs and therefore all the collective REQUIRED_SITES_PARAMETERS.

2. when calling `SiteCollection.from_points` the engine will pass as an extra argument the
required sites parameters, then the SiteCollection will be instantiated with an underlying array
which has fields only for the required parameters.

3. the Site object has to be extended, we have to change the site model reader and the site model exporter and a few other things, but I have everything under control.

I will make a PR on Monday with the machinery required for an extensible SiteCollection and then
@g-weatherill  will just have to add the geotermal parameters. As I said, this is something I wanted
to do anyway, a simple mechanism to add new site parameters is a good thing in itself.

mmpagani:
@micheles I don't argue on the utility of having a more flexible management of the site object. My point is different i.e. is it efficient to bring into the ground motion calculation information that will be used only a posteriori (@g-weatherill pls correct me if I'm wrong. btw I suppose you framework you are implementing is mainly designed for an event-based like calculation workflow)?
I think that the site ID and location should be unique but different information should be associated to 
it depending on the analysis.

g-weatherill:
@micheles Yes, I wasn't happy about adding so many new parameters into the site collection, but mostly I'm trying to avoid making too many changes throughout at leverage on the current workflows as far as possible. In reality, depending on the models selected in the geotechnical hazard analysis only a few additional site parameters (maybe no more than two or three) are needed on any given calculation. Your approach makes sense to me, so if there is a general gain for OpenQuake from your revision to the site collection object I'm very happy to adopt this new format.

@mmpagani The frameworks I'm adopting are implementing both event-based and classical. So far I'm trying to keep all of the geotechnical hazard calculators as self-contained as possible (with a get_poes function for classical) to ensure that I minimise the number of changes in the rest of the code. I'm certainly open to having a geotechnical site object if that would improve the efficiency, though I guess that will incur more upstream changes. In any case though, a geotech site object would gain even more so from the changes @micheles proposes.

PR #3841, opened 2018-07-10
Use the README.md as description for PyPi

daniviga:
https://test.pypi.org/project/openquake.engine/3.2.0.dev39/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,6 @@
 # along with OpenQuake. If not, see <http://www.gnu.org/licenses/>.
 
 import os
-import io
 import re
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -40,16 +42,16 @@ def get_version():
     return version
 
 
-version = get_version()
+def get_readme():
+    with io.open(os.path.join(os.path.abspath(os.path.dirname(__file__)),
+                              'README.md'), encoding='utf-8') as readme:
```
micheles: Why `io.open` instead of just `open`?

PR #3844, opened 2018-07-11
Implemented extensible site collections

micheles:
Now only the required site parameters are stored in the site collection (see comments in https://github.com/gem/oq-engine/pull/3834).


openquake/hazardlib/site.py
```diff
@@ -45,9 +45,6 @@ class Site(object):
     :param z2pt5:
         Vertical distance from earth surface to the layer where seismic waves
         start to propagate with a speed above 2.5 km/sec, in km.
-    :param backarc":
-        Boolean value, ``True`` if the site is in the subduction backarc and
-        ``False`` if it is in the subduction forearc or is unknown
```
g-weatherill: I can see it makes sense to remove the "optional" parameters from the doc strings here, but is it possible to put somewhere into the API documentation the list of recognised site parameters and their units?

openquake/hazardlib/site.py
```diff
@@ -56,37 +53,36 @@ class Site(object):
 
         :class:`Sites <Site>` are pickleable
     """
-    _slots_ = 'location vs30 vs30measured z1pt0 z2pt5 backarc'.split()
-
-    def __init__(self, location, vs30, vs30measured, z1pt0, z2pt5,
-                 backarc=False):
-        if not vs30 > 0:
+    def __init__(self, location, vs30=numpy.nan, vs30measured=False,
+                 z1pt0=numpy.nan, z2pt5=numpy.nan, **extras):
```
g-weatherill: To some extent, `z1pt0`, `z2pt5` and `vs30measured` are not so different from other optional attributes, in the sense that only a couple of GMPEs use them, so they would not be critical for every calculation but only those using those particular GMPEs, and it can often be a nuisance to the users to have to set values for these when they are not being used. But, I guess these have been a legacy in OQ for it's entire life, so changing that would probably require some effort.

micheles: Indeed, @g-weatherill , it is just a matter of legacy. I do not want to break all the code instantiating the `Site` object. For `backarc` it is not an issue, since it is a relatively "new" parameter and it has a default, while the others have been mandatory parameters since the beginning of time.

openquake/calculators/tests/event_based_risk_test.py
```diff
@@ -162,6 +162,8 @@ def test_case_3(self):
     def test_case_4(self):
         # Turkey with SHARE logic tree
         self.run_calc(case_4.__file__, 'job.ini')
+        # if os.environ.get('TRAVIS'):
```
ptormene: Please remember to remove this

openquake/hazardlib/site.py
micheles: See https://docs.openquake.org/oq-engine/master/openquake.hazardlib.html#openquake.hazardlib.site.SiteCollection

PR #3846, opened 2018-07-11
Fixed heisenbug by sorting the dictionary of parameters

micheles:
Without the sort scenario/test_case_4 was breaking randomly on Jenkins, due to a vs30=0 causing an infinite GMF.


micheles:
LGTM

PR #3847, opened 2018-07-11
Fixed `gmf_data/indices` to work with HDFView 3.0

micheles:
This change is backward-compatibile!


micheles:
After the fix in https://github.com/gem/oq-engine/pull/3855 I found a better solution, so I am closing this,

PR #3850, opened 2018-07-11
Move some Jenkins customizations from Sphinx Makefile to packager

### LINKED ISSUE #3849: I cannot build locally the documentation
The reason is that doc/sphinx/Makefile contains the line
```
SPHINXBUILD   = /opt/openquake/bin/python3 /opt/openquake/bin/sphinx-build
```
To build the docs locally in a virtualenv I would need the line
```
SPHINXBUILD   = sphinx-build
```
Care should be taken to avoid breaking Jenkins.

daniviga:
In this way we'll have a standard makefile that can be used locally or on Travis but at the same time we are not breaking jenkins.

Fixes #3849 

https://ci.openquake.org/job/zdevel_oq-engine/2788/

### DIFF
--- a/doc/sphinx/Makefile
+++ b/doc/sphinx/Makefile
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    = -j 8 # -n -W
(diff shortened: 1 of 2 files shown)


micheles:
LGTM

PR #3854, opened 2018-07-12
First step towards python 3.7 compatibility

daniviga:
Wheels will come later on...


micheles:
There is no urgency @daniviga . Python 3.7 could come even next year, so there is no point in preparing the wheels now.

PR #3856, opened 2018-07-13
Fixed `gmf_data/indices` to work with HDFView 3.0

micheles:
This is compatible with the past, i.e. we can still run the risk starting from GMFs generated/imported with old versions of the engine.


micheles:
Superseded by https://github.com/gem/oq-engine/pull/3857

Issue #3607, opened 2018-04-11
Spatial Cross-Correlation in GMFs and Related Issues


g-weatherill:
A recent PR (https://github.com/gem/oq-engine/pull/3545) related to issue https://github.com/gem/oq-engine/issues/3544 introduced spatial cross-correlation into the oq-engine for shakemap applications. 

Briefly, spatial cross-correlation is a generalised form of spatial correlation for application when multiple IMTs are required and the correlations between IMTs at the same site and at different sites needs to be taken into account. When implementing this it is no longer possible to build or treat each IMT field separately in the simulation. 

The implementation of spatial cross-correlation in https://github.com/gem/oq-engine/pull/3545 hard-codes one particular approach. However, there are a growing number of models in the literature, some may need to be constructed differently from the approach implemented here, and depending on application it may be necessary to support more of them. Furthermore, implicit within a spatial cross-correlation model is the property that for two different IMTs located at the same site, the correlation should collapse to the known cross-correlation between IMTs. This can be a useful property to consider for many applications as it opens up the eventual possibility of being able to run vector hazard analyses.

To this end it may be beneficial to refactor the correlation model objects to support the more general cases when one wishes to consider spatial correlation, cross-correlation and spatial cross-correlation, whilst still permitting the possibility of running the calculation with only the spatial correlation and not the cross-correlation. This would need some additional changes to the GMFComputer to support this case.

Additionally:
1. For various reasons, many spatial cross-correlation models apply only to the total standard deviation, as they are derived by considering the total residual only and not by explicitly separating our the inter-event residual correlation. OpenQuake's current approach of raising an error when applying spatial correlation to a GMPE that does not have inter- an intra-event standard deviations, rather than just applying the spatial correlation to the total residual, was never actually necessary, as it would have not have been a problem to apply the correlation to the total residual, given that the inter-event residual is constant for a given IMT. This behaviour should be changed anyway and both spatial correlation and spatial cross-correlation should be permitted when the GMPE returns total standard deviation. Depending on the spatial cross-correlation model it should only apply to the total residual in any case.

[...]

PR #3865, opened 2018-07-16
Implements a MultiGMPE - calling different IMTs from different GMPEs

g-weatherill:
The implementation of geotechnical hazard calculations requires defining ground motions for multiple different intensity measure types for the same rupture. This PR addresses point 6 of #3833, implementing a "WrapperGMPE" that is instantiated with a dictionary (or string) mapping different GMPEs to IMTs. This approach minimises the number of additional changes required in order call multiple ground motions from the same rupture, as needed for ground motion fields and classical PSHA. Also lays groundwork for addressing point 5 of #3833 as well as #3607.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,9 +1,13 @@
+  [Michele Simionato]
+  * Saved the size of the datastore in the database and used it in the WebUI
+
   [Graeme Weatherill]
   * Adds geotechnical related IMTs
-  * Adds a "WrapperGMPE" object that can call different GMPEs for different
+  * Adds a "MultiGMPE" object that can call different GMPEs for different
     intensity measure types
-  
+
   [Michele Simionato]
+  * Renamed /extract/agglosses -> /extract/agg_losses and same for aggdamages
   * Supported equivalent epicentral distance with a `reqv_hdf5` file
   * Fixed the risk from ShakeMap feature in the case of missing IMTs
   * Changed the way gmf_data/indices and ruptures are stored
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,5 +1,7 @@
   [Graeme Weatherill]
   * Adds geotechnical related IMTs
+  * Adds a "WrapperGMPE" object that can call different GMPEs for different
+    intensity measure types
   
   [Michele Simionato]
   * Supported equivalent epicentral distance with a `reqv_hdf5` file
--- a/doc/sphinx/openquake.hazardlib.gsim.rst
+++ b/doc/sphinx/openquake.hazardlib.gsim.rst
@@ -579,6 +579,13 @@ montalva_2016
     :undoc-members:
     :show-inheritance:
 
+multi_gsim
+------------------------------------------------
+.. automodule:: openquake.hazardlib.gsim.multi_gsim
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
 munson_thurber_1997
 ---------------------------------------------
 
@@ -751,13 +758,6 @@ utils_swiss_gmpe
 ------------------------------------------------
 
 .. automodule:: openquake.hazardlib.gsim.utils_swiss_gmpe
-    :members:
-    :undoc-members:
-    :show-inheritance:
-
-wrapper
-------------------------------------------------
-.. automodule:: openquake.hazardlib.gsim.wrapper
     :members:
     :undoc-members:
     :show-inheritance:
--- a/doc/sphinx/openquake.hazardlib.gsim.rst
+++ b/doc/sphinx/openquake.hazardlib.gsim.rst
@@ -755,6 +755,13 @@ utils_swiss_gmpe
     :undoc-members:
     :show-inheritance:
 
+wrapper
+------------------------------------------------
+.. automodule:: openquake.hazardlib.gsim.wrapper
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
 youngs_1997
 -------------------------------------------
 
--- a/openquake/calculators/event_based_risk.py
+++ b/openquake/calculators/event_based_risk.py
@@ -341,7 +341,8 @@ class EbrCalculator(base.RiskCalculator):
         with self.monitor('building agg_curves', measuremem=True):
             array, arr_stats = b.build(dstore['losses_by_event'].value, stats)
         self.datastore['agg_curves-rlzs'] = array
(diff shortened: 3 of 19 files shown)


g-weatherill:
Switching the name to multiGMPE makes sense, so this is now fixed

For most use cases the input to the MultiGMPE is a string indicating the mapping of the GMPEs to the IMTs. Something like:
`{PGA: BooreEtAl2014, SA(0.5): BindiEtAl2014Rjb}`

For the case of the GMPETable, this would work like:
`{PGA: BooreEtAl2014, SA(1.0): GMPETable(gmpe_table=path/to/table.hdf5)}`
a test case for this is added.

For the XML, so far in prototyping this has been attached to a particular geotechnical calculator, and since that needs it's own logic tree then it takes the form as below, with the GMPE being instantiated from the string inside the `uncertaintyModel` text:
```xml
<?xml version="1.0" encoding="UTF-8"?>

<nrml xmlns:gml="http://www.opengis.net/gml"
      xmlns="http://openquake.org/xmlns/nrml/0.4">
    <logicTree logicTreeID='lt1'>
        <logicTreeBranchingLevel branchingLevelID="bl1">
            <logicTreeBranchSet uncertaintyType="gmpeModel" branchSetID="bs1"
                    applyToTectonicRegionType="Active Shallow Crust">
                <logicTreeBranch branchID="b11">
                    <uncertaintyModel>{PGA: BooreEtAl2014, SA(1.0): AkkarEtAlRjb2014}</uncertaintyModel>
                    <gdemModel>HAZUSLiquefaction</gdemModel>
                    <uncertaintyWeight>0.5</uncertaintyWeight>
                </logicTreeBranch>
                <logicTreeBranch branchID="b12">
                    <uncertaintyModel>{PGA: BindiEtAl2014Rjb, SA(1.0): AkkarBommer2010}</uncertaintyModel>
                    <gdemModel>HAZUSLiquefaction</gdemModel>
                    <uncertaintyWeight>0.5</uncertaintyWeight>
                </logicTreeBranch>
            </logicTreeBranchSet>
        </logicTreeBranchingLevel>
    </logicTree>
</nrml>
```
This has its own logic tree reader (coming in a future PR), which is a subclass of the GSIM Logic Tree with some minor modifications. This is simply because a given geotechnical hazard calculation branch has to combine both the choice of calculator and the choice of models. This could be replaced by having a separate logic tree for the GSIMs and a separate logic tree for the choice of geotechnical calculator, but this requires a substantial change of the logic tree calculators in general given that certain geotechnical calculators will require multiple IMTs. Setting a given branch and associating it to its given GMPE-IMT set in this manner would meet all of the use cases that we have encountered so far in practice (which, admittedly, is not many)

micheles:
I am not happy with the proposed syntax of `{PGA: BooreEtAl2014, SA(1.0): GMPETable(gmpe_table=path/to/table.hdf5)}`. It feels alien inside an XML file. We need something that looks like XML, so that the node library can read it. Also the XML syntax takes care of the quotes properly.

micheles:
We need to start from the input files. Consider for instance the test
https://github.com/gem/oq-engine/tree/master/openquake/qa_tests_data/classical/case_22.
In that case the IMTs are PGA, SA(0.1), SA(0.2), SA(0.5), SA(1.0), SA(2.0).
The logic tree is
```
            <logicTreeBranchSet uncertaintyType="gmpeModel" branchSetID="bs1"
                    applyToTectonicRegionType="Active Shallow Crust">

                <logicTreeBranch branchID="Youngs1997">
                  <uncertaintyModel>
                    YoungsEtAl1997SInterNSHMP2008
                  </uncertaintyModel>
                  <uncertaintyWeight>0.25</uncertaintyWeight>
                </logicTreeBranch>

                <logicTreeBranch branchID="Sadigh_et_al_1997">
                    <uncertaintyModel>SadighEtAl1997</uncertaintyModel>
                    <uncertaintyWeight>0.25</uncertaintyWeight>
                </logicTreeBranch>

                <logicTreeBranch branchID="Abrahamson_Silva_1997">
                    <uncertaintyModel>AbrahamsonSilva1997</uncertaintyModel>
                    <uncertaintyWeight>0.25</uncertaintyWeight>
                </logicTreeBranch>

                <logicTreeBranch branchID="Campbell_Bozorgnia_2003">
                  <uncertaintyModel>
                    CampbellBozorgnia2003NSHMP2007
                  </uncertaintyModel>
                  <uncertaintyWeight>0.25</uncertaintyWeight>
                </logicTreeBranch>

            </logicTreeBranchSet>
```
How could this change? Could we have something like this?
```
            <logicTreeBranchSet uncertaintyType="gmpeModel" branchSetID="bs1"
                    applyToTectonicRegionType="Active Shallow Crust">
                <logicTreeBranch branchID="multiGMPE">
                  <uncertaintyModel>
                    <multiGMPE class="YoungsEtAl1997SInterNSHMP2008" imt="PGA"/>
                    <multiGMPE class="Sadigh_et_al_1997" imt="SA(0.1)"/>
                    <multiGMPE class="Abrahamson_Silva_1997" imt ="SA(0.2)"/>
                    <multiGMPE class="Abrahamson_Silva_1997" imt ="SA(0.5)"/>
                    <multiGMPE class="CampbellBozorgnia2003NSHMP2007" imt="SA(1.0)"/>
                    <multiGMPE class="CampbellBozorgnia2003NSHMP2007" imt="SA(2.0)"/>
                  </uncertaintyModel>
                  <uncertaintyWeight>1.0</uncertaintyWeight>
                </logicTreeBranch>
            </logicTreeBranchSet>
```
If the job.ini lists IMTs which are not in the gsim_logic_tree, is it an error? I am also worried about your introduction of  `DEFINED_FOR_TECTONIC_REGION_TYPE = const.TRT.UNDEFINED`. The engine groups the GSIMs by tectonic region type, so there MUST be in the gsim_logic_tree a session applyToTectonicRegionType.
Rather than code, it would be more useful if you could write down a set of specifications with the expected input files, the expected outputs and the foreseeable error situations. Then I can write down the tests that initially will be stubs and then will become real once the implementation grows out.
Essentially, let's restore the workflow we used with Damiano.

g-weatherill:
Thanks @micheles. I think your logic tree suggestion works better. I had been trying to leverage on the existing logic tree architecture, but if a more significant change is warranted regardless then I think there is a cleaner way to approach the problem. I will close the PR, do as you suggest and revise.

Issue #3864, opened 2018-07-16
Raise reference Python to 3.6


daniviga:
Meta-issue for Python 3.6 migration

| ITEM | PR | STATUS | MERGE ORDER |
| ----------------|--------|-----|---|
| oq-python-deb | https://github.com/gem/oq-python-deb/pull/14 | merged | 1 |
| oq-python-rpm | https://github.com/gem/oq-python-rpm/pull/6  | merged | 1 |
| oq-libs | https://github.com/gem/oq-libs/pull/40 | merged | 2  |
| env | https://github.com/gem/oq-builders/pull/102 | merged |  |
| Windows | https://github.com/gem/oq-builders/pull/102  | merged |  |
| opt | https://github.com/gem/oq-builders/pull/102| merged |  |
| env | https://github.com/gem/oq-builders/pull/102 | merged |  |
| Docker | https://github.com/gem/oq-builders/pull/103 | merged  | 3 |
| Engine | #3868 | merged | 4 |
| VM | na | pending | 5 |

### Changes in oq-engine  (#3868)

- [x] ~Add deprecation warning for Python 3.5 in `setup.py`~
- [x] Add deprecation warning for Python 3.5 in `openquake/commands/__main__.py`

### Documentation  (#3868)

- [x] Development install
- [x] Other references to Python 3.5 and snippets

### Other business

- [x] oq-engine-celery

micheles:
Note: one should remember to change `openquake/commands/__main__.py`.

daniviga:
Good catch @micheles

daniviga:
Closing

PR #3868, opened 2018-07-17
Update the documentation for Python 3.6

daniviga:
Also add a deprecation warning.

I would not add the deprecation warning to the `setup.py` since it is not propagated to the user (unless an hard fail is performed, see https://github.com/pypa/pip/issues/2933)

I would merge this just before the next release to avoid confusion between description and links to the installers. At least it should merged after everything else in #3864 has been completed.
Tests are green here: https://ci.openquake.org/job/zdevel_oq-engine/2794/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 language: python
 
 python:
- - "3.5"
+ - "3.6"
 
 # FIXME temporarily disable cache
 # cache:
@@ -31,7 +31,7 @@ jobs:
         - nosetests --with-doctest -xvs -a'!slow' openquake.commonlib
         - nosetests --with-doctest -xvs -a'!slow' openquake.commands
         - oq webui migrate
-    after_success:  # old sphinx does not work well with Python 3.5.4
+    after_success:  # old sphinx does not work well with Python 3.5.4+
         - pip install sphinx==1.6.5 && cd doc/sphinx && make html
     after_script:
         - oq dbserver stop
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         - nosetests --with-doctest -xvs -a'!slow' openquake.commonlib
         - nosetests --with-doctest -xvs -a'!slow' openquake.commands
         - oq webui migrate
-    after_success:  # old sphinx does not work well with Python 3.5.4
+    after_success:  # old sphinx does not work well with Python 3.5.4+
```
nastasi-oq: Do you mean "+" -> "or newer"

doc/requirements.md
```diff
@@ -18,7 +18,7 @@ A 64bit operating system and 64bit capable hardware are required.
 
 ### Python dependencies
 
-Python 3.5 or Python 3.6.
+Python 3.6+
```
nastasi-oq: Do you still mean "+" -> "or newer" ?

openquake/commands/__main__.py
```diff
@@ -25,9 +25,14 @@
 from openquake.commonlib import __version__
 from openquake import commands
 
+PY_VER = sys.version_info[:3]
+
 # check for Python version
-if sys.version < '3.5':
+if PY_VER < (3, 5):
     sys.exit('Python 3.5+ is required, you are using %s', sys.executable)
```
nastasi-oq: "+" is misleading IMHO, please use "newer version" instead.

PR #3874, opened 2018-07-19
Update atkinson_2015.py

krisvanneste:
Changed CoeffsTable, which was completely wrong, to the actual values in the paper.


krisvanneste:
The PDF you are referring to does not appear to be the final published version, it lacks page numbers.
Have a look at the attached PDF, which I retrieved from the BSSA website:
https://pubs.geoscienceworld.org/ssa/bssa/article-pdf/105/2A/981/3666370/981.pdf
[Atkinson_Bulletin of the Seismological Society of America_2008.pdf](https://github.com/gem/oq-engine/files/2210250/Atkinson_Bulletin.of.the.Seismological.Society.of.America_2008.pdf)

mmpagani:
Yes, I see. In order to accept your PR you also need to provide the associated verification tables https://github.com/gem/oq-engine/tree/master/openquake/hazardlib/tests/gsim/data/ATKINSON2015 
I think that for producing them we asked Gail a code implementing the GMPE; @g-weatherill can confirm.

krisvanneste:
The only way I could produce these verification tables (see attached) is with a script that calls the Atkinson2015 class itself, so this is a bit like circular reasoning. If these tables were really produced using code from Gail Atkinson, it would be best to contact the author about this important issue.

[ATKINSON2015_verification_tables.zip](https://github.com/gem/oq-engine/files/2210696/ATKINSON2015_verification_tables.zip)

g-weatherill:
Yes, we received the original coefficients tables and Excel implementation directly from the author, along with the pre-print of the publication. In the Excel tables there were two sets of coefficients referred to, but it seems that in the final paper only one is published and the other "version" of the GMPE simply referred to in the discussion. So I agree with @krisvanneste the coefficients should be switched to the ones in the paper, although I can see merit in keeping both implementations in separate classes. From the Excel file I think I have enough information to produce the test tables, but if you wish to update them in this same PR then it could be cleaner. Let me know what you prefer

krisvanneste:
I still think it would be useful to contact Atkinson about the original coefficients. I doubt they are correct. To further document this, I attach a response spectrum obtained with the original coefficients, and the corresponding response spectrum obtained with the updated coefficients. The unrealistic shape of the former is what prompted me to investigate what was going wrong.

Response spectrum - original coefficients:
rs_atkinson2015_original_coeffs

Response spectrum - updated coefficients:
rs_atkinson2015_new_coeffs

Let me know if you need further input from me, but note that I'm on holiday the next two weeks.

AflahZul:
Greetings, is the Atkinson 2015 updated?

Issue #3583, opened 2018-04-04
Investigate ScenarioDamage with NaNs for CCARA


micheles:
Here are the files: [sd-ccara.zip](https://github.com/gem/oq-engine/files/1874338/sd-ccara.zip)
The problems seems to be the missing noDamageLimit. Here are the full specs for the noDamageLimit feature:  https://docs.google.com/document/d/1xfBqflXY-Y8hN9lAazJy55PPt0c44qWMuonY1BHG6p4/edit

PR #3875, opened 2018-07-20
Removed 0-levels from the fragility functions

micheles:
The 0-levels cause issues when computing the annual frequency of exceedence with the formula 
`-numpy.log(1. - poe) / t_haz` because they correspond to a poe=1. At the end, this produces NaNs in the outputs (see also https://github.com/gem/oq-engine/issues/3583). Here I am removing the 0-levels from the fragility functions. Since we may have already  0-levels in the fragility functions in our vulnerability database I cannot flag those as errors, but they should.
A good thing would be to upgrade IPT to reject 0-levels.


micheles:
You have an impressive memory and searching skills @raoanirudh, I had forgotten about that PR of 2.5 years ago. This PR is in the same direction but more strict. Still not strict enough to solve Catalina's problem in https://github.com/gem/oq-engine/issues/3583.

micheles:
Closed because it does not solve Catalina's problem. It is better to think more.

Issue #3883, opened 2018-07-26
New IMTs and corresponding GMPEs


pheresi:
Different authors have proposed novel IMTs that are better correlated with damage and collapse. Specifically, these are SDI (inelastic displacement ordinate), SAAVG (average Sa) and FIV3 (filtered incremental velocity 3). Ground motion prediction models have been implemented and I'll include them in separate pull requests:

- [x] Changes in openquake/hazardlib/imt.py, including the definitions of the new IMTs

- [x] Changes in openquake/hazardlib/tests/gsim/check_gsim.py, in order to be able to read test tables for GMPEs of these new IMs

- [x] Changes in openquake/commonlib/oqvalidation.py, in order to check new IMTs 

- [ ] New gsim and gsim_test files for the GMPM of SDI

- [ ] New gsim and gsim_test files for the GMPM of SAAVG

- [ ] New gsim and gsim_test files for the GMPM of FIV3

micheles:
Nothing is happening here, so I am closing the ticket.

PR #3885, opened 2018-07-26
Change to imt.py to add new IMTs

pheresi:
Different authors have proposed novel IMTs that are better correlated with damage and collapse. Specifically, these are SDI (inelastic displacement ordinate), SAAVG (average Sa) and FIV3 (filtered incremental velocity 3).

This is part of issue #3883 for implementing new IMTs and their GMPEs.


pheresi:
Hi, one question: 
I also have these two changes:
- Changes in openquake/hazardlib/tests/gsim/check_gsim.py, in order to be able to read test tables for GMPEs of the new IMs
-  Changes in openquake/commonlib/oqvalidation.py, in order to check new IMTs

Should I include those two files in this PR?

micheles:
Yes, please add those changes. I want to see everything which is needed. My goal is to change the engine to make it possible for an user to add a new IMT without having to touch anything in the code base. Not sure if this is possible, but probably we can get very close to that.

pheresi:
Ok, good. I'm gonna add the changes right now. It's a good idea to make it possible to add new IMs in an easier way. Good luck with that!

micheles:
Superseded by https://github.com/gem/oq-engine/pull/3921

PR #3907, opened 2018-08-07
Adapts hmtk source model parser for cases when name is missing in the source model

g-weatherill:
Earlier usages of the hmtk could take into account source models for which the name is not present. This is causing an error in the current parser re-running previous files, and the name is not a critical feature of the workflow. This PR deals with this case.

### DIFF
--- a/openquake/hmtk/parsers/source_model/nrml04_parser.py
+++ b/openquake/hmtk/parsers/source_model/nrml04_parser.py
@@ -461,9 +461,14 @@ class nrmlSourceModelParser(BaseSourceModelParser):
         sm_node = node_from_xml(self.input_file)[0]
         if sm_node[0].tag.startswith('{http://openquake.org/xmlns/nrml/0.4}'):
             node_sets = [sm_node]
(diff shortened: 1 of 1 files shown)


openquake/hmtk/parsers/source_model/nrml04_parser.py
```diff
@@ -461,9 +461,14 @@ def read_file(self, identifier, mfd_spacing=0.1, simple_mesh_spacing=1.0,
...
             node_sets = [sm_node]
+            if "name" in sm_node.attrib:
+                sm_name = sm_node["name"]
+            else:
+                sm_name = ""
         else:  # format NRML 0.5+
             node_sets = sm_node
-        source_model = mtkSourceModel(identifier, name=sm_node["name"])
+            sm_name = sm_node["name"]
+        source_model = mtkSourceModel(identifier, name=sm_name)
```
micheles: Perhaps we can replace the `if` with a call `sm_node.get("name", "")`?

PR #3909, opened 2018-08-13
Equivalent distances for point and area sources

rcgee:
This PR builds on #3823 and adds the ability to use equivalent distances instead of finite rupture distances (Rjb, Rrup) for Point and Area sources. This allows background sources with random strike to be assigned equivalent (average) distances, following what is done in the USGS Fortran code (e.g. Petersen et al., 2008 Appendix C). 

Equivalent distances are specified in a lookup table (indicated by the reqv_hdf5 flag in the job.ini) and provide the equivalent horizontal distance to a rupture (Reqv) given an epicentral distance and magnitude. Then Rjb =Reqv. The equivalent closest distance to the rupture is computed using Reqv and the hypocentral depth, which replaces Rrup.

Fault sources are not affected by this correction. This is tested in classical test case 14 (see #3823), where reqv_hdf5 is specified in the job.ini but the resulting hazard curves don't change since we are only using a fault source.


openquake/hazardlib/contexts.py
```diff
@@ -185,9 +187,14 @@ def make_contexts(self, sites, rupture):
...
-            if param == 'repi' and self.reqv:
-                distances = self.reqv.get(distances, rupture.mag)
             setattr(dctx, param, distances)
+        if self.reqv and reqv_point:
+            reqv = self.reqv.get(getattr(dctx, 'repi'), rupture.mag)
+            if 'rjb' in self.REQUIRES_DISTANCES:
+                setattr(dctx, 'rjb', reqv)
+            if 'rrup' in self.REQUIRES_DISTANCES:
+                reqv_rup = numpy.sqrt(reqv**2 + rupture.hypocenter.depth**2)
+                setattr(dctx, 'rrup', reqv_rup)
```
micheles: You can just write `dctx.rrup = reqv_rup`

openquake/hazardlib/contexts.py
```diff
@@ -185,9 +187,14 @@ def make_contexts(self, sites, rupture):
         sites, dctx = self.filter(sites, rupture)
         for param in self.REQUIRES_DISTANCES - set([self.filter_distance]):
             distances = get_distances(rupture, sites, param)
-            if param == 'repi' and self.reqv:
-                distances = self.reqv.get(distances, rupture.mag)
             setattr(dctx, param, distances)
+        if self.reqv and reqv_point:
+            reqv = self.reqv.get(getattr(dctx, 'repi'), rupture.mag)
+            if 'rjb' in self.REQUIRES_DISTANCES:
+                setattr(dctx, 'rjb', reqv)
```
micheles: You can just write `dctx.rjb = reqv`

openquake/hazardlib/contexts.py
```diff
@@ -185,9 +187,14 @@ def make_contexts(self, sites, rupture):
         sites, dctx = self.filter(sites, rupture)
         for param in self.REQUIRES_DISTANCES - set([self.filter_distance]):
             distances = get_distances(rupture, sites, param)
-            if param == 'repi' and self.reqv:
-                distances = self.reqv.get(distances, rupture.mag)
             setattr(dctx, param, distances)
+        if self.reqv and reqv_point:
+            reqv = self.reqv.get(getattr(dctx, 'repi'), rupture.mag)
```
micheles: You can just use `dctx.repi`

openquake/hazardlib/contexts.py
```diff
@@ -208,8 +215,6 @@ def poe_map(self, src, sites, imtls, trunclevel, rup_indep=True):
             initvalue=rup_indep)
         eff_ruptures = 0
         with self.ir_mon:
```
micheles: I thought that setting `src.magnitude_scaling_relationship = pointMSR` was needed

openquake/hazardlib/contexts.py
```diff
@@ -175,6 +172,11 @@ def make_contexts(self, sites, rupture):
             Instance of
             :class:`openquake.hazardlib.source.rupture.BaseRupture`
 
+        :param reqv_point:
+            True if reqv is set and source is a point or area source.
+            If True, rjb and rrup are replaced in the distances context
+            by equivalent values.
+
```
micheles: It this new boolean parameter needed? I mean, we have already  the `self.reqv` boolean, can't we use that and only that?

rcgee: I added the boolean as a way to distinguish between point and area sources (for which the equivalent distance correction should be applied) and fault sources (for which the equivalent distance correction should NOT be applied, even when self.reqv is set). 

Point and area sources are distinguished by the line: if hasattr(src, 'location') or hasattr(src, 'area_discretization')

I'm open if you know a better way of doing this.

openquake/hazardlib/contexts.py
rcgee: ok, will do

openquake/hazardlib/contexts.py
rcgee: will do

openquake/hazardlib/contexts.py
rcgee: Why did you think that? I don't see why it would be needed, since we are just replacing rjb and repi (which have been computed using any MSR) with the equivalent distances.

Maybe Marco can comment in case I'm misunderstanding, since I think he discussed with you the previous PR.

openquake/hazardlib/contexts.py
micheles: Point sources and area sources produces PlanarSurfaces, so you can remove the flag and add a check like

```python
if self.reqv and isinstance(rupture, PlanarSurface)
```
```

mmpagani:
Hi, I tested this implementationby running a comparison between the results provided by the OQ engine, by OQ engine with requiv and with my local implementation of the USGS code. 

I attach a figure showing that OQ engine requiv and USGS-like code provide the same results. I did this comparison some time ago using the version I discussed with Michele and the results where similar, hence there is still something I don't understand.

image

rcgee:
Hi Marco, thanks for this comparison. One thing I noticed in your notebook  rupture_site_distance_analysis.ipynb are these two lines in cell 10:

setattr(dctx, 'rjb', reqi)
setattr(dctx, 'rjb', repi)

Both lines are uncommented, which means the code is doing rjb=epi, however as we discussed in our email exchange, we need to do Rjb = Reqv. 

I think this is something worth checking. I wonder if the results would still be similar if the second line is commented out

mmpagani:
I updated the nb yesterday. Now it works with the changes recently introduced by Michele on site parameters. I sent you the new nb on Skype.

micheles:
LGTM from me too.

Issue #3919, opened 2018-08-18
Make the IMTs extensible


micheles:
Recently there were two request to introduce new IMTs in hazardlib:

1. new IMTs for geotechnical hazard
2. new IMTs for more sophisticated correlation models

Introducing a new IMTs is now simpler than in the past, but still not ideal. See for instance
https://github.com/gem/oq-engine/pull/3885; the user has to

1. modify the imt.py module
2. modify the oqvalidation.py module
3. modify the check_gsim.py module

Adding a new IMT requires forking the engine or opening a pull request to the main engine and wait months until it is reviewed. This is bad. We must make it possible *to add a new IMT without touching the engine code base at all*. While doing that one could take the occasion to fix early design errors.
In particular I have always regarded the fact that there is a new class for each IMT as a design error.
All IMTs should have the same class, they can by distinguished by their name (i.e. "PGA", "SA", etc)
and we just need a registry `IMT name -> factory function generating instances of that IMT`.
Then the user can add a new IMT in his own code. In the case of the IMTs introduced by Pablo he would need to write something like this

```python
from openquake.hazardlib import imt

imt.registry.add(
"FIV3", "Filtered Incremental Velocity (3)", "period")

imt.registry.add(
"SAAVG", "Average Spectral Acceleration", "period, damping", defaults=(5.,))

imt.registry.add(
"SDI", "Spectral inelastic displacement", "period, Cy, damping", defaults=(5.,))
```

and nothing else.

PS: an alternative, more backward compatible, would be to keep the IMTs as classes. Then Pablo
should write

```python
from openquake.hazardlib.imt import IMT, positivefloat

class FIV3(IMT):
    "Filtered Incremental Velocity (3)"
    _fields = [(period, positivefloat)]

class SAAVG(IMT):
     "Average Spectral Acceleration"
    _fields =  [("period", positivefloat), ("damping", positivefloat)]
    _defaults=(5.,)

class SDI(IMT):
    "Spectral inelastic displacement"
   _fields =  [("period", positivefloat),  ("Cy",positivefloat), ("damping", positivefloat)]
    _defaults=(5.,)
```

PR #3921, opened 2018-08-20
Making the IMTs extensible in user code

micheles:
Closes https://github.com/gem/oq-engine/issues/3919 while keeping compatibility with the past.


pheresi:
I've been working with other IMTs, and I've found that most of the code is coded in order to assume that any period-dependent IMT is SA. I think this should be changed in order to make the IMTs extensible. However, changing this at the present moment would break a lot of the code. 
One example: the coefficient tables for GMPEs assume that if the imt is a number, then it's the period of SA. To partially solve this problem in GMPEs of other period-dependent IMTs, I've found a somewhat messy solution: to create an auxiliar imt variable, being SA, and searching in the table with that auxiliar variable, like this:

PR #3923, opened 2018-08-20
Fix an error with the global mosaic tests

micheles:
The calculation number 12 hangs: https://gitlab.openquake.org/hazard/mosaic/mosaic/-/jobs/165.
The hanging is related to the logging; the calculation number 12 must log a warning from a worker in the process pool and it hangs there. The suboptimal solution is to replace the database logging with a simple warning on stderr, then the calculation runs. Further analysis is required, but it is left for the future.
While at it, I made some improvement in the function `parallel_pickle_source_models`.


daniviga:
I guess that this means that warnings aren't visible in the webui, right?

micheles:
Yes, this is one of the reasons why this approach is suboptimal.

Issue #3924, opened 2018-08-20
Make it possible to write plugins for the engine


micheles:
Until version 3.1, in order to add a new IMT or a new GSIM it was necessary to modify the source code of the engine. In the future this will not be necessary. An user will be able to write a plugin for the engine, i.e. a Python package with a format like the following:

```
openquake/myplugin/__init__.py
openquake/myplugin/imt.py  # optional file where to put additional IMTs
openquake/myplugin/correlation.py  # optional file where to put additional correlation models
openquake/myplugin/gsim1.py  # optional file where to put additional GSIMs
openquake/myplugin/gsim2.py  # optional file where to put additional GSIMs
...
```

The idea is that an user should be able to write his own plugin without having to touch the engine code base. Then, if he wants the plugin to be included in the official code base, he can just open a PR with everything.

PR #3925, opened 2018-08-20
Added a registry for the GSIM classes

micheles:
Second step towards https://github.com/gem/oq-engine/issues/3924.


micheles:
Uhm... we could replace `deprecated = True` with something like `superseded_by = MontalvaEtAl2017SInter`, is that what you have in mind? But is it a different pull request from this one, please open an issue if you want that functionality.

pheresi:
In PR #3926 I've included a new check_gsim that goes in line with this objective. Now it is able to read any IMT in the test tables. The idea is that now it takes the title of the column and uses from_string of imt.py. If the title is just a number, it assumes that it is the period of SA, and the IMT is SA.

PR #3926, opened 2018-08-20
Gsim for SDI

pheresi:
- New gsim file: openquake/hazardlib/gsim/heresi_2018.py.py
- New gsim test file: openquake/hazardlib/tests/gsim/heresi_2018_test.py.py
- Corresponding test tables
- Change in openquake/hazardlib/imt.py in order to include the new IM
- check_gsim.py can't read tables for this IM, we need to fix this file for new IMs. I have some ideas, but I'm not sure if it's the best option.

[EDIT]
- Change in check_gsim.py so now it can read any IMT from the test tables


openquake/hazardlib/tests/gsim/check_gsim.py
```diff
@@ -296,10 +297,9 @@ def _parse_csv_line(headers, values, req_site_params):
             value = float(value)
             if param == 'arias':  # ugly legacy corner case
                 param = 'ia'
-            imtclass = registry.get(param.upper(), None)
-            if imtclass:
-                imt = imtclass()
-            else:  # assume the IMT is a Spectral Acceleration
+            try:    # The title of the column should be IMT(args)
```
pheresi: With this we now can read any IMT in the test tables

micheles:
It is good that you include as many tests as necessary, but not more ;-)

openquake/hazardlib/tests/gsim/check_gsim.py
micheles: Good solution!

pheresi:
:)
I'll fix it then, hehe

openquake/hazardlib/gsim/heresi_2018.py
```diff
@@ -0,0 +1,1051 @@
...
+            4.20: self.Coeffs_4p20,
+            4.40: self.Coeffs_4p40,
+            4.60: self.Coeffs_4p60,
+            4.80: self.Coeffs_4p80,
+            5.00: self.Coeffs_5p00
+        }.get(imt_per, 0)
+
+        if CoeffsTable_thisT == 0:
+            print('Warning: Period not found')
+            mean = 0
```
g-weatherill: Is there any reason to raise a warning here rather than an error? These outputs will simply break things further downstream inside a PSHA calculation, probably misdirecting users to the cause of the error. If there are secondary tools that use this GMPE probably it would be safer to raise the error here and place the call to the GMPE inside a try-except block.

openquake/hazardlib/gsim/heresi_2018.py
```diff
@@ -0,0 +1,1051 @@
...
+
+        Cy_max = round(min(1.5, 1.5 * 0.6/imt_per), 4)
+        Cy_min = round(Cy_max / 15, 4)
+        if not Cy_min <= imt.Cy <= Cy_max:
+            raise ValueError('Cy should be between {0:.4f} and {1:.4f}.  \
+                           Used Cy = {2:.5f}'.format(Cy_min, Cy_max, imt.Cy))
+
+        imt_aux = imt_module.SA(damping=5, period=imt.Cy)
+
+        C = CoeffsTable_thisT[imt_aux]
```
g-weatherill: This approach seems quite messy and likely to break under many circumstances. Firstly, you are using floating point values as dictionary keys, so I wonder how often you will encounter cases where the corresponding coefficients are not found because of floating point inequalities. Secondly, is there any reason that intermediate periods cannot be supported by interpolating between the nearest two periods? If not then I would suggest to add a function to return the coefficients for the periods immediately above and below imt_aux and then apply log-T linear-coefficient interpolation for each coefficient by looping over the keys of the dictionary.

openquake/hazardlib/gsim/heresi_2018.py
pheresi: You're right. This is something that I did at the beginning and never fixed back to raise an error. I'll fix it

openquake/hazardlib/gsim/heresi_2018.py
pheresi: Yes. This GMPE is a little tricky because it depends on 2 inputs: T and Cy. I'll work on a more elegant solution for a 2D interpolation of the coefficients. This interpolation should, as you correctly point out, be able to interpolate coefficients for any T.

pheresi:
It will take me some time to fix the 2D interpolation. I'll create a separate PR for check_gsim and imt.

micheles:
After nearly 1.5 years nothing is moving with this PR so I am closing it.
Feel free to reopen it if you want to pursue this approach.

PR #3929, opened 2018-08-21
Optimized the parsing of the source model tree

micheles:
When a source model logic tree has attributes "applyToSources" on the branchsets it is necessary to read all of the source models to check that the source IDs in the logic tree match the source IDs in the models.
Unfortunately, at the moment this operation is performed always, even when there is no "applyToSources" in the logic tree. This has a negative performance effect on extra large source models (i.e. Trevor's model which has no "applyToSources").  This PR fixes the issue. On Trevor's model we can save ~15 minutes.

### DIFF
--- a/openquake/commonlib/logictree.py
+++ b/openquake/commonlib/logictree.py
@@ -28,6 +28,7 @@ import os
 import re
 import sys
 import copy
+import logging
 import itertools
 import collections
 import operator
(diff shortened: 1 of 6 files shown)


openquake/commonlib/logictree.py
```diff
@@ -558,20 +591,45 @@ def __init__(self, filename, validate=True, seed=0, num_samples=0,
         except AttributeError:
             raise ValidationError(
                 root, self.filename, "missing logicTree node")
-        self.apply_to_sources = set()
         self.parse_tree(tree, validate)
 
     def on_each_source(self):
         """
         :returns: True if the logic tree is defined on each source
```
mmpagani: I think this comment must be updated since now you return a set

micheles: Actually I am still returning a boolean here, `self.info.applytosources and self.info.applytosources == self.source_ids`

Issue #3918, opened 2018-08-17
Solve the pickling error/memory issue by making the distribution mechanism chunking the output


micheles:
Consider the following example

```python
def operation(items, monitor):
    total_result = []
    for item in items:
        partial_result = process(item)
        total_result.append(partial_result)
    return total_result
```

and suppose that each partial result takes 10 MB when pickled. Then if the number of items is ~400 the total amount of data to send back is around 4 GB, which is the pickle limit, and the calculation fails with a pickling error. If there are less items the calculation may still fail because we have only 2 GB per core and we can run out of memory even with 200 items.

This happens in event_based calculations, when generating ground motion fields. There is a solution that has been working for 6 years, just increase the number of concurrent_tasks so that there are less items and the calculation fits within the limits.

Still a cleaner solution would be to change the task function into a generator yielding the partial results:

```python
def operation(items, monitor):
    for item in items:
        yield process(item)
```

The distribution library would understand generators and would return back to the controller node only 10 MB at the time: we would never run out of memory and we would never run into the pickling error.

Changing the distribution library to work in this way is possible but nontrivial, so it was not done until now. But as the calculations becomes larger, we may want to reconsider the idea and spend the time to implement it.

micheles:
PS: the problem is already there, for instance for this calculation for Jordania by Desmond on cluster2: 
[jordania.zip](https://github.com/gem/oq-engine/files/2313980/jordania.zip)

PR #3930, opened 2018-08-22
Removed the ProcessPool and used subprocesses instead

micheles:
This really solves https://github.com/gem/oq-engine/pull/3923 and paves the way for https://github.com/gem/oq-engine/issues/3918. This proves times and again my old saying: *any problem can be solved by removing enough dependencies*. The threadpool distribution is not used by the engine, so it is not important right now and will be restored in the future.

NB: using subprocesses has a high startup time compared to multiprocessing that uses fork. This is not an issue in real calculations that take hours, but it is an issue in the tests that take seconds: they are getting a lot slower (spawning 100 subprocesses can take 2 seconds or more, forking 0.05 seconds or less) .

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,5 @@
   [Michele Simionato]
+  * Removed the multiprocessing.ProcessPool
   * Optimized the parsing of the logic tree when there is no "applyToSources"
   * Made the IMT classes extensible in client code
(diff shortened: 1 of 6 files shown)


daniviga:
Are we sure we want this in 3.2.0 and not keep it for 3.3.0? It looks to me too untested for 3.2.0

micheles:
I am closing this PR, I have a better solution in the works.

PR #3932, opened 2018-08-23
Replaced the processpool with the zmq pool

micheles:
Some of the differences are highlighted here.

1. before a ProcessPool was spawned at each Starmap call. A test could make several Starmap calls (i.e. to pickle the source models, to prefilter the sources, to generate the GMFs, to build the risk, etc)
and since there are 100+ tests hundreds of pools where created/destroyed. Now there is single zmq pool, so the tests are faster. I measured the time to run the risk tests on wilson and the improvement is from 314s to 120s i.e.nearly three times faster. The more the cores, the better the speedup.

2. for some reason, even big computations with a large data transfer gets a significant speedup by the new approach (say 10-15%); presumably, the zmq transfer is more efficient than the multiprocessing transfer.

3. On Windows the fork has a large performance penalty, so the new approach should be even better.

Running with the packages is hanging: https://ci.openquake.org/job/zdevel_oq-engine/2795
Running with celery is green: https://travis-ci.org/gem/oq-engine-celery
Running on macOS is green: https://ci.openquake.org/job/macos/job/zdevel_macos_engine/17/
There are strange errors here: https://ci.openquake.org/job/zdevel_oq-engine/2796/console


micheles:
Having replaced the forking processpool with spawn, now the fork bug is gone, so the zmq approach is less crucial. Also, it has a permission problem if the pool is spawned by the openquake user and the calculation is run by a different user. So I am closing this PR for the moment.

PR #3935, opened 2018-08-25
Skipped two correlation tests

micheles:
While those tests pass on my machine and also on Jenkins, they broke travis: https://travis-ci.org/gem/oq-engine/jobs/420402789. Probably one has to fix the random number generation. Not a problem since there are no calculators using the new HM2018ApplyCorrelation model.
@pheresi should have a look at the issue.


micheles:
Merging because master should not stay broken.

pheresi:
Oh, we had bad luck with the random numbers. I'll fix the seed and push it back

PR #3936, opened 2018-08-25
Back correlation tests

pheresi:
This PR fixes the correlation tests that were skipped in PR #3935.
Now Travis should be fine :)


micheles:
LGTM

PR #3947, opened 2018-08-29
Introduce a key `distribution.multi_node` in openquake.cfg

micheles:
I really need to distinguish the three cases `single_machine`, `cluster_with_shared_dir` and `cluster_without_shared_dir`, because the calculators will use different strategies depending on the available hardware.

### DIFF
--- a/doc/installing/cluster.md
+++ b/doc/installing/cluster.md
@@ -26,12 +26,15 @@ Note: you have to **restart every celery node** after a configuration change.
 
 ### Enable Celery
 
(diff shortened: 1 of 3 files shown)


daniviga:
At least, having it simplified to `multi_node = True` and then getting the shared dir status from its settings?

micheles:
This is possible.

daniviga:
I was wrong, no changes are required to the packages, but we/I need to update `cluster.md`.

PR #3954, opened 2018-08-31
Moved the rupture sampling in the prefiltering phase

micheles:
Essentially, I am going back to the algorithm used in engine 3.1 because otherwise the slow tasks will kill the performance. Here is a real case, a computation for Jordania by Desmond:
```
   in current master:
   Calculation 19056 finished correctly in 5856 seconds
   (py36) michele@wilson:~/jordania$ oq1 show task_info 19056
   ==================== ======= ======= ======= ======= =========
   operation-duration   mean    stddev  min     max     num_tasks
   pickle_source_models 285     223     127     443     2        
   prefilter            0.36820 0.31733 0.15391 4.17458 1,998    
   compute_hazard       54      197     0.23506 4,438   2,040    
   ==================== ======= ======= ======= ======= =========

   in this branch:
   Calculation 19066 finished correctly in 2367 seconds
   $ oq1 show task_info 19066
   ==================== ======= ====== ======= === =========
   operation-duration   mean    stddev min     max num_tasks
   pickle_source_models 285     226    125     445 2        
   prefilter            4.11750 13     0.01659 164 2,007    
   compute_gmfs         55      64     1.36520 704 1,873    
   ==================== ======= ====== ======= === =========
```
This branch is more than twice faster than master. The reason is that the slowest task of kind compute gmfs is taking only 704s instead of 4,438s (before compute_hazard was also producing the ruptures).


openquake/commonlib/source.py
```diff
@@ -861,9 +864,13 @@ def get_sources(self, kind='all'):
...
         sources = []
-        for src_group in self.src_groups:
-            if kind in ('all', src_group.src_interdep):
-                sources.extend(src_group)
+        for sm in self.source_models:
+            for src_group in sm.src_groups:
+                if kind in ('all', src_group.src_interdep):
+                    for src in src_group:
+                        if sm.samples > 1:
+                            src.samples = sm.samples
```
ptormene: Please check about this

PR #3963, opened 2018-09-03
Generate the statistical hazard maps together with the statistical hazard curves

micheles:
This allows to remove several lines of codes and it is faster than before.


openquake/calculators/event_based.py
```diff
@@ -459,7 +458,18 @@ def post_execute(self, result):
                     for sid in pmap:
                         arr[sid] = pmap[sid].array[:, 0]
                     self.datastore['hcurves/' + kind] = arr
-            self.save_hmaps()
+                    if oq.poes:
+                        P = len(oq.poes)
+                        I = len(oq.imtls)
+                        self.datastore.create_dset(
+                            'hmaps/' + kind, F32, (N, P * I))
```
ptormene: Please change all occurrences of `kind` into `statname`, when it's not a tuple

PR #3966, opened 2018-09-03
adding a GMPE modification for use in the USGS Hawaii hazard model

kejohnso:
Added a new class MunsonThurber1997Hawaii, which modifies some coefficients, magnitude dependent, and adds the calculation of SA 0.1 and 0.2 s. This is necessary to implement the USGS Hawaii hazard model.


micheles:
As always, tests are needed.

kejohnso:
Okay - tests have been added

PR #3972, opened 2018-09-05
Doing the prefilter with rtree+processpool, the preprocess with numpy+celery

micheles:
This is the faster combination on our cluster. I have also change the logging so that the estimated max_gmf_size is always logged.


micheles:
LGTM, I want this code to run on Jenkins tonight to make the release tomorrow.

PR #3976, opened 2018-09-06
adding Campbell1997 and version of MunsonThurber1997

kejohnso:
adding GMPEs necessary to implement USGS Hawaii 1998


openquake/hazardlib/gsim/campbell_1997.py
```diff
@@ -0,0 +1,152 @@
...
+        return f
+
+    def get_Ssr_term(self, vs30):
+        """
+        Returns site term for soft rock (pg 157)
+        """
+        ssr = []
+        for v in vs30:
+            if ((v >= 760.) & (v < 1500.)):
+                ss = 1.
```
micheles: You have two extra parenthesis,   `if (v >= 760.) & (v < 1500.)` is fine.

openquake/hazardlib/gsim/campbell_1997.py
```diff
@@ -0,0 +1,152 @@
...
+        Returns site term for hard rock (pg 157)
+        """
+        shr = []
+        for v in vs30:
+            if v >= 1500.:
+                sh = 1
+            else:
+                sh = 0
+            shr.append(sh)
+        return shr
```
micheles: Assuming vs30 is a numpy array, you can do that with a single line, `return vs30 > 1500.`

openquake/hazardlib/gsim/munson_thurber_1997.py
```diff
@@ -98,6 +99,97 @@ def get_mean_and_stddevs(self, sites, rup, dists, imt, stddev_types):
...
+
+        # Site term only distinguishes between lava and ash;
+        # since ash sites have Vs30 in the range 60-200m/s,
+        # we use this upper value as class separator
+        S = np.zeros(R.shape)
+        S[sites.vs30 <= 200] = 1
+
+        # Mean ground motion (log10)
+        if rup.mag <= 7.:
+            mean = (0.518 + 0.387*M - np.log10(R) - 0.00256*R + 0.335*S)
```
micheles: The parenthesis here are not needed.

kejohnso: Okay I removed these but now cannot get the indentation to pass pep-8

openquake/hazardlib/gsim/campbell_1997.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+        Returns site term for soft rock (pg 157)
+        """
+        ssr = []
+        for v in vs30:
+            if (v >= 760.) & (v < 1500.):
+                ss = 1.
+            else:
+                ss = 0.
+            ssr.append(ss)
```
micheles: Even this can be replaced with a single line

```python
return (vs30 >= 760) & (vs30 < 1500)
```

openquake/hazardlib/gsim/campbell_1997.py
```diff
@@ -0,0 +1,145 @@
...
+    REQUIRES_DISTANCES = set(('rrup', ))
+
+    #: Verification of the mean value was done by digitizing Figs. 9 and 10
+    #: using Engauge Digitizer. The tests check varied magnitude, distance,
+    #: vs30, and faulting type. Maximum error was ~1.3%. OpenQuake trellis
+    #: plots match these figures. Also tested against a matlab implementation
+    #: (web.stanford.edu/~bakerjw/GMPEs/C_1997_horiz.m), which also has no
+    #: verification tables.
+
+    non_verified = True
```
g-weatherill: From the description above I would say this has been reasonably verified, so I don't think the non_verified assignment is warranted. Getting source code and/or tables for legacy GMPEs of this age is pretty difficult, and the fact that you tested against Baker's matlab implementation I would say is sufficient for verification.

openquake/hazardlib/gsim/campbell_1997.py
```diff
@@ -0,0 +1,145 @@
...
+        rake = rup.rake
+        F = self.get_fault_term(rake)
+
+        # compute mean
+        mean = -3.512 + (0.904 * M) - (1.328 * np.log(np.sqrt(R**2
+               + (0.149 * np.exp(0.647 * M))**2))) \
+               + (1.125 - 0.112 * np.log(R) - 0.0957 * M) * F \
+               + (0.440 - 0.171 * np.log(R)) * Ssr \
+               + (0.405 - 0.222 * np.log(R)) * Shr
+        stddevs = self.get_stddevs(mean)
```
g-weatherill: I would make two notes here for general implementation of GMPEs, that are not necessarily fundamental to this one (as this is legacy rather than current best practice) but can help significantly in keeping the GMPE library consistent (which is better for the GMPE toolkit or secondary uses of the library):

1. It is more useful to split the GMPE into separate functions for the different elements of the model (i.e. magnitude scaling, geometric spreading, anelastic attenuation, depth/hanging wall/style-of-faulting adjustments and site amplification) rather than execute the entire function in one continuous line. This makes it easier to spawn modified/adjusted GMPEs in subclasses that adapt certain parts of the model (e.g. changing the anelastic term for application to a different region). This is becoming more important to keep doing now than ever before, as it is more common in best practice PSHA to see "backbone" GMPEs adjusted for local conditions and applications than it is to see new GMPEs for different regions/applications.

2. Even though this GMPE is only for PGA I would maintain it is still better to use a coefficients table and place the coefficients there rather than coding them into the call to the main function itself. The coefficients table is another way of helping users make modifications, check things like spectral ranges (though that's easily side-stepped here), and make on-the-fly adjustments to the models (yes, that's possible!) without having to re-write too many functions.

openquake/hazardlib/gsim/campbell_1997.py
```diff
@@ -0,0 +1,136 @@
...
+        const.StdDev.TOTAL,
+    ])
+
+    #: Requires vs30
+    REQUIRES_SITES_PARAMETERS = set(('vs30',))
+
+    #: Required rupture parameters are magnitude and top of rupture depth
+    REQUIRES_RUPTURE_PARAMETERS = set(('mag', 'rake'))
+
+    #: Required distance measure is closest distance to rupture
```
rcgee: After quickly scanning the publication, I see they are actually using Rseis (distance to seismogenic rupture ). Might be worth mentioning that we using Rrup instead, which should be very similar to Rseis. Otherwise LGTM

openquake/hazardlib/gsim/campbell_1997.py
kejohnso: Okay, thanks, I removed this

openquake/hazardlib/gsim/campbell_1997.py
```diff
@@ -0,0 +1,145 @@
...
+        for m in mean:
+            if np.exp(m) < 0.068:
+                sig = 0.55
+            elif (np.exp(m) >= 0.068) & (np.exp(m) <= 0.21):
+                sig = 0.173 - 0.140 * np.log(np.exp(m))
+            else:
+                sig = 0.39
+            sig_n[i] = (sig)
+            i += 1
+        stddevs.append(sig_n)
```
g-weatherill: Try this:
```python
def get_stddevs(self, mean, stddev_types):
    """
    Returns the ...
    """
    mean = np.exp(mean)
    sigma = 0.39 + np.zeros(mean.shape)
    sigma[mean < 0.068] = 0.55
    idx = np.logical_and(mean >= 0.068, mean <= 0.21)
    sigma[idx] = 0.173- 0.140 * np.log(mean[idx])
    stddevs = []
    for stddev in stddev_types:
        if stddev == const.StdDev.Total:
            stddevs.append(sigma)
    return stddevs
```

Then change invoking call accordinly.

kejohnso: Okay, I changed to this

micheles:
LGTM

PR #3928, opened 2018-08-21
Added a new experimental distribution mode `subprocess`

micheles:
This solves the problem of the logging of yesterday https://github.com/gem/oq-engine/pull/3923.

@daniviga, can you run the global mosaic test with this branch?

PR #3978, opened 2018-09-07
Solved the fork bug

micheles:
The so called "fork bug" is insolvable because it is a *feature* of POSIX (see https://codewithoutrules.com/2018/09/04/python-multiprocessing/). To us the bug caused infinite pain:
it was the cause of the KeyError on macOS (and sometimes on Linux) on the day before yesterday, the reason why if you do not close the datastore before forking hp5y sometimes reads bogus numbers and sometimes segfaults, and also the reason why the global mosaic calculation for Mexico was hanging during logging.
However in Python 3 there is a builtin way to avoid forking that I failed to recognize before: `multiprocessing.get_context('spawn')`. With that all my work to remove the fork becomes useless (see https://github.com/gem/oq-engine/pull/3928).

The price to pay is a longer startup time, which is not acceptable in the tests (they would became too slow for Travis) unless with a smart trick one disable `Starmap.shutdown` in the tests, so that there is a single pool for all tests.

### DIFF
--- a/openquake/baselib/parallel.py
+++ b/openquake/baselib/parallel.py
@@ -553,7 +553,7 @@ class Starmap(object):
             cls.dask_client = Client()
 
     @classmethod
-    def shutdown(cls, poolsize=None):
+    def shutdown(cls):
         if hasattr(cls, 'pool'):
             cls.pool.close()
             cls.pool.terminate()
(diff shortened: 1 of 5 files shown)


daniviga:
Win: https://ci.openquake.org/job/windows/job/zdevel_oq-engine/34/
macOS: https://ci.openquake.org/job/macos/job/zdevel_macos_engine/21/
Ubuntu: https://ci.openquake.org/job/zdevel_oq-engine/2807/

PR #3979, opened 2018-09-07
Bugfix on hmtk plotting completeness adjusted table and more catalogue special methods

g-weatherill:
Two changes:
1. The `get_completeness_adjusted_table` function in `hmtk.plotting.seismicity` module is cleaned up to verify that rates are consistent with standard recurrence table under the single completeness case.
2. Adds some more special methods to the `hmtk.seismicity.catalogue.Catalogue` class (`__str__`, `__getitem__`, `__iter__` and `__len__`)


micheles:
Please add a line to the changelog, for the rest LGTM

PR #3980, opened 2018-09-10
changing MunsonThurber1997Hawaii from class to subclass

kejohnso:


klunk386:
I would further simplify the get_mean_and_stddev method by calling the corresponding (original) method from the base class while passing a magnitude modified according to needs. This can be done using:
`mean, stddevs = super().get_mean_and_stddevs(sites, rup, dists, imt, stddev_types)`
The advantage is that if you change the base method, you don't need to manually modify the derived one as well.

klunk386:
Move all code related to calculation of the stddev after the part related to the calculation of the mean, to avoid logical jumps in the code flow. By the way, if you implement the modification above, you do not need to recompute the sttdev anyway.

klunk386:
I am quite sure that all those log/exp conversions between lines 146-152 can be further simplified

PR #3990, opened 2018-09-12
Avoided a deepcopy of the source models

micheles:
This will make `oq info --report` and the global mosaic tests a bit faster. For instance, for the case of Trevor Allen's model, the runtime of `oq info --report` goes down from 2325s to 1828s, which is noticeable.

### DIFF
--- a/openquake/commonlib/readinput.py
+++ b/openquake/commonlib/readinput.py
@@ -489,8 +489,7 @@ class SourceModelFactory(object):
         """
         check_nonparametric_sources(fname, sm, investigation_time)
         if apply_uncertainties:
-            sm = copy.deepcopy(sm)
-            for group in sm:
+            for group in copy.deepcopy(sm):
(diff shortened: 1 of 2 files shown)


openquake/commonlib/readinput.py
```diff
@@ -485,14 +485,15 @@ def __call__(self, fname, sm, apply_uncertainties, investigation_time):
...
+            a copy of the original source model with changed sources (if any)
+            or the original model with unchanged sources
         """
-        sm = copy.deepcopy(sm)
         check_nonparametric_sources(fname, sm, investigation_time)
-        for group in sm:
-            for src in group:
-                changed = apply_uncertainties(src)
-                if changed:
+        if apply_uncertainties:
```
daniviga: Not sure what are you checking here. Does it means that the argument that is passed to the method can be 'False' ? If yes please update the docstring.

PR #3992, opened 2018-09-12
Send the sources while splitting

micheles:
Currently the engine *first* splits all the sources and *then* send them to the workers; with this PR it sends the sources while it is splitting them, so the workers start working immediately.

PS: this PR is no good, for Trevor's model it takes more time (2346s vs 1857s), twice the memory and 106,674 tasks are generated.


micheles:
The right way is to split the sources in parallel, as done in https://github.com/gem/oq-engine/pull/4013

PR #3999, opened 2018-09-14
Saved the source information inside `source_info` only

micheles:
To avoid redundancy/inconsistencies. Also, I moved the source splitting inside  `get_composite_source_model`. Several lines of code were removed.


openquake/calculators/event_based.py
```diff
@@ -46,8 +45,8 @@
 
 def weight(src):
     # heuristic weight
-    return len(src.eb_ruptures) * src.ndists  # this is the best
-    # return sum(ebr.multiplicity for ebr in src.eb_ruptures) * src.ndists
+    return len(src.eb_ruptures)  # this is the best
+    # return sum(ebr.multiplicity for ebr in src.eb_ruptures)
```
ptormene: Is this commented line still needed?

PR #4005, opened 2018-09-17
Remove supervisor in Ubuntu packages

### LINKED ISSUE #2030: Drop supervisord and use only native inits
Already done for RHEL + `systemd`, the same scripts can be used with Ubuntu 16.04 and greater.

~~Provide `upstart` scripts for Ubuntu 14.04 and eventually for 12.04.~~

daniviga:
And switch to systemd, like we do in RHEL.
This is now possible because all supported distros have systemd.

Closes #2030 

Tests for both Bionic and Xenial are green: https://ci.openquake.org/job/zdevel_oq-engine/2810/


packager.sh
```diff
@@ -601,9 +599,8 @@ celery_wait() {
     return 1
 }
 
-sleep 30
-sudo supervisorctl status
-sudo supervisorctl start openquake-celery
+sudo systemctl status openquake\\*
```
nastasi-oq: ``openquake-\\*`` seems to me more strict and self-explaining about what kind of services you could find related to openquake.

PR #4012, opened 2018-09-19
20180918 silva2002

mmpagani:
This PR adds two versions of the Silva et al. (2002) GMPE as described in the report titled "DEVELOPMENT OF REGIONAL HARD ROCK ATTENUATION RELATIONS FOR CENTRAL AND EASTERN NORTH AMERICA", available at http://www.pacificengineering.org/.

We computed the verification tables for the mean values using a version of these GMPE in the haz43 code. We do not test sigma since the values in the independent version do not correspond to the original ones.

These GMPEs are required for the calculation of hazard with the Canada 2015 model.


micheles:
What about the documentation files? I would have expected some changes in the .rst files.

PR #4013, opened 2018-09-19
Parallelized split_filter

micheles:
With this trick the Jordania calculation by Desmond becomes twice as fast in the rupture generation phase, from 1700s to 900s. `oq info -r` for Trevor's calculations goes down from 50m to 25 minutes.
Here is a run of the mosaic tests: https://gitlab.openquake.org/hazard/mosaic/mosaic/-/jobs/1381


openquake/commonlib/readinput.py
```diff
@@ -710,82 +713,122 @@ def get_composite_source_model(oqparam, monitor=None, in_memory=True,
...
-        source_info['num_split'] = U32(num_split)
+    :returns: a new :class:`openquake.commonlib.source.CompositeSourceModel`
+    """
+    mon = monitor('split_filter')
+    sample_factor = float(os.environ.get('OQ_SAMPLE_SOURCES', 0))
+    logging.info('Splitting/filtering sources')
+    smap = parallel.Starmap.apply(
+        split_filter,
+        (csm.get_sources(), srcfilter, min_mag, seed, sample_factor, mon),
+        maxweight=10000,
```
ptormene: Please add a comment about this number

PR #4017, opened 2018-09-24
Yielding ruptures

micheles:
Our users keep hitting the 4 GB pickling limit when computing the ruptures. Now that we have https://github.com/gem/oq-engine/issues/3918 the issue can be easily solved. While at it, I made several refactorings.

1. removed the special case for task names starting with '_' (there are no such tasks anymore)
2. added the time to receive back the outputs in the log message
3. moved the build_ruptures routine from get_composite_source_model to the event base calculator
4. avoided redundant calculations when ground_motion_fields = false
5. removed some dead code


openquake/calculators/event_based.py
```diff
@@ -242,26 +265,40 @@ def zerodict(self):
         self.grp_trt = self.csm_info.grp_by("trt")
         return zd
 
-    def process_csm(self):  # called after split_all
+    def build_ruptures(self):
         """
         Prefilter the composite source model and store the source_info
         """
+        param = {'RUPTURES_PER_BLOCK': RUPTURES_PER_BLOCK}
```
ptormene: It's not obvious to me, why are we using an all-capitalized format for this key

PR #4018, opened 2018-09-24
20180921 fix disaggr

mmpagani:
This PR fixes a bug in the disaggregation calculation which was causing a wrong binning of magnitudes. The bug was caused by an improper handling of numerical approximation.


micheles:
Apart from the minor comment, LGTM.

PR #4023, opened 2018-09-25
Improved the point->multipoint converter

micheles:
The global mosaic models with lots of point sources (typically for distributed seismicity) should be converted to multiPoint sources for efficiency, both in speed and space.


openquake/hazardlib/tests/sourceconverter_test.py
```diff
@@ -85,14 +85,89 @@
...
+    def test_complex(self):
+        testfile = os.path.normpath(os.path.join(
+            testdir, '../../../qa_tests_data/classical/case_30/ssm/shallow/'
+            'gridded_seismicity_source_4.xml'))
+        sm = nrml.read(testfile).sourceModel
+        update_source_model(sm, testfile)
+        with io.BytesIO() as f:
+            nrml.write(sm, f)
+            got = f.getvalue().decode('utf-8')
+            print(got)
```
ptormene: Should we remove this print?

micheles: No, it is there for debugging purpose. Notice that since the tests are run with nose, print are hidden unless the `-s` option is given, which sometimes I do since that print is useful.

Issue #2999, opened 2017-09-06
The memory monitoring is not working


micheles:
Since a long time ago, we are losing the log messages when a computation is running out of memory. We should add a logserver_url to the `Monitor` and call `db.actions.log(db, job_id, timestamp, level, process, message)` in `Monitor.send`.

micheles:
I have added a log message https://github.com/gem/oq-engine/commit/d64d3371f43e598d1322a53ebfbf8a5acd9a7e56 but it only logs in the celery logs, not in the main console.

daniviga:
At least with #3005 I've been able to get the warning logs in the main console also using celery. Should we consider this closed?

micheles:
No, it is not closed. The goal is to get in the console log also the warnings coming from remote machines going out of memory, not only the warning coming from the master.

daniviga:
That's right. I was able to see warnings because I was running it on a single machine and those were actually warnings coming from the master process.

PR #4024, opened 2018-09-25
Added a call to `check_mem_usage` in the workers

micheles:
Now that the forking bug has been solved we can finally close https://github.com/gem/oq-engine/issues/2999. By playing with `ulimit` one can reduce the memory available and induce warnings like the following:

```
[2018-09-25 15:35:12.00 #18946 WARNING] Using over 82% of the memory in thinkpad!
[2018-09-25 15:35:12.63 #18946 WARNING] Using over 83% of the memory in thinkpad!
```

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,6 @@
   [Michele Simionato]
+  * If a worker runs close to out of memory, now a warning appears in the
+    main log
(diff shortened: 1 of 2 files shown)


micheles:
Bad idea, since it is a WIP and I have decided that tomorrow I will change completely approach ;-)

PR #4025, opened 2018-09-25
Remove a piece of code that is preventing nohup from working

daniviga:
When running on a cluster we had to ignore a signal, in a certain condition. This seems to have a side effect of making `nohup` working badly.

I'm disabling that code for now. Further investigation is required, also to see if the new `spawn` mechanism works properly when calculation is orphan.

Before the change:
```bash
$ grep SigIgn /proc/$!/status
SigIgn:	0000000001001000
```

After the change
```bash
$ grep SigIgn /proc/$!/status
SigIgn:	0000000001001001
```

Explanation is here: https://unix.stackexchange.com/a/420663

This makes anyway `nohup` only working after the `splitting/filtering source` phase because if session is disconnect before that the computations becomes stuck.

### DIFF
--- a/openquake/engine/engine.py
+++ b/openquake/engine/engine.py
@@ -186,8 +186,14 @@ def raiseMasterKilled(signum, _stack):
     :param _stack: the current frame object, ignored
     """
     # Disable further CTRL-C to allow tasks revocation when Celery is used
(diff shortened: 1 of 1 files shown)


daniviga:
After further investigations I'm closing this since the picture is too far complex.

daniviga:
Replaced by #4026

PR #4030, opened 2018-09-26
Removed obsolete code used only in the tests

micheles:
Let's read the CSV hazard curves from the official export format, not from the non-official format used years and years ago, before there was an official CSV format for the hazard curves. More than 300 lines have been removed.

PS: while at it, I fixed a bug when import a curve with PoE=1, thus producing NaNs in the damage outputs.


micheles:
Left for a post-mortem review.

PR #4033, opened 2018-09-27
Optimized event based hazard task distribution

micheles:
The trick is to start computing the GMFs as soon as the first block of ruptures is available and not after the last block has arrived. This makes a significant difference when there are slow tasks. For instance for a calculation for Colombia by Catarina the total runtime goes down from 2268 seconds to 1863 seconds (18% better). Also, less memory is needed in the controller node (15% less in that example) and 11 lines of code were removed ;-)


ptormene:
Liar! +69 −72 = (only) -3 lines of code... not -11!

Issue #4034, opened 2018-09-27
4GB limit in UCERF event based


micheles:
This happened to @raoanirudh:
```python
OverflowError: 
  File "/opt/openquake/oq-engine/openquake/baselib/parallel.py", line 388, in safely_call
    zsocket.send(res)
  File "/opt/openquake/oq-engine/openquake/baselib/zeromq.py", line 148, in send
    self.zsocket.send_pyobj(obj)
  File "/opt/openquake/lib/python3.6/site-packages/zmq/sugar/socket.py", line 601, in send_pyobj
    msg = pickle.dumps(obj, protocol)
OverflowError: cannot serialize a bytes object larger than 4 GiB
````
The files are on wilson:  /home/oqdata_shared/arao/projects/usa/calculations/california/statewide/calc_gmfs_nshmp2014_ucerf3.ini

micheles:
The trick here is to raise the number of SES so that the job parallelize better. With a single SES you will get a huge task and you will run into the limit.

micheles:
Actually the error is also on the risk side, even after building 8033 inputs, because California has 3,233,384 assets, so the calculation must be reduced.

PR #4040, opened 2018-09-28
Fighting against the 4GB limit in event based risk

micheles:
We can just yield a result for each riskinput, i.e. one result for each hazard site.
Solves https://github.com/gem/oq-engine/issues/4034 for the moment.

### DIFF
--- a/openquake/calculators/event_based_risk.py
+++ b/openquake/calculators/event_based_risk.py
@@ -68,7 +68,6 @@ def event_based_risk(riskinputs, riskmodel, param, monitor):
     :returns:
         a dictionary of numpy arrays of shape (L, R)
     """
-    results = []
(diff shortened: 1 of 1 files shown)


openquake/calculators/event_based_risk.py
```diff
@@ -299,13 +297,12 @@ def _save_maps(self, dic, aids):
                 for aid, arr in zip(aids, loss_maps):
                     self.datastore[key][aid] = arr
 
-    def combine(self, dummy, results):
+    def combine(self, dummy, res):
         """
         :param dummy: unused parameter
         :param res: a list of result dictionaries
```
ptormene: If this is now a single item instead of a list, we should also update the docstring accordingly

PR #4041, opened 2018-09-29
Using Python 3.6 __init_subclasses__ hook

micheles:
Thanks to this feature of Python 3.6 I was able to remove the metaclass MetaGSIM. @mmpagani will be happy because now the regular instantiation syntax `GSIM(value)` works again (before the metaclass magic forced the user to use the syntax `GSIM(name=value)`.

@daniviga after this PR the engine will stop working with Python 3.5.


micheles:
Yes, but let me do that in a separate PR.

Issue #3748, opened 2018-05-31
More work on the Aristotle project


micheles:
- [x] make it possible to import vulnerability functions and fragility functions together in scenario_hazard
- [ ] add an example for scenario_damage with noDamageLimit = 0 and != 0
- [x] make it possible to disable spatial correlation on demand
- [x] fix the download of uncertainty.xml.zip when the archive has an unexpected but usable format
- [ ] add a test for spatial_correlation=full
- [x] investigate the interpolation_error discovered by Petros
- [ ] add a test for cutoffed GMFs (i.e. for shakemap_id = usp0008ssb)
- [ ] add a way to disable site amplification
- [ ] add an output for the total damage distributions, similar to agglosses

micheles:
Let's close this since we have #9227

PR #4043, opened 2018-10-01
Renamed the composite_risk_model dataset into fragility or vulnerability

micheles:
This is preliminary to be able to store fragility and vulnerability functions at the same time, see https://github.com/gem/oq-engine/issues/3748

### DIFF
--- a/openquake/calculators/extract.py
+++ b/openquake/calculators/extract.py
@@ -480,8 +480,9 @@ def build_damage_dt(dstore, mean_std=True):
        a composite dtype loss_type -> (mean_ds1, stdv_ds1, ...) or
        loss_type -> (ds1, ds2, ...) depending on the flag mean_std
     """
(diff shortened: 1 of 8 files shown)


openquake/calculators/extract.py
```diff
@@ -552,3 +551,12 @@ def extract_mean_std_curves(dstore, what):
     for imt in getter.imtls:
         yield 'imls/' + imt, getter.imtls[imt]
         yield 'poes/' + imt, arr[:, getter.imtls(imt)]
+
+
+@extract.add('composite_risk_model.attrs')
+def crm_attrs(dstore, what):
+    """
+    Yield imls/IMT and poes/IMT containg mean and stddev for all sites
```
ptormene: Please fix the docstring

micheles: Done

PR #4050, opened 2018-10-04
20181004 ceus

mmpagani:
This PR adds: 
1. The parameters required by the NRCan15SiteTerm modified GMPE, namely the reference soil condition. 
2. A modified version of the Atkinson and Macias (2009) GMPE for subduction interface earthquakes. This GMPE is modified following what proposed in Atkinson and Adams (2013).


mmpagani:
@micheles I think I covered all the notes. I also added a test to make sure an exception is raised.

raoanirudh:
@mmpagani, the reference soil condition needs to be specified also for the ToroEtAl2002 GMPE

mmpagani:
@raoanirudh where are you using ToroEtAl02 in the LT I sent you?

PR #4053, opened 2018-10-05
Reduce the risk model to the taxonomies present in the exposure

micheles:
In the case of British Columbia the model is reduced from 7260 to 104 taxonomies. This has an huge benefit on the data transfer that goes down from 60 GB to 1.3 GB (46x improvement). The workers are no  more waiting from the data to arrive.


micheles:
Assigned as post-mortem review

PR #4059, opened 2018-10-10
Reduced the site collection instead of just filtering it

micheles:
Things become simpler and there is no need for the LargeExposureGrid check; also disk space and memory are saved. The numbers may change slightly in the case of GMF-correlation because of the
line
```python
            stddev_intra = numpy.matlib.repmat(
                stddev_intra, len(sites.complete), 1)
```
and now the complete site collection can be shorter than before.This is not a problem because we never use correlation in the global models, and the GMF computation is tricky anyway (even without this change removing a site is changing the GMFs on all other sites).

PS: this PR also fixes a bug in event based.  In this case the source filter is defined as

```python
SourceFilter(self.sitecol.complete, oq.maximum_distance, self.hdf5cache)
```
so in master it is NOT filtering sources that should be filtered. This explain why event_based/case_16 produces less ruptures than before and why I had to increase the maximum_distance to 300 km to avoid filtering out everything.


micheles:
Post-mortem review

PR #4060, opened 2018-10-10
More work on the command `oq prepare_site_model`

micheles:
I have added documentation and tests. Short story:
```bash
$ oq prepare_site_model Exposure/Exposure_Res_Ecuador.csv,Exposure/Exposure_Res_Bolivia.csv Vs30/Ecuador.csv,Vs30/Bolivia.csv --grid-spacing=10
```
will produce a sites.csv file with the right vs30 values.


micheles:
Post mortem review

doc/oq-commands.md
```diff
@@ -182,3 +182,58 @@ Be warned that for large calculations the extraction will likely be slower
...
+```
+
+The command work in two modes: with non-gridded exposures (the
+default) and with gridded exposures. In the first case the assets are
+aggregated in unique locations and for each location the vs30 coming
+from the closest vs30 record is taken. In the second case, i.e. when a
+grid spacing parameter is passed, a grid containing of all the
+exposure is built and the points with assets are associated to the
+vs30 records. The maximum association distance is given
+by the grid spacing multiply by the square root of 2, so by enlarging
```
ptormene: multiply -> multiplied

doc/oq-commands.md
```diff
@@ -182,3 +182,58 @@ Be warned that for large calculations the extraction will likely be slower
...
+the grid spacing you can avoid discarding the assets.
+
+In large risk calculation one wants to *use the gridded mode always* because:
```
ptormene: calculation -> calculations

openquake/commands/prepare_site_model.py
```diff
@@ -32,52 +32,65 @@
 
 
 F32 = numpy.float32
+SQRT2 = 1.414
```
ptormene: Is this approximation ok?

micheles:
Yes, more than enough.

doc/oq-commands.md
micheles: Done

Issue #4062, opened 2018-10-10
Add to the rupture a reference to the source generating it


micheles:
We can just add the source index in the `source_info` dataset. This was requested by M. Pagani and also by our users.

PR #4065, opened 2018-10-11
Stored in the rupture the source index

micheles:
Closes https://github.com/gem/oq-engine/issues/4062.


micheles:
@CatalinaYepes does not want to exposure the `srcidx` in the CSV exports and I agree with her, so I am removing it.

PR #4069, opened 2018-10-12
Do not discard assets unless `discard_assets=true` is set

micheles:
The engine was discarding assets with a warning, now it raises an error instead, unless the error is silenced. Notice that in scenario calculations, since there is a `maximum_distance`, it is normal that some assets are not affected by the rupture, so we should keep the current behavior there (just a warning). On the other hand, if the scenario comes from imported GMFs, there should be an error, because it is easy to make mistake in that case and they must be very visible.


micheles:
Post mortem review

PR #4076, opened 2018-10-15
Enhanced prepare_site_model to compute also z1pt0 and z2pt5

micheles:
And enhanced the engine to read site model files in CSV format. Here is an example of usage:
```bash
   $ oq prepare_site_model Exposure/Exposure_South_America.xml Vs30/Peru.csv Vs30/Chile.csv Vs30/Ecuador.csv Vs30/Argentina.csv Vs30/Venezuela.csv Vs30/Colombia.csv Vs30/Bolivia.csv --grid-spacing 10
```


openquake/commands/prepare_site_model.py
```diff
@@ -29,12 +29,48 @@
...
+req_site_params = {'vs30', 'z1pt0', 'z2pt5'}
+
+
+def calculate_z1pt0(vs30):
+    '''
+    Reads an array of vs30 values (in m/s) and
+    returns the depth to the 1.0 km/s velocity horizon (in m)
+
+    :param vs30: the shear wave velocity (in m/s) at a depth of 30m
+    '''
```
g-weatherill: Is there a literature source for this conversion model between Vs30 and Z1.0. It looks like the Abrahamson & Silva (2008) model, so this should be clarified in the doc string.

raoanirudh:
References for the 2014 NGA-West2 regression equations are listed below.

---
#### Z1.0–Vs30 Relationship
**Chiou, B.S.-J. & Youngs, R.R., 2014**. Update of the Chiou and Youngs NGA model for the average horizontal component of peak ground motion and response spectra. _Earthquake Spectra_, 30(3), pp.1117–1153.

---
#### Z2.5–Vs30 Relationship
z1pt0

**Campbell, K.W. & Bozorgnia, Y., 2014**. NGA-West2 ground motion model for the average horizontal components of PGA, PGV, and 5% damped linear acceleration response spectra. _Earthquake Spectra_, 30(3), pp.1087–1114.

vs30-z2pt5

---
#### Z2.5–Z1.0 Relationship
**Campbell, K.W. & Bozorgnia, Y., 2013.** NGA-West2 Campbell-Bozorgnia Ground Motion Model for the Horizontal Components of PGA, PGV, Response Spectra for Periods Ranging from 0.01 to 10 sec, Berkeley, CA.

z1pt0-z2pt5

---

raoanirudh:
The above regression equations were developed for California and Japan, so caution must be exercised before applying them to other regions.

Also, a quick check of the GMPEs requiring either z1.0 or z2.5 shows that many of them have the equations included in their implementation. Eg.:

https://github.com/gem/oq-engine/blob/engine-3.2/openquake/hazardlib/gsim/abrahamson_2014.py#L295-L301
```python
def _get_z1pt0ref(self, vs30):
        """
        This computes the reference depth to the 1.0 km/s interface using
        equation 18 at page 1042 of Abrahamson et al. (2014)
        """
        return (1. / 1000.) * np.exp((-7.67 / 4.)*np.log((vs30**4 + 610.**4) /
                                                         (1360.**4 + 610.**4)))
```

https://github.com/gem/oq-engine/blob/engine-3.2/openquake/hazardlib/gsim/chiou_youngs_2014.py#L224-L236
```python
def _get_centered_z1pt0(self, sites):
        """
        Get z1pt0 centered on the Vs30- dependent avarage z1pt0(m)
        California and non-Japan regions
        """
        #: California and non-Japan regions

        mean_z1pt0 = (-7.15 / 4.) * np.log(((sites.vs30) ** 4. + 570.94 ** 4.)
                                           / (1360 ** 4. + 570.94 ** 4.))
        centered_z1pt0 = sites.z1pt0 - np.exp(mean_z1pt0)

        return centered_z1pt0
```

https://github.com/gem/oq-engine/blob/engine-3.2/openquake/hazardlib/gsim/campbell_bozorgnia_2014.py#L308-L319
```python
def _select_basin_model(self, vs30):
        """
        Select the preferred basin model (California or Japan) to scale
        basin depth with respect to Vs30
        """
        if self.CONSTS["SJ"]:
            # Japan Basin Model - Equation 34 of Campbell & Bozorgnia (2014)
            return np.exp(5.359 - 1.102 * np.log(vs30))
        else:
            # California Basin Model - Equation 33 of
            # Campbell & Bozorgnia (2014)
            return np.exp(7.089 - 1.144 * np.log(vs30))
```

Thus, if the user does not have better estimates for the z1.0 or z2.5 terms when using one of these GMPEs, these equations that are already implemented in the engine could be used for estimating the missing parameters from the provided vs30 values.

mmpagani:
As pointed out some GMPEs have also equations to infer basin parameters from vs30 (e.g. Chiou and Youngs 2014; Abrahamson et al., 2014). It would be probably better to avoid redundancies. If I'm not mistaken, similar conversion equations are in the gmtk under smtk.trellis.configure. https://github.com/GEMScienceTools/gmpe-smtk/blob/master/smtk/trellis/configure.py#L193

Now we are under the gun and I don't think we have time to make a proper cleanup but I think this something worth doing in 2019.

openquake/commands/prepare_site_model.py
micheles: Changed to use the Chiou & Youngs (2014)

CatalinaYepes:
**The `parse_site_model` command:**

This command facilitates the preparation of "efficient" site models to be used for risk calculations. 
Sometimes Vs30 data is available on a fine grid that contains millions of points. Users might want to prepare the site model in advance to run efficiently risk calculations. 

The association of vs30 values to the locations of the assets can be done in two different ways:
1) At the location of the assets
2) At the location of a grid, when using big exposure models with thousands of locations that exceed the computational resources.

**Vs30 at the asset locations:**
The `parse_site_model` command helps users to create a tailored site model file by associating the closest vs30 values at the locations of the exposure model.

Example: 
The image below presents an example using the USGS vs30 database and a simplified exposure model with 10 sites. 
exposure_sites

Using the command `oq parse_site_model Exposure.xml Vs30_values.csv` the resulting site model file contain the vs30 values closest to the exposure sites. The blue dots represent the generated site model file.
vs30_exposure_sites

**Vs30 in gridded points close to the asset locations:**
When dealing with large exposure files (thousands of sites), it is possible to create an uniformly spaced grid  #3499 . In this case the vs30 values will be the closest to the gridded points. This is possible by passing the flag `-g distance_in_km`. 

The image below shows in red dots the gridded points close to exposure sites, and the site model file includes the closest vs30 values to the gridded points. In this case a 20km grid was used (`parse_site_model Exposure.xml Vs30_values.csv -g 20`)

vs30_grid_sites

PR #4079, opened 2018-10-16
Offshore GMPEs - tested against results obtained using tables

mmpagani:
This is updating the Oceanic GMPE used for calculations in Canada. The comparison I did between the hazard curves obtained with this GMPE and its tabular version is not ideal. However, the contribution to total hazard by these sources is limited therefore I suggest to proceed for the time being. 

oceanic_median


micheles:
It looks like we miss a test for Oceanic

mmpagani:
Yes. Actually, the whole set of tests for these GMPEs needs to be revised.

micheles:
I added a regression test.

PR #4083, opened 2018-10-16
Pezeshk et al 2011 - New adjustment for NEHRP B/C

klunk386:
In this version is introduced a new approach to adjust reference conditions from original (>2000m/s) to NEHRP class B/C (about 760m/s). The method is based on the Atkinson and Adams (2013) paper.


micheles:
LGTM on my side

julgp:
It is ready to use on gemcontrol too?
thanks,
j.

PR #4087, opened 2018-10-17
Added check for missing exposure both in hazard and risk

micheles:
A bit better than the current error:
```python
        File "/home/michele/oqcode/oq-engine/openquake/calculators/base.py", line 614, in _read_risk_data
       taxonomies = set(taxo for taxo in self.assetcol.tagcol.taxonomy
   AttributeError: 'ClassicalRiskCalculator' object has no attribute 'assetcol'
```


micheles:
I am not adding a test on purpose (in the worst case a regression could generate a worse error message, but it would not be an issue anymore, since this is a very rare case).

PR #4088, opened 2018-10-17
Modified site term and added test to cover corner case + add reference

mmpagani:
This PR fixes a corner case related to the use of the NRCan15 modified GMPE (perhaps we should consider changing name) and adds the reference soil conditions for two GMPEs used in SHARE so that they can be also used with the site term.

### DIFF
--- a/openquake/hazardlib/gsim/mgmpe/nrcan15_site_term.py
+++ b/openquake/hazardlib/gsim/mgmpe/nrcan15_site_term.py
@@ -56,6 +56,10 @@ class NRCan15SiteTerm(GMPE):
             tmps = '{:s} does not use vs30 nor a defined reference velocity'
             msg = tmps.format(str(self.gmpe))
             raise AttributeError(msg)
+        if not (hasattr(self.gmpe, 'REQUIRES_SITES_PARAMETERS')):
+            self.REQUIRES_SITES_PARAMETERS = set(('vs30'))
+        if 'vs30' not in self.gmpe.REQUIRES_SITES_PARAMETERS:
+            self.REQUIRES_SITES_PARAMETERS.add('vs30')
         #
         # Check compatibility of reference velocity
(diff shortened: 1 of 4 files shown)


micheles:
I am in favor of changin the name of NRCan15, since it does not apply to Canada only.

openquake/hazardlib/gsim/mgmpe/nrcan15_site_term.py
```diff
@@ -56,6 +56,10 @@ def __init__(self, gmpe_name):
             tmps = '{:s} does not use vs30 nor a defined reference velocity'
             msg = tmps.format(str(self.gmpe))
             raise AttributeError(msg)
+        if not (hasattr(self.gmpe, 'REQUIRES_SITES_PARAMETERS')):
+            self.REQUIRES_SITES_PARAMETERS = set(('vs30'))
```
micheles: `(hasattr(self.gmpe, 'REQUIRES_SITES_PARAMETERS'))`

you do not need external parenthesis here.

micheles: `set(('vs30'))` is wrong, you want  `set(('vs30',))` or  `set(['vs30'])` or `{'vs30'}`.

klunk386:
I would change:
`if not str(imt) == 'PGA':`
to:
`if str(imt) != 'PGA':`

klunk386:
In my opinion, since this mgpe can be of general purpose, the site term should not be hardcoded as a class method, but should be passed by handle as it is done for the GMPE. In this way, the user can attach any site correction he wants by just specifying the site correction object (which can be ideally also a numerical solver)
I would then move BA08_AB06 in a separate file.

klunk386:
Since the "new rock reference conditions" are arbitrary (are imposed at line 77) I would propose to leave the reference Vs30 as free parameter, which should then be provided as input

Issue #4047, opened 2018-10-03
Make the tag names and occupancies lowercase


micheles:
Otherwise the risk team will make me go crazy.

micheles:
The idea is that user can write the tag names in any case (to avoid breaking old calculations), but when stored they become lowercase. Lowercase is the recommended way.

CatalinaYepes:
I don't agree that the tags should be forced to be lower case. That will break future joins or matches for which the tags are intended for.

PR #4091, opened 2018-10-18
Solve case issues with the tag names

micheles:
Closes https://github.com/gem/oq-engine/issues/4047. We should check that the aggregation still works at the plugin side.


micheles:
Since the tag names comes from shapefiles and are case-sensitive, we cannot normalize the case :-(
So I am forced to close the PR.

PR #4095, opened 2018-10-18
Added extract/avglosses_by_occupancy

micheles:
Added also an export; the usage is
```bash
$ oq export losses_by_tag/occupancy <calc_id>
$ oq export losses_by_tag/taxonomy <calc_id>
...
```


openquake/calculators/extract.py
```diff
@@ -562,3 +562,24 @@ def crm_attrs(dstore, what):
     """
     name = dstore['oqparam'].risk_model
     return ArrayWrapper(0, dstore.get_attrs(name))
+
+
+@extract.add('avglosses_by_occupancy')
+def avglosses_by_occupancy(dstore, what):
```
raoanirudh: Can we generalize this and make it `avglosses_by_tag`?

PR #4096, opened 2018-10-19
Add categorical site class as permitted site parameter

cvanhoutte:
Modified openquake/hazardlib/site.py, openquake/hazardlib/oqvalidation.py, and openquake/hazardlib/tests/gsim/check_gsim.py to take new site parameter ('siteclass').

Updated mcverry_2006.py and test tables.


openquake/hazardlib/site.py
```diff
@@ -56,7 +56,8 @@ class Site(object):
         :class:`Sites <Site>` are pickleable
     """
     def __init__(self, location, vs30=numpy.nan, vs30measured=False,
-                 z1pt0=numpy.nan, z2pt5=numpy.nan, **extras):
+                 z1pt0=numpy.nan, z2pt5=numpy.nan, siteclass=numpy.nan,
+                 **extras):
```
micheles: Remove siteclass from here: it will magically go in the `**extras` and work.

openquake/hazardlib/site.py
```diff
@@ -68,6 +69,7 @@ def __init__(self, location, vs30=numpy.nan, vs30measured=False,
         self.vs30measured = vs30measured
         self.z1pt0 = z1pt0
         self.z2pt5 = z2pt5
+        self.siteclass = siteclass
```
micheles: Remove siteclass from here

openquake/hazardlib/site.py
```diff
@@ -84,7 +86,7 @@ def __str__(self):
             "<Location=%s, Vs30=%.4f, Vs30Measured=%r, Depth1.0km=%.4f, "
             "Depth2.5km=%.4f>") % (
             self.location, self.vs30, self.vs30measured, self.z1pt0,
-            self.z2pt5)
+            self.z2pt5, self.siteclass)
```
micheles: Remove siteclass from here

micheles:
This is good. The missing thing is a complete end-to-end test. For that you need to add some example files. See for instance openquake/qa_tests_data/classical/case_31 and make a similar directory openquake/qa_tests_data/classical/case_32 with the same structure. My suggestion is to use a single source, 1 site, 1 trivial logic tree and parameters such that the test run in 1-2 seconds.

openquake/qa_tests_data/classical/case_32/job.ini
```diff
@@ -0,0 +1,38 @@
+[general]
+
+description = Classical PSHA with source specific logic tree (3**2 realizations)
```
micheles: You need a different description in this case ;-)

micheles:
You are nearly there. Now you need to go in the file /openquake/calculators/tests/classical_test.py and
add something like this:

```python
    @attr('qa', 'hazard', 'classical')
    def test_case_32(self):
        # source specific logic tree
        self.assert_curves_ok(['hazard_curve-mean-PGA.csv'], case_32.__file__)
```
Also, you must make sure that case_32 contains an empty file `__init__.py`. Then you can run the test as follows:
```bash
$ nosetests -vsx openquake/calculators/tests/classical_test.py:ClassicalTestCase.test_case_32
```
When it is green commit and push and then our continuous integration system (travis) should run all the other tests automatically.

cvanhoutte:
Hi Marco, thanks for your feedback. I completely understand what you are saying in that it would be simpler to keep Vs30 as the primary site parameter for all GMPEs in the engine. However, in New Zealand the site class is not based on Vs30 and the boundaries are instead defined by other parameters. Relatively frequently in NZ we end up in situations where the Vs30 cannot be generalised to the observed NZ site class, which is the method currently implemented in OQ. The existing McVerry model implementation is technically not quite correct because of this, which is why we propose this change. Having a SC variable becomes important in a logic tree where the McVerry model is used in conjunction with other Vs30-based models. Unfortunately we feel we need this change in order to be able to properly implement our Canterbury and Kaikōura seismic hazard models.

mmpagani:
Thanks a lot Chris for the explanation. What you write makes complete sense to me and your work on introducing the site class will be quite helpful in a lot of other cases. If it's not too much of an effort I would kindly ask to have a separate class for the McVerry since the current version is also used in a couple of other models of the mosaic.

micheles:
Also, if you use a different class name than the test_case_23 which is currently broken because it uses the old McVerry will work again.

cvanhoutte:
Hi Marco, no problem. Happy to implement as a separate class - I'll get onto that early next week, and will fix test_case_23. Chris

mmpagani:
Hi Chris, thanks and have a great we

micheles:
Just a couple of minor observations. First, there is an inconsistency between `siteclass` (no underscore) and `reference_site_class` (too many underscores): I would call it `reference_siteclass` instead. Second, I wonder if `siteclass` is too generic of a name. If it is going to be used only for the McVerry GMPE, should it be called `mcverry_siteclass`? Or , if, we can use this 1-character code even in other cases, what would be a good name for it?

cvanhoutte:
I'm ok with removing the underscore and changing to `reference_siteclass.`

I had thought about whether to name it `nz_siteclass` or simply `siteclass`. I chose the latter because other users might want to implement a gsim with a site class predictor, but with different definitions of site class than NZ. For example, users of AbrahamsonSilva1997 might be having similar issues as we are, as the site classification in their paper is based on the US Geomatrix site class but the implementation here only uses vs30. 

Of course, I'm more than happy to change it to `nz_siteclass` if you feel that is more appropriate.

Issue #3392, opened 2018-01-29
Investigate the `corner points do not lie on the same plane` error


micheles:
```python
  File "/home/michele/oq-engine/openquake/commonlib/calc.py", line 577, in __iter__
    for m1 in mesh])
  File "/home/michele/oq-engine/openquake/commonlib/calc.py", line 577, in <listcomp>
    for m1 in mesh])
  File "/home/michele/oq-engine/openquake/hazardlib/geo/surface/planar.py", line 191, in from_array
    return cls(mesh_spacing, strike, dip, tl, tr, br, bl)
  File "/home/michele/oq-engine/openquake/hazardlib/geo/surface/planar.py", line 142, in __init__
    raise ValueError("corner points do not lie on the same plane")
ValueError: corner points do not lie on the same plane
```

This happens with UCERF or with Canada, but I was never able to find a simple model exhibiting the error :-(

g-weatherill:
The problem, insofar as I've been able to discern, is due to the switching between spherical and cartesian coordinate systems. Certain configurations of plane and/or location can mean that when you generate a feasible plane in spherical coordinates, it can become distorted when translated to planar a planar system. I've encountered this a couple of times, which are both relevant for your case studies:

1. In UCERF the planes were built from fault sections that were projected down dip. However, rather than using the continuous mesh (the "Stirling method" that OpenQuake uses) UCERF builds its sources as planar faults for each section. The problem comes when you get a fault segment that is very short (say < 1 - 2km) but is then projected at a dip shallower than 90 degrees. Then the inclined fault plane is much wider than it is longer (maybe 15 - 17 km down dip and < 1 km along strike), and it seems that when this is projected into the cartesian system it does not define a plane according to OQs checks. Unfortunately, this occurs on a couple of faults in UCERF in which a small < 1 - 2 km "kink" at the end of certain dipping faults produces these types of very short segments.

2. The second case I encountered was for large ruptures at very high latitudes, which is probably why this is affecting the Canada model! I think here the problem is that as you get closer to the pole the potential distortion grows when translating to a cartesian system when the distance between points is large.

Honestly, I don't think there is a solution to this other than to switch off the planar correctness checks, or massively relax them, under certain conditions such as those labelled above.

micheles:
I should raise the priority of this because the Canada model is important now.

PR #4097, opened 2018-10-19
Removed the PlanarSurface check when unneeded

micheles:
Closes https://github.com/gem/oq-engine/issues/3392. Now ruptures from Canada can be exported, as requested by many people, including our sponsors. More importantly, I have added a check that the GMFs generated from ruptures in the datastore are the same - within numerical errors - than the GMFs generated from ruptures kept in memory. The procedure of saving/loading ruptures is not buggy, as it was suspected: instead, it is the PlanarSurface check which is buggy. In the future we may want to change the mesh to be 32 bit even when kept in memory - now it is 64 bit in memory and 32 bit in the datastore - then the numbers for the GMFs should be identical.


openquake/hazardlib/geo/surface/planar.py
```diff
@@ -90,14 +90,14 @@ def mesh(self):
...
     def __init__(self, strike, dip,
-                 top_left, top_right, bottom_right, bottom_left):
-        if not (top_left.depth == top_right.depth and
-                bottom_left.depth == bottom_right.depth):
-            raise ValueError("top and bottom edges must be parallel "
-                             "to the earth surface")
-
-        NodalPlane.check_dip(dip)
-        NodalPlane.check_strike(strike)
+                 top_left, top_right, bottom_right, bottom_left, check=True):
```
mmpagani: I'm fine with this solution for the time being. I wonder if the error is due to numerical approximation and could be solved by checking that the absolute difference is below a threshold.

micheles: Actually it is the first thing I tried but the threshold would be very high.

PR #4101, opened 2018-10-19
Reduce the GMPE logic tree upfront

micheles:
If there are missing TRTs in the source model (even before filtering) it must become smaller.


rcgee:
LGTM

rcgee:
LGTM

PR #4104, opened 2018-10-20
Turned vs30measured into an optional site parameter

micheles:
99% of the GMPEs do not require this parameter, still we are forced to pass it around always, because for the `Site` class it is a mandatory parameter. Instead, it should be an optional parameter like the backarc parameter or the geotechnical parameters. I wanted to do this for a long time, and I am doing it now
because users are starting to add additional site parameters (few months ago Pablo Heresi, now Chris van Houtte https://github.com/gem/oq-engine/pull/4096) and make mistakes trying to add all new parameters as mandatory parameters. BTW, in my opinion also the z1pt0 and z2pt5 parameters should become optional, but I am not doing that for the moment.
The change is not backward compatible so I would not do it if there is opposition from the scientists. I have already fixed all the tests in the gmpe-smtk repository.


g-weatherill:
I completely support this measure and eventually doing the same for Z1pt0 and Z2pt5. This has perhaps more urgency to change because by requiring it to be set even when the GMPEs that might use it are not being called may give the user the impression that changing Vs30 between measured and inferred might actually be doing something in the calculation, when in fact it is not.

PR #4111, opened 2018-10-23
defining reference velocity as 800

kejohnso:
Defined reference velocity for Allen2012 base class (800m/s) as was done for Somerville2009.


klunk386:
LGTM

PR #4116, opened 2018-10-24
Introduce tiling for risk

micheles:
The goal is to be able to run continental scale event based risk calculation, i.e. *the whole of Canada in a single calculation* without running out of memory, with tiles of `risk_tile_size` sites each (default 2000). Another new parameter to control the memory consumption is `assets_per_site_limit` (default 1000). To be tested.


micheles:
This does not save memory at all :-( `event_based_risk` is already yielding its result, so there is no point in producing one task per site.

PR #4120, opened 2018-10-25
20180830 sm pga

mmpagani:
This PR adds the possibility of computing PGA with the Si and Midorikawa GMPE. The tests are based on the values computed using a Fortran code kindly provided by Hongjun Si. The corrections are instead tested using an alternative implementation created with the support kindly provided by K. X. Hao and N. Morikawa. This updated version of the Si and Midorikawa requires a value of Vs30. This is used to decide if the motion must be computed for the original bedrock (600 m/s), scaled to 400 m/s by a factor of 1.41 or scaled to a reference of 800 m/s (factor of 1/1.25). The accepted values of vs30 are therefore only 400, 600 and 800.


klunk386:
LGTM, but as Graeme stated make sure you describe the correction factors, in particular that these are valid just for the elastic case (no adjustment for kappa is done so far).

mmpagani:
Added a short note. Thanks to all for the comments.

PR #4128, opened 2018-10-29
Disabled memory monitoring in macos

micheles:


openquake/baselib/parallel.py
```diff
@@ -454,7 +454,9 @@ def __iter__(self):
                 self.received.append(len(result.pik))
             else:  # this should never happen
                 raise ValueError(result)
-            if OQ_DISTRIBUTE == 'processpool':
+            if OQ_DISTRIBUTE == 'processpool' and sys.platform != 'darwin':
+                # it normally works on macOS, but not in notebooks calling
+                # notebooks, which is the case relevant for Marco Pagani
```
daniviga: This may happens also in other OS I guess. Taking into account that final users will use the VM with Linux.

PR #4147, opened 2018-11-05
Fixes the memory bug

micheles:
Since the beginning we had a memory bug in event_based_risk + sampling, in the sense of using more memory than needed. Since the unification between sampling and full enumeration now the memory bug is also in full enumeration. This is a good thing: it has become manifest and it can be fixed. This PR solves the issue. It is critical for the Global Risk Model.

The Indonesia calculation now does not run with a different error:
```python
  File "/opt/openquake2/oq-engine/openquake/calculators/getters.py", line 239, in init
    self.eid2idx = dict(zip(self.eids, range(len(self.eids))))
MemoryError: 
```


micheles:
By reducing the number of realizations to 25 via sampling the improvement is good enough that Indonesia can be run on oq2 (see https://oq2.wilson.openquake.org/engine/20313/outputs) so I am merging this.

PR #4164, opened 2018-11-09
20181109 site term

mmpagani:
This PR adds a second site term, similar in structure to the one already implemented in the OQengine but with a different scaling of the motion for values of Vs30 greater that 760 m/s. This implementation follows what suggested in http://www.daveboore.com/pubs_online/ab06_gmpes_programs_and_tables.pdf. 
The help of Michal Kolaj NRCan was critical for completing this.

### DIFF
--- a/doc/sphinx/openquake.hazardlib.gsim.rst
+++ b/doc/sphinx/openquake.hazardlib.gsim.rst
@@ -628,6 +628,22 @@ nga_east
     :undoc-members:
     :show-inheritance:
 
+nrcan15_site_term
+------------------------------------------
+
+.. automodule:: openquake.hazardlib.gsim.mgmpe.nrcan15_site_term
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+nrcan15_site_term_linear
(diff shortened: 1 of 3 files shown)


klunk386:
The docstring of NRCan15SiteTermLinear should be modified. I guess this is the old one.

klunk386:
Remove print at line 74

mmpagani:
> Remove print at line 74

Valerio I don't find it. In which file did you find it?

mmpagani:
The motion already refers to Vs30 760 hence we do not need to modify it 
(i.e. = is not needed).

> Modify idx = np.where(vs30 > 760) in idx = np.where(vs30 >= 760) in 
> order to include the 760 value (or alternatively idx = np.where(vs30 
> <= 760) at line 78)
>
> -- 
> You are receiving this because you authored the thread.
> https://github.com/gem/oq-engine/pull/4164#issuecomment-437413199




MARCO PAGANI | Seismic Hazard Team Coordinator | Skype mm.pagani | 
+39-0382-5169863

PR #4206, opened 2018-11-20
Reusing hazard for event_based_risk

micheles:
As requested by Catalina. There is a limitation, which is also a feature: it works only in single file mode. This means that:

1) if you have job_hazard.ini and job_risk.ini files nothing will change (backward compatibility!)
2) in order to use the feature you must merge job_hazard.ini and job_risk.ini files into a single job.ini

Then `oq engine --run job.ini` will

1. look if there is already a hazard calculation performed with the same hazard parameters
2. if yes, re-use it and run only the risk part of the calculation
3. if not, run both the hazard and risk parts of the calculation

PS: @daniviga, if we merge this branch, we will need to upgrade the engine database on each machine.


daniviga:
I would also check that if a job is found in the DB also the corresponding hdf5 is still on the filesystem. Otherwise we may have calculations that are still in the DB but without the required data. We do often a cleanup of hdf5 since we produce almost 1TB per day.

raoanirudh:
@micheles, @CatalinaYepes: As I see it, we have two separate issues that are somewhat linked.

The first one is that although hazard models are relatively stable compared to the risk models, one typically needs to rerun the hazard calculations frequently to ensure that the formats of the events and gmf data is compatible with the current version of the engine.

The second issue is that once the hazard jobs are run, the user needs to always keep track of the hazard calculation ids to correctly associate them with the subsequent risk jobs.

Having the engine try to automatically associate a risk job with an existing hazard calculation if it is still compatible will solve the two issues, but it will probably be very difficult to test all the cases well.

An alternate solution could be to allow the user to associate an event based risk job explicitly with previously exported outputs from the hazard job, like the gmf data (in addition to the current approach where the association is with a hazard calculation id). This would be easier to test since the hazard↔︎risk association is more explicit, remove the need for constantly rerunning hazard calculations, and also remove the need for keeping track of calculation ids on different servers.

micheles:
I am not sure I understand, @raoanirudh. If the user wants to know what was the hazard underlying a given risk calculation there are various ways to get it. For instance
```
$ oq1 db get_job 20978 damooduro
<Row(id=20978, description=Risk Iraq (ses=10k), user_name=damooduro, calculation_mode=event_based_risk, hazard_calculation_id=20977, status=complete, is_running=0, start_time=2018-11-20 10:17:52, stop_time=2018-11-20 10:21:46.391298, relevant=1, ds_calc_dir=/home/oqdata_shared/damooduro/oqdata/calc_20978, pid=9460, size_mb=40.68635654449463)>
```
or even
```
$ oq1 engine --sl 20978 |less
```
and look at the first line.

micheles:
After some thinking I decided to *disable* the reuse_hazard feature by default, even in the single job.ini situation. The user must enable it explicitly with the flag `--reuse-hazard` (`-r` for short). So by default everything is recomputed: this is the safest approach. It is the user responsibility to enable the use of the cache, for instance if she knows that the hazard was computed a short while ago and the engine code did not change in the meanwhile.
With this, I feel that I can merge.

PR #4228, opened 2018-11-26
Made it possible to specify the hazard map poes as return periods (optionally)

micheles:
As a convenience. In the conversion the poe is rounded to 6 digits after the decimal point.


micheles:
It does not look such an useful feature, so I am closing the PR.

PR #4234, opened 2018-11-27
BSSA14 distance-dependent phi computation fails when Rjb > R2

vSahakian:
In ~line 249 of boore_2014 in gsim, the following line will fail when Rjb is larger than R2, because `idx2` will be False

```
base_vals[idx2] += (C["DfR"] * (np.log(rjb[idx2] / C["R1"]) /
                                        np.log(C["R2"] / C["R1"])))
```

Therefore, `np.log(rjb[idx2])` will be empty, and cannot be broadcast into `basevals`. 
Suggest replacing with the following in order to match Equation 16 of Boore et al. (2014):
```
if idx2 == True:
            base_vals[idx2] += (C["DfR"] * (np.log(rjb[idx2] / C["R1"]) /
                                        np.log(C["R2"] / C["R1"])))
```


daniviga:
This MUST be an issue, not a pull request. Closing.

PR #4291, opened 2018-12-12
Gmpe weights

micheles:
TODO:
- [ ] what happens in case of sampling
- [ ] what happens in case of logic tree reduction
- [ ] what happens in case of multiple TRTs with multiple GMPEs each
- [ ] performance in large calculations


micheles:
This is too complicated. Superseded by https://github.com/gem/oq-engine/pull/4300

PR #4302, opened 2018-12-15
Added more FAQs

micheles:
I have added a few FAQs for hazard and none for risk, but I will keep adding stuff as time goes on.


micheles:
Here is the first risk FAQ! https://github.com/gem/oq-engine/blob/master/doc/faq-risk.md

daniviga:
Sorry I missed that

PR #4307, opened 2018-12-17
The beginning of a new manual for advanced users

micheles:
Written in Sphinx, not in latex. I am adding @raoanirudh as co-author because I expect him to add a lot of information from this manual (I think I have already extracted some relevant advice from one of his mails).


daniviga:
> I would like to have the same look and feel of the SVIR plugin manual, since it uses the same sphinx technology.

Which is the same we use for the engine

Issue #4309, opened 2018-12-18
ValueError raised when hypocenter variability


rcgee:
A ValueError is (incorrectly?) being raised in contexts.py when hypocenter variability is considered in the source model. This issue was raised by Francesco Visini and Bruno Pace on 17-12-2018. Hypocenter variability is used for some sources in the new Italian National Hazard model, e.g. in the Mt Etna region. 

https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/contexts.py#L244

For example, when we consider 3 possible locations of the hypocenter:

<hypo alongStrike="0.0" downDip="0.5" weight="0.33"/>
<hypo alongStrike="0.5" downDip="0.5" weight="0.34"/>
<hypo alongStrike="1.0" downDip="0.5" weight="0.33"/>

There is the following ValueError:

  File "/opt/openquake/src/oq-engine/openquake/hazardlib/calc/hazard_curve.py", line 106, in classical
    raise etype(msg).with_traceback(tb)
  File "/opt/openquake/src/oq-engine/openquake/hazardlib/calc/hazard_curve.py", line 102, in classical
    not rup_mutex)
  File "/opt/openquake/src/oq-engine/openquake/hazardlib/contexts.py", line 246, in poe_map
    src.num_ruptures, len(rups)))
ValueError: Expected at max 12 ruptures, got 36 (source id=SVF:0)

daniviga:
I would mark it critical since it's quite important for the national model completion

mmpagani:
Do we have the input file with the source SVF:0 that is raising this 
problem?

rcgee:
Yes I have it, and when I remove the hypocentre variability, the error is not raised.

PR #4310, opened 2018-12-18
Added a test for the directivity feature

micheles:
And closes https://github.com/gem/oq-engine/issues/4309. The bug was in the counting of the ruptures for split fault sources with directivity.


rcgee:
LGTM

PR #4312, opened 2018-12-19
shared_dir is now mandatory

daniviga:
Later on I will add some how-to about installing and configuring NFS, but this is not blocking the 3.3 branch detach anymore.


daniviga:
Added WIP because I forgot to update the part related to docker

PR #4318, opened 2018-12-19
Saving memory in Context.poe_map

micheles:
This should solve the problem with Australia.


micheles:
It solves the problem on the workers, but the master is still dying :-(

PR #4319, opened 2018-12-20
Fixed the number of generated tasks for optimize_same_id_sources=true

micheles:
This was affecting the Australia calculation.
Before the fix: Sent 670383 sources in 20545 tasks
After the fix: Sent 670383 sources in 1270 tasks


micheles:
LGTM

PR #4320, opened 2018-12-20
Changed the dictionary returned by hazard_curve.classical

micheles:
To be consistent with stochastic.sample_ruptures.


micheles:
LGTM

PR #4321, opened 2018-12-20
Remove tests from debian package.

nastasi-oq:
Remove any ``tests`` directory from oq-engine debian package.


daniviga:
`I` was expecting tests to be remove from the python package (and from packages in cascade), otherwise we'll not address the original issue (users running `nose` after a pip install). @micheles ?

packager.sh
```diff
@@ -1214,7 +1214,7 @@ GEM_BUILD_SRC="${GEM_BUILD_ROOT}/${GEM_DEB_PACKAGE}"
 mksafedir "$GEM_BUILD_ROOT"
 mksafedir "$GEM_BUILD_SRC"
 
-git archive HEAD | (cd "$GEM_BUILD_SRC" ; tar xv ; rm -rf rpm)
+git archive HEAD | (cd "$GEM_BUILD_SRC" ; tar xv ; rm -rf rpm ; rm -rf $(find . -type d -name tests))
```
daniviga: OT: I see you are removing `rpm`. We recently added `docker` too: does this mean that you have also to remove it? Why we don't simply take files coming out by the python package that are already filtered?

nastasi-oq: Currently we are packaging directly sources using git archive as source instead using python package, we can address this problem in a separated PR

PR #4325, opened 2018-12-20
Add a warning bar when using old browsers with the webui

### LINKED ISSUE #3746: Warn about unsupported browser
We should warn users about using an unsupported browser (IE < Edge, Safari) when they try to use the WebUI with one of them.

This is even more important in IPT (@nastasi-oq)

daniviga:
Closes #3746 

screenshot from 2018-12-20 12-20-28

Using https://github.com/burocratik/outdated-browser


hascar:
LGTM

PR #4327, opened 2018-12-20
Update message box when an unsupported browser is used

daniviga:
As requested by @pslh 

screenshot from 2018-12-20 15-27-52

Screenshot is old, bu I'm too lazy to make a new one. Now the button says "Download a supported browser" (instead of "newer")


hascar:
LGTM

pslh:
LGTM

Issue #4330, opened 2018-12-21
A missing retrofitted cost in the exposure should be an error for classical_bcr?


micheles:
I think so. Currently the BCR demo is broken because whoever added the tags to the exposure forgot the retrofitted costs, so the ratios are NaNs for all assets!

raoanirudh:
Agreed. For a BCR calculation, missing retrofitting costs in the exposure model should be an error.

PR #4331, opened 2018-12-21
Fixed the exposure in the ClassicalBCR demos

micheles:
The current exposure is broken! It has been broken for a few releases, because the retrofitted costs are missing, see https://github.com/gem/oq-engine/issues/4330. I am restoring the old exposure we had years ago, which has no tags but has the retrofitted costs.


raoanirudh:
We could keep the new exposure with the tags, and add the missing retrofitted cost to each asset. Looking at the old exposure xml, looks like we would simply need to add `retrofitted="5670.0"` to each asset in the ClassicalBCR demo.

micheles:
While we are at it, I would also switch to CSV exposure instead of XML exposure.

Issue #4328, opened 2018-12-21
Manage the case of site model and exposure in oq run


micheles:
Currently with `oq engine --run` the site model sites take the precedence over the exposure sites, but the opposite happens with `oq run`. Probably some tests will have to change, in particular event_based_risk/case_4.

PR #4337, opened 2019-01-08
Extract the hazard sites from the site_model, not the exposure

micheles:
This closes https://github.com/gem/oq-engine/issues/4328. Also, I was able to unify the parameters `asset_hazard_distance`  and `max_site_model_distance` by simplifying the association logic.


micheles:
LGTM

Issue #4151, opened 2018-11-07
Do not split the sources in event based


micheles:
In the classical calculator it is of paramount important to split the sources, so that the tasks can be distributed equally. In event based this used to be important too, but due the improvements of the last weeks now we can afford *not to split* the sources. Then the slow tasks will be dominated by sources with a slow `.iter_ruptures()` (typically an AreaSource with small discretization spacing or a ComplexFaultSource with small mesh spacing) but usual such sources can be run in minutes, unless the parameters are very wrong. Not splitting the sources makes it possible to devise new and more efficient strategies of sampling.
Some caveats are in order:

1. due to numerical errors, not splitting the sources will give (slightly) different numbers and lots of tests will have to be fixed
2. the splitting is done in the `split_filter` procedure; by removing it we will also remove the prefiltering; this means that we will lose the reduction of the logic tree, which has to be restored somewhere else
3. the filtering will still be done, but in the `event_based.sample_ruptures` tasks, not in `split_filter`
4. the logic tree reduction for missing TRTs will keep working as before, since it is done at an early stage.

*Interesting consequence 1*: running the Chile with full enumeration will give the error
```python
ValueError: There are too many potential logic tree paths (324) use sampling instead of full enumeration
```
because there is no more logic tree reduction from 324 to 54 realizations and I have recently added a limit on the maximum number of realizations in full enumeration (only for event based). By increasing `max_potential_paths` one can circumvent the error, ~but then the performance will be poor because 324 realizations will be considered~.

*Interesting consequence 2*: ~OQ_SAMPLE_SOURCES will not work anymore because the feature is inside split_filter~ (will work again because the feature is now inside `get_source_models` too).

*Interesting consequence 3*: by removing the splitting ~we will lose the minimum_magnitude feature,
which is implemented in `calc.filters.split_sources`~.

**The case of many SES**
Normally risk calculations have parameters like ses=1, years=10,000, however in some model the hazard  requires a specific investigation time (usually 1 year) and so the risk model is forces to use 10,000 ses or so. In this situation the calculation is dominated by the sampling, not by `.iter_ruptures()`, and the slow tasks can become a lot slower. However, thanks the recent improvement in https://github.com/gem/oq-engine/pull/4172, the problem disappears if the sources are not split.

PR #4338, opened 2019-01-08
Not splitting the sources in event_based

micheles:
I did not want to make this change in the middle of the global risk model calculations, but now it is the time to do it. With this approach the sources are not split (closes https://github.com/gem/oq-engine/issues/4151) and the seeds are build differently, resulting in huge speedups when there are lots of SES and/or samples.


micheles:
LGTM

PR #4352, opened 2019-01-11
Tl16

rcgee:
While preparing the new Italian National Hazard model, INGV noticed that some models within the tusa_langer_2016.py GMPE mistakenly used frequency values as periods in the coefficient tables. This bug was present in the the epicentral models, but not the hypocentral model. 

The values have been corrected in the GMPE and in the verification tables.


daniviga:
Does this needs backport to 3.3? @micheles

micheles:
This is a question for @mmpagani

daniviga:
> I don’t think it’s necessary. Recently they run calculation from master.

No, they don't. They are running a "feature frozen" nightly build which _is_ 3.3 in practice, but they must be brought back to stable ASAP (the effort to keep them on unstable is too high).

micheles:
LGTM on my part, if @mmpagani is happy. Please update the changelog.

Issue #4348, opened 2019-01-10
The aggregated loss curves must be computed from the aggregate losses, not by aggregating the loss curves


micheles:
Says @raoanirudh:

> Indeed, the algorithm for computing the loss curves by tag needs to be corrected in the engine. To get the 200-year loss for occupancy=COM, say, the engine is currently simply summing the 200-year loss for all assets that belong to the tag-group occupancy=COM. This leads to losses for each return period much larger than expected for the country. We are summing the 200-year loss for Veneto with the 200-year loss for Sicily, although they're most likely coming from very different events. 

> Instead, the engine should create an event loss table for each tag-group (this table will contain for each event, the sum of the losses coming from all assets that belong to the tag-group), sort the event losses in descending order, and obtain the losses corresponding to the different return periods for the tag-group in the same way as is done for the aggregate loss curves.

> @Michele Simionato, a quick test would be to compare the value for the occupants loss (for any return period) between the aggregate loss curves file (eg: https://gitlab.openquake.org/risk/global_risk_model/Europe/blob/master/Risk/Italy/agg_loss-mean.csv#L9) and the loss curves by occupancy file for occupancy=Res (eg: https://gitlab.openquake.org/risk/global_risk_model/Europe/blob/master/Risk/Italy/curves_by_tag-occupancy-mean.csv#L37). The values in the two files should be identical for all return periods.

My plan is to remove the support for loss curves and maps from the event_based_risk calculator and use a brand new ebrisk calculator when computing the aggregate loss curves and maps is required. The ebrisk calculator should be simpler than the current calculator powerful enough for the global risk model purposes. In particular it must be able to compute

- [x] losses_by_event
- [x] loss_curves
- [x] loss_maps
- [x] avglosses_by_rlz

and at the same time

- [x] use little memory

My analysis shows that ebrisk calculator is a lot more efficient than event_based_risk, but there may be slow tasks :-(

micheles:
The test suggested by Anirudh passes apart from minor numerical differences due to the 32 bit precision.

PR #4354, opened 2019-01-12
Implemented avglosses_by_rlz in ebrisk

micheles:
Another step for https://github.com/gem/oq-engine/issues/4348.


micheles:
LGTM

PR #4355, opened 2019-01-12
Ordered the events by realization

micheles:
This makes the ebrisk calculator a lot faster when building the losses by realization, loss curves and maps (measured speedup 100x with 1.5 GB of data and 81 taxonomies, but it can be a lot more). In case of asset correlation the numbers will be different because there is a different epsilon<->eid association (again, this association is arbitrary, akin to a change of master_seed, so there is no issue there).


micheles:
LGTM

Issue #4356, opened 2019-01-12
The GMF csv exported should also export the event IDs


micheles:
Exporting the event IDs was a bad idea, since the export will change if I change the order in which events 
are stored (a thing I want to do in engine 3.4). The fix should be backported to engine 3.3 if possible.

PR #4357, opened 2019-01-12
Fixed the GMF export to export the event IDs again

micheles:
In this way the export becomes robust if in a future version of the engine I change the ordering of the events table. Closes https://github.com/gem/oq-engine/issues/4356.


micheles:
LGTM

PR #4359, opened 2019-01-14
Changed the agg_curves exporter

micheles:
In such a way that it can be generalized to multiple tags, so that aggregate loss curves and maps can be exported from an ebrisk calculation. Another step towards https://github.com/gem/oq-engine/issues/4348.


micheles:
LGTM

PR #4363, opened 2019-01-15
Reading the exposures in parallel

micheles:
For instance in the case of Europe there are 46 exposures. By reading in parallel the import time goes down from 9 minutes to 2 minutes, which is a decent speedup.


micheles:
LGTM

PR #4367, opened 2019-01-17
Shutdown the processpool after parallel_split_filter

micheles:
This saves a bit of memory. For the Australia calculation it means ~16 GB on wilson, better than nothing, but not enough to run the full calculation.


micheles:
LGTM

PR #4372, opened 2019-01-20
Speed-up store_source_info [aka Australia runs!]

micheles:
The reason why the Australia calculation (but also the calculation for Italy by the INGV) fails is that the controller node is too slow: too much data comes from the workers, it is not saved fast enough, goes into the zmq queue and eventually the master node runs out of memory. The good news is that the aggregation of the hazard curves is fast, the problem is only in the function `store_source_info` which is 20x slower and which is used only for information purposes. The solution is easy: do not call for each task result, but only at the end, a single time.


micheles:
LGTM

PR #4373, opened 2019-01-20
Tl16 fix coeff

rcgee:
After noticing an anomalous-looking spike in the response spectra of the TusaLanger2016RepiBA08SE model at T=0.14s, we contacted the authors who found a mistake in one of the coefficients in the publication. It has now been updated according to the authors suggestion in the GMPE and test tables.


micheles:
I guess this is important for the INGV again, right?

rcgee:
@micheles Yes, this GMPE is important for INGV. Although, I do not know which model they are specifically using (TusaLanger2016RepiBA08SE or one of the others). I asked, but have not heard back yet.

PR #4374, opened 2019-01-21
Fixed a bug in scenario_damage from ShakeMap with noDamageLimit=0

micheles:
Added a test and converted the script `utils/shakemap` into a command `oq download_shakemap`.


VSilva:
LGTM

PR #4376, opened 2019-01-21
Remove risk stats

micheles:
The risk stats that are currently implemented in the engine are actually hazard stats. It is better to remove the confusion now, before implementing a logic tree for the risk functions (i.e. real risk statistics).


raoanirudh:
The epistemic uncertainty in the hazard (modeled through the source and gsim logic trees) gets propagated to the risk. Although the engine doesn't provide a way to model the epistemic uncertainty in the exposure or vulnerability yet, it is still quite valuable to be able to export the quantiles for the risk results.

raoanirudh:
Won't that end up being more confusing, though, if the user has to write `quantile_hazard_curves` in the risk job file in order to produce the quantile loss curves?

micheles:
Right, so now the user can specify "mean" and "quantiles" and all the statistics will be computed (hazard curves if any, avglosses if any, loss curves is any, loss maps if any, etc).

raoanirudh:
@micheles, this change needs to be reflected in the manual

Issue #4377, opened 2019-01-21
Improve the situation with the slow task in the ebrisk calculator


micheles:
The problem is visible in the Chile calculation. We should reduce it to make it more debuggable.

micheles:
By using a small enough number of tasks and a rupture weight sensitive to the number of occurrences and affected taxonomies I was able to compute Chile with 25 samples and 20,000 SES in 1 hour, with
the slowest task taking only 15 minutes on cluster2.
```
$ oq2 show task_info 22297
============================ ======= ======= ======= ======= =======
operation-duration           mean    stddev  min     max     outputs
ebrisk                       187     184     0.87947 919     1,357  
```
This is good enough for the moment.

PR #4381, opened 2019-01-23
Changed the rupture weight to depend on the number of affected assets

micheles:
It should address https://github.com/gem/oq-engine/issues/4377
We are now storing the  number of assets per site and the time spend to compute each site.


micheles:
LGTM

PR #4388, opened 2019-01-25
More work on the slow tasks in ebrisk

micheles:
See https://github.com/gem/oq-engine/issues/4377. Non in oq2 we can run Chile in 1.5h instead of 10h bu we are not there yet, oq1 is still slow :-(


micheles:
LGTM

PR #4389, opened 2019-01-27
Small fix in the generation of the .rst reports

micheles:
The weight and the number of effective ruptures were not reported correctly. I have fixed it by using the preclassical calculator in the reportwriter.


micheles:
LGTM

PR #4393, opened 2019-01-29
Postfilter the ruptures after sampling

micheles:
In engine < 3.3 the ruptures were filtered with the `filter_distance` after the sampling and this was very slow, so I removed that part in engine 3.3. The consequence was that a lot more ruptures than needed were stored (i.e. ruptures outside of the maximum distance) and then discarded at reading time. This was deemed acceptable since the storing operation is fast, faster than distance filtering.
Now I am adding a simple filtering based on the rupture bounding box which is faster: ruptures outside of the maximum distance may still be saved, but not a huge amount. This will save disk space and data transfer; the saving will be a little faster at a cost of a little time spent on bounding box filtering.

NB: the ruptures are still filtered with `filter_distance` during the GMF calculation, that has not changed, so ruptures exceeding the maximum_distance will never have an effect, as before.

PS: users should be happier now, see https://groups.google.com/d/msg/openquake-users/I5t5OaljIfY/g_HcKqlSDQAJ

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,6 +1,9 @@
+  [Michele Simionato]
+  * Only ruptures boundingbox-close to the site collection are stored
+
   [Marco Pagani]
   * Added cluster model to classical PSHA calculator
-  
+
   [Michele Simionato]
   * Fixed a bug in scenario_damage from ShakeMap with noDamageLimit=0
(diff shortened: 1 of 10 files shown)


micheles:
LGTM

Issue #4396, opened 2019-01-29
Add an option in local_settings to disable WebUI


daniviga:
Add a simple option in `local_settings.py` that allows to expose the REST API but not the WebUI. This may be useful for pure machine driven services.

PR #4397, opened 2019-01-29
Allow only REST API to be exposed

daniviga:
With this PR the WebUI and all the HTML part (i.e. tools) of the Engine server can be disabled leaving the REST API as the only end point exposed.

Behavior is not changed compared to the past with the default settings.

Closes https://github.com/gem/oq-engine/issues/4396

### DIFF
--- a/openquake/server/urls.py
+++ b/openquake/server/urls.py
@@ -31,6 +31,9 @@ urlpatterns = [
     url(r'^v1/on_same_fs$', views.on_same_fs, name="on_same_fs"),
 ]
 
+# it is useful to disable the default redirect if the usage is via API only
+# 'collectstatic' and related configurationis on the reverse proxy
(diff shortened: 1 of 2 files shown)


openquake/server/urls.py
```diff
@@ -23,24 +23,27 @@
...
-    url(r'^engine/license$', views.license,
-        name="license"),
 ]
 
-for app in settings.STANDALONE_APPS:
-    app_name = app.split('_')[1]
-    urlpatterns.append(url(r'^%s/' % app_name, include('%s.urls' % app,
-                           namespace='%s' % app_name)))
+if settings.WEBUI:
+    urlpatterns += [
```
micheles: Please add a comment that it is useful to disable the redirect if the usage if via API only

micheles:
LGTM

Issue #4390, opened 2019-01-28
Optimize the ebrisk calculator


micheles:
Surprisingly, the calculator is dominated not by the risk calculation but by index manipulations in a Python loop that must be vectorized. This is nontrivial and will require several pull requests, but I expect a speedup of over an order of magnitude, so it must be done.

micheles:
The problem was worked around by reducing the power of the calculator: now it can only compute the mean losses_by_site, but that is enough according to the risk team.

PR #4400, opened 2019-01-30
Some refactoring of the ebrisk calculator

micheles:
The speed is the same, but the code is cleaner. Also, now I am logging the time spent in the avg_losses computation, which is the dominating factor for Chile:
```
   building risk                  88,946   111       208        
   building avg_losses            61,809   0.0       243,200,742
   getting hazard                 10,450   355       208        
   getting assets                 2,938    0.0       208        
   computing risk                 2,214    0.0       213,779    
```
Useful for https://github.com/gem/oq-engine/issues/4390.


micheles:
LGTM

PR #4402, opened 2019-01-31
Saving only the mean losses_by_site

micheles:
This is enough for our purposes (plotting a map) and it makes the calculation of the losses by site 4x faster.
```
building avg_losses            61,809   0.0       243,200,742 # before
building avg_losses            15,484   0.0       243,200,742 # now
```
The total runtime is now half than before and under 25 minutes for Chile on cluster2.


micheles:
LGTM

PR #4403, opened 2019-01-31
Speedup losses_by_site even more

micheles:
I gained nearly another order of magnitude by smarter array manipulations.
```
building avg_losses        15,484   0.0       243,200,742 # before
building avg_losses        1,919    0.0       243,200,742 # now
```
Now Chile runs in less than 20 minutes.


micheles:
LGTM

PR #4409, opened 2019-02-01
Forced ebrisk to start from precomputed ruptures

micheles:
So that the risk guys are happy. Also fixed a bug that made it impossible to run the calculation unless the exposure was stored in the rupture file. I have changed `case_global` in oq-risk-tests to test the workflow. Notice that the `job_h.ini` file requires (even if looks surprising) the site model and intensity_measure_types parameters, on top of the less surprising source model and gsim logic tree files. The `job_h.ini` may contain the site model file (in that case the site model parameters need not to be specified) and the vulnerability functions (in that case the intensity measure types need not to be specified).


micheles:
LGTM

PR #4412, opened 2019-02-01
Tried RtreeFilter in South America

micheles:
It is 50% slower, so it is an approach to reject :-(
```
   ============================== ======== ========= =========
   ============================== ======== ========= =========
   calc rupture weight            3,087    0.0       3,708,800  # numpy
   calc rupture weight            4,542    0.0       3,708,800  # Rtree
```


micheles:
LGTM

Issue #4420, opened 2019-02-04
Ideas for the future of the disaggregation calculator


micheles:
At the moment the disaggregation calculator is pretty slow, particularly for models with large logic trees.
There various options that can make the calculation faster.

- [x] force the user to specify the realization he is interested in in the job.ini (for instance with a syntax like `rlz_index=42`) unless there is a single realization.
- [x] now that the classical calculator stores information about the ruptures (https://github.com/gem/oq-engine/pull/4414) the disaggregation calculator should take advantage of it to avoid repeating already performed calculations
- [x] ~we could store even more information, in particular the output of `gsim.get_mean_and_stddevs()` for each rupture, then the disaggregation could use that~ [it is not worth it]
- [x] a revolutionary idea by @mmpagani  is to NOT parallelize by sources; instead we could parallelize by ruptures ~and we could order them in distance bins, removing the ones that contribute little~
- [x] probably the disaggregation calculator should be turned into a postprocessing calculator, i.e. with a `hazard_calculation_id` referring to a previous classical calculation (with few sites)
- [x] if we force a specific realization, chosen by the user or by the engine (for instance by choosing the realization closest to the mean hazard curve) then we must remove the disaggregation statistical outputs
- [x] ~Laurentiu suggests to remove the lon-lat disaggregation, since it is little used and not so interesting~
- [x] ~Julio would like to run 1 classical calculation with 1000 cities and then 1000 disaggregations, one per  city, using only the realization closest to the mean hazard curve for that city, efficienty~ [we can do everything in a single computation now]

micheles:
I am closing the ticket since everything that was worth doing has been done, the speedup is up to 10,000+ times and there is no reason to do further optimizations for the moment.

PR #4422, opened 2019-02-05
When computing the hazard curves, also determine the realization closest to mean curve

micheles:
For each site, across all levels and IMTs. Store the closest realizations as an array `best_rlz` of length N, the number of sites. To see if two hazard curves are close, I am using a square distance on the logarithm of the PoEs. I expect this feature to be useful for disaggregation (see https://github.com/gem/oq-engine/issues/4420), so I am restricting it to the case of few sites.


micheles:
LGTM since it is harmless

PR #4429, opened 2019-02-06
Renamed AverageSA-> AvgSA

micheles:
Now the name of the IMT is consistent with the name of the class GenericGmpeAvgSA (and it is also shorter, consistently with other IMT names).


klunk386:
LGTM

Issue #4427, opened 2019-02-06
Remove restriction that all site depths must be zero for Mesh.get_distance_matrix()


g-weatherill:
The `get_distance_matrix` function is called when applying spatial correlation and currently has an assertion that all depths should be zero. This was OK when site depth/elevation was always considered to be 0, but now that more 3D distances for sites can be used then this check prevents one from running spatial correlation when sites are at depths different from zero. 

There are two options:
1. Given that the spatial correlation models have tended to ignore depths in their distance definitions then simply removing the `assert (self.depths==0).all()` would mean that 2D distances would still be used for the spatial correlation model and simply depth/elevation differences would still be ignored (probably the preferred/easiest option).

2. Modify `openquake.hazardlib.geo.geodetic.geodetic_distance` to account for depth\elevation (optionally) in order to account for 3D distance. This is not entirely faithful to the definition of the spatial correlation models but would very likely be inconsequential most of the time given that elevation differences would be substantially smaller than horizontal distances.

mmpagani:
I am fine with option one. Perhaps if sites have a depth (i.e. elevation) different than 0 we can raise a warning.

PR #4431, opened 2019-02-06
Remove depth check

rcgee:
This PR removes the check that prevents one from running spatial correlation when sites are at depths different from zero. It addresses Issue #4427 (uses Option 1)

Now get_distance_matrix allows 3D sites (i.e. sites with depth/elevation)


openquake/hazardlib/geo/mesh.py
```diff
@@ -268,8 +268,7 @@ def get_distance_matrix(self):
         """
         Compute and return distances between each pairs of points in the mesh.
```
micheles: Just add in the docstring something like
```
NB: the depth of the points is ignored.
```

openquake/hazardlib/correlation.py
```diff
@@ -55,6 +55,10 @@ def apply_correlation(self, sites, imt, residuals, stddev_intra=0):
         # of N random numbers (where N is equal to number of sites).
         # we need to do that multiplication once per realization
         # with the same matrix and different vectors.
+
+        if numpy.any(abs(sites.depths) > 1e-7):
+            raise Warning('Depths are not considered when dealing with spatial'
+                          ' correlation')
         try:
             corma = self.cache[imt]
         except KeyError:
```
micheles: I am against the warning. The risk is that  it will be raised lots and lots of time flooding the log and causing a general slowdown of the engine, on top of annoying the users.

PR #4436, opened 2019-02-08
Moved the Akkar coefficient table into a csv

micheles:
This solves the pylint violation issue and make the data a lot more readable.


klunk386:
I don't quite understand this. I was told that the philosophy behind the currently implemented GMPEs is to have hard-coded coefficient tables. What is then the advantage to have these coefficients in a separate csv now? If the issue is with readability, we should probably do the same with all GMPEs in the hazardlib, I guess.....

klunk386:
A part from this philosophical issue, LGTM

micheles:
The issue is that Jenkins keeps sending annoying emails for code style violations: 
https://ci.openquake.org/job/master_oq-engine/4760/
The other coefficients tables in hazardlib are not so long or better formatted so that Jenkins did not complain yet.

klunk386:
OK then

PR #4438, opened 2019-02-08
Removed the realization index from the event loss table export

micheles:
It is redundant now that the realization can be extracted from the event. In the future we could remove the realization index even from the datastore, but not yet.


micheles:
I must extend this work to scenario_risk too.

PR #4454, opened 2019-02-14
Removed most of nose usage

micheles:
This is in preparation for the move to pytest. The packager should be changed a bit to set the variable OQ_RUN_SLOW_TESTS or I should manage the variable GEM_DEVTEST_SKIP_SLOW_TESTS.


openquake/hazardlib/tests/source/rupture_test.py
```diff
@@ -178,7 +176,7 @@ def test_get_dppvalue(self):
 
             self.assertAlmostEqual(dpp, ref_dpp, delta=0.1)
 
-    @attr('slow')
+    @unittest.skipUnless(os.environ.get('OQ_RUN_SLOW_TESTS') == '1', 'slow')
```
daniviga: I would just check that the variable exists

daniviga:
We can merge https://ci.openquake.org/job/zdevel_oq-engine/2832/

Issue #4450, opened 2019-02-13
GMPE tables do not work anymore in event based


micheles:
Here is the error we get for Canada:
```python
     File "/home/michele/oqcode/oq-engine/openquake/hazardlib/gsim/gmpe_table.py", line 519, in _return_tables
       periods = self.imls["T"][:]
   TypeError: 'NoneType' object is not subscriptable for GMPETable(gmpe_table='./gm_tables/Wcrust_med_clC.hdf5'), SA(0.3), srcidx=16
```
The problem is that the engine is trying to read the tables from the wrong directory. The clean solution, given that the tables are already in the datastore, is to read them from there. This will require a lot of work and several pull requests.

PR #4458, opened 2019-02-14
Added a mechanism to delay the initialization of the GMPEs

micheles:
The trick is to add to the GSIM classes a method `.init()` callable with no arguments. The final goal would be to make the GSIM serializable to JSON and back without need to read external files at instantiation time, but later on. Then I can save the JSON in the datastore and fully instantiate the GSIMs from the datastore if the internal files information is stored too (this is the case for GMPETables). Then https://github.com/gem/oq-engine/issues/4450 can be solved properly. The complete solution will require more pull requests.


openquake/hazardlib/gsim/gmpe_table.py
```diff
@@ -302,50 +301,24 @@ class GMPETable(GMPE):
...
-        # we want to run a calculation on machine 1, copy the datastore
-        # on machine 2 (that misses the .hdf5 files) and still be able to
-        # export the results, i.e. to instantiate the GsimLogicTree object
-        # which is required by the engine to determine the realizations
-        if os.path.exists(self.GMPE_TABLE):
-            with h5py.File(self.GMPE_TABLE, "r") as f:
-                self.init(f)
-
-    def init(self, fle):
+    amplification = None
```
mmpagani: Can you pls add a comment to explain the meaning of this variable?

micheles: Will do in the next PR

Issue #4460, opened 2019-02-15
TOML format for the GsimLogicTree class


micheles:
The XML syntax to manage GMPEs with arguments is currently insufficient. For instance the following specification
```xml
                  <uncertaintyModel gmpe_name="BooreAtkinson2008">
                     NRCan15SiteTerm
                  </uncertaintyModel>
```
only works because `BooreAtkinson2008` is a GSIM without parameters, but it would not be possible to
specify a GSIM with parameters, say the `GenericGmpeAvgSA` recently introduced by Valerio. The same limitations applies to `MultiGMPE`, i.e. dictionaries of gsims by intensity measure type recently introduced by Graeme, their underlying GSIMs cannot have parameters.
Such limitations are artificial and could be lifted by

1. complicating the XML parser
2. alternatively switching to a more decent format ([TOML](https://github.com/toml-lang/toml#user-content-example))
3. alternatively keeping the XML but using bits of TOML inside

Because of backward compatibility issues option 2 is not attractive (if I deprecated the old format, users would be confused, and then I would have to keep the old parser anyway, in order to be able to run old calculations). Option 1 is a lot of work while option 3 could be acceptable and can be made totally backward compatible.

Apart from the users, we also need a compact human readable text format for the GsimLogicTree class, so that it can be serialized in the datastore and then deserialized. Since this part is totally internal I can choose whatever I like and TOML looks like the best contender among the XML, JSON and YAML formats - [YAML is bad](https://arp242.net/weblog/yaml_probably_not_so_great_after_all.html), JSON is not readable and nobody likes XML.
Here are examples of how the GSIM logic tree would like inside the datastore (currently it is in XML)
starting from the ordinary case of GSIMs without parameters:
```toml
["Active Shallow Crust".b1]
AkkarBommer2010 = {}
weight = 0.35

["Active Shallow Crust".b2]
CauzziFaccioli2008 = {}
weight = 0.35

["Active Shallow Crust".b3]
ChiouYoungs2008 = {}
weight = 0.20

["Active Shallow Crust".b4]
ZhaoEtAl2006Asc = {}
weight = 0.10
```
An example with the Marco's site term
```toml
["stable continental".b1]
NRCan15SiteTerm = {gmpe_name = "BooreAtkinson2008"}
weight = 1.0
```
An example with Valerio's GSIM
```toml
["Active Shallow Crust".b1.GenericGmpeAvgSA]
gmpe_name= "BooreAtkinson2008"
[...]

PR #4461, opened 2019-02-15
Added toml to the dependencies

micheles:
This is needed for the plan described in https://github.com/gem/oq-engine/issues/4460. I am not specifying the toml version since the format is stable and every version should work even if the current release number is 0.10 (see https://pypi.org/project/toml/). I guess they did not have the courage to call it 1.0.


daniviga:
For the wheelhouse we need to chose a version. I would take the current stable on, sound good?

https://pypi.org/project/toml/0.10.0/

PR #4463, opened 2019-02-15
Added a forgotten ses_ratio in ebrisk

micheles:
The avg_losses were bigger by a factor equal to the number of stochastic event sets.


micheles:
@raoanirudh, the ses_ratio in the engine is
```python
(self.risk_investigation_time or self.investigation_time) / (
            self.investigation_time * self.ses_per_logic_tree_path)
```
so it should be correct already, can you point me to a test which gives incorrect average losses after the fix applied here?

raoanirudh:
You're right, with this fix, the values also account for the `risk_investigation_time` correctly.

PR #4467, opened 2019-02-15
20190123 ebased

mmpagani:
This PR adds a new OQ engine calculator which offers the possibility of generating SES with source models containing *cluster* sources with mutually exclusive ruptures, as in the case of the New Madrid FS included in the 2008 and 2014 USGS National Seismic Hazard Model. 

I performed a number of external checks to make sure that the calculator provides the number of ruptures expected. For example, using the USGS cluster model for the NMFS with a recurrence interval of 500yrs, the number of ruptures produced for 10000 years is 60 i.e. 20 clusters with 3 ruptures each. 

In an additional pull request, I will also incorporate and homogenise the capabilities requested to properly simulate SES for some of the sources in the Japan model.


openquake/calculators/event_based_advanced.py
```diff
@@ -0,0 +1,119 @@
...
+                #  par['group'] = True
+                #  par['cluster'] = sg.cluster
+                #  par['src_interdep'] = sg.src_interdep
+                #  par['rup_interdep'] = sg.rup_interdep
+                #  Here we create a WeightedSequence similarly to what is
+                #  done with traditional sources
+                #  dat = [(s, 1.) for s in sg.sources]
+                #  block = WeightedSequence(dat)
+                if hasattr(sg, 'atomic'):
+                    smap.submit(sg, self.src_filter, par)
```
micheles: This is no good. You want if `sg.atomic` instead.

openquake/hazardlib/calc/stochastic.py
```diff
@@ -161,22 +161,90 @@ def sample_ruptures(sources, srcfilter, param, monitor=Monitor()):
     [grp_id] = set(src.src_group_id for src in sources)
     eb_ruptures = []
     ir_mon = monitor('iter_ruptures', measuremem=False)
```
micheles: This function is too long, please split it at least in two pieces

micheles:
Please update the changelog

Issue #4442, opened 2019-02-08
It looks like that process spawning is sometime unstable on Windows


daniviga:
Demos on Jenkins hang randomly when running in Windows, also running all the demos in batch in different environments exercised similar behaviors. It seems that it depends heavily on the load on the machine.

See https://ci.openquake.org/job/windows/job/master_oq-engine/2132/console but aslo
![screenshot from 2019-02-08 17-24-19](https://user-images.githubusercontent.com/1818657/52491065-87994a80-2bc6-11e9-8f44-7b809084ab0a.png) (has been stuck there for >30mins)

I was able to reproduce similar issues on
- Windows 10 Pro (1803) - KVM on Optiplex
- Windows 7 Pro - KVM on Optiplex
- Windows 7 Pro - KVM on P710 (jenkins)

but not (yet?) on a physical Windows 10 (1803) Home.

All machines have 4 physical cores and 8 GB of RAM. VM are on spinning HDD, physical one has an SSD.

It needs further investigation (i.e. with an old stable): unfortunately Jenkins Win was off since Oct 30th 2018, and Engine 3.3 has been a huge release, so we don't know when problems started appearing.

/cc @pslh

daniviga:
A testing environment can be downloaded here: https://cloud.openquake.org/s/b8TqiYLfKnTGo6w

How to use it:
 - Extract the `7z` archive
 - Run `oq-console.bat`
 - Run `pip install -e oq-engine` (only for the first time)
 - Run `oq-demos.bat`

daniviga:
![screenshot from 2019-02-08 18-04-29](https://user-images.githubusercontent.com/1818657/52493557-12307880-2bcc-11e9-9a3e-1e3e7834413e.png)

Another point (stuck for 20 mins then killed manually)

daniviga:
@micheles are we sending back results via zmq also from spawned processes on a single machine?

daniviga:
We have a problem with sockets anyway: socket are created but not reused.
[dump.txt](https://github.com/gem/oq-engine/files/2846163/dump.txt)

daniviga:
As @micheles suggests it may be a regression introduced in https://github.com/gem/oq-engine/pull/3986

micheles:
Daniele says that the problematic sockets are only the ones connected to the port 1908, so we must investigate the DbServer.

PR #4475, opened 2019-02-20
Keeping open the DbServer client socket

micheles:
This is an attempt to close https://github.com/gem/oq-engine/issues/4442. If we are lucky, zmq will manage internally socket reconnection and this should work even for calculations taking days. I have checked on linux that with this change the number of connections in TIME_WAITS goes down from 400+ to a dozen while running the demos.

Windows seems to be happy: https://ci.openquake.org/job/windows/job/zdevel_oq-engine/36/


daniviga:
This PR has been reverted in https://github.com/gem/oq-engine/pull/4665

Issue #4476, opened 2019-02-20
Same file with different cases in a gsim test folder


daniviga:
Windows complains

![screenshot from 2019-02-20 14-18-46](https://user-images.githubusercontent.com/1818657/53094493-8f3be600-351a-11e9-92e5-3ad742826472.png)

Regression of #4472

rcgee:
#4478 should solve this issue

PR #4477, opened 2019-02-20
Removed mixed case test filename

micheles:
I have added a test to force people to use lowercase filenames. In this instance Robin committed a file called `yu_2013_mw_test.py` while there was a file called ` yu_2013_Mw_test.py`, so we ended up with two copies of the same test.

Closes https://github.com/gem/oq-engine/issues/4476.


mmpagani:
`yu_2013_test.py` should be also removed since it's equal to `yu_2013_ms_test.py`

rcgee:
LGTM

PR #4472, opened 2019-02-19
Update yu13 gmpe

rcgee:
This PR updates the Yu (2013) GMPEs for China.  Changlong Li sent the latest version of the GMPEs via email on Feb 18, 2019.  The only changes are to the method:

`def get_ras(repi, theta, mag, coeff)`

This affects the Ms and Mw models. The changes are small because the tests are still passing.

PR #4478, opened 2019-02-20
remove duplicated test files

rcgee:
My intention in #4472 was to rename the test files as:

yu_2013_Mw_test.py > yu_2013_mw_test.py
yu_2013_test.py > yu_2013_ms_test.py 

because I wanted to remove the capital letter, and make the naming more consistent. 

However, I forgot to commit the deleted files, hence we ended up with duplicate files.


micheles:
No need for this,@rcgee, it is already fixed in https://github.com/gem/oq-engine/pull/4477

Issue #4176, opened 2018-11-12
Add a check for missing occupants column in the exposure


micheles:
The error should appear at the beginning, not in the middle of the computation. This happened to Anirudh for Japan.

micheles:
Blocked until somebody gives me an example exhibiting the problem.

raoanirudh:
> .. an example exhibiting the problem.

[event_based_risk.zip](https://github.com/gem/oq-engine/files/2940537/event_based_risk.zip)

PR #4514, opened 2019-03-07
Raise an error for missing occupants, early

micheles:
Closes https://github.com/gem/oq-engine/issues/4176.


openquake/commonlib/readinput.py
```diff
@@ -962,13 +955,18 @@ def get_sitecol_assetcol(oqparam, haz_sitecol=None, cost_types=()):
...
+        exposure, assets_by_site, oqparam.time_event)
+    if assetcol.occupancy_periods:
+        missing = set(cost_types) - set(exposure.cost_types['name']) - set(
+            ['occupants'])
+    else:
+        missing = set(cost_types) - set(exposure.cost_types['name'])
+    if missing and not oqparam.calculation_mode.endswith('damage'):
+        expo = oqparam.inputs.get('exposure', '')
+        raise InvalidFile(
+            'Expected cost types %s but the exposure %r contains %s' % (
```
raoanirudh: Wouldn't it be better to point out the missing data in the exposure directly? Otherwise we could end up with a message like the following: 
```sh
Expected cost types ['occupants', 'structural'] but the exposure ['exposure_model.xml']
contains ['business_interruption' 'contents' 'nonstructural' 'structural']
```

micheles: Done

Issue #2801, opened 2017-05-16
Restore the validation of the logic tree


micheles:
By looking back at the engine history it seems that this has been disabled from the beginning of time (i.e. since before engine 1.0). It is not clear why. Perhaps it would be good to enable it in `readinput.get_source_model_lt`. It certainly is nontrivial, because the validation code assume NRML 0.4 while now we have NRML 0.5 to support too. Moreover, perhaps it was disabled because not working. I have no recollection of why it was disabled: perhaps it predates my coming to GEM or perhaps I have totally forgotten.

NB: one must test the situation described by M. Pagani in https://github.com/gem/oq-engine/pull/2800

Issue #4503, opened 2019-02-27
applyToSources for a non-existing source should be an error


micheles:
If you have a source model logic three with more than one branch and the source belongs to one source model branch but not to the others, setting applyToSources must give an error unless applyToBranches is specified and points to the right branch, i.e. the one containing the source.

micheles:
@mmpagani says that we should force the user to specify `applyToBranches` if there is more than one source model and `applyToSources` is used.

PR #4517, opened 2019-03-08
Restored logic tree validation

micheles:
After two years, closes https://github.com/gem/oq-engine/issues/2801. This is another step towards https://github.com/gem/oq-engine/issues/4503.


micheles:
LGTM

Issue #4486, opened 2019-02-22
When can we remove the XML hazard outputs?


micheles:
It is really annoying to maintain them.

micheles:
Not yet.

PR #4524, opened 2019-03-09
Deprecated the hazard XML exporters

micheles:
See https://github.com/gem/oq-engine/issues/4486. People are still using the XML exporters, so we cannot remove them, but we can at least deprecate them so that people will get annoying warnings and will know that there are better exporters to use. Still, I do not plan to remove them any time soon.


mmpagani:
LGTM

Issue #4426, opened 2019-02-05
Assign GSIM weight by IMT when GSIM does not support certain IMTs


rcgee:
We are interested in using the new feature "Implemented weights by IMT" (see https://github.com/gem/oq-engine/pull/4300) for the ground motion logic trees of some models

We need to use this new feature for the following (hypothetical) case:

GMPE 1: PGA [1.0], SA [0]
GMPE 2: PGA [0], SA [1.0]

However, we get an error when we try to assign a weight of 0 if the corresponding GMPE does not support that intensity measure type (which is often the reason why we want to assign a weight of 0!)

For example: "ValueError: The IMT SA is not accepted by the GMPE 1"

micheles:
This is not strange, the case of weight=0 has to implemented specially. You need to give me the files you are using.

kejohnso:
[in.zip](https://github.com/gem/oq-engine/files/2832850/in.zip)

Currently, we are using three separate job and gmmLT files, one for each set of IMTs. We want to combine everything into one job file (job.ini in the zip folder) which uses a single gmmLT (gmmLT.xml in the zip folder)

micheles:
This is surprisingly hard to implement and will take some time.

micheles:
@rcgee and @kejohnso, can you use a MultiGMPE for your use case? See
https://docs.openquake.org/oq-engine/advanced/parametric-gmpes.html#multigmpe

kejohnso:
I gave it a try, but I think no, (at least not in a straightforward way).
It seems that using multiGMPE, the GMPEs can be different for different
IMTs, but the weights for each logic tree branch must still be the same.
Even if each branch is given only a single IMT to allow for differing
weights (e.g., three for TRT-1 and four for TRT-2), then the branch set
sums to > 1.0 and the job fails with an error. I also tried using two
different branch sets, but there is an error here because they are applied
to the same TRT. The workaround I could test would be to duplicate the
sources and assign them different TRTs (e.g., TRT-SA0.1 and TRT-PGA), which
could allow us to at least run the full calculation using one job file, but
I don't think this is the optimal solution, especially where source files
are large.

PR #4529, opened 2019-03-11
Zero weight

micheles:
An attempt to solve the issue https://github.com/gem/oq-engine/issues/4426.


micheles:
Superseded by https://github.com/gem/oq-engine/pull/4673

Issue #4571, opened 2019-03-27
Remove insured_losses


micheles:
So that they can be re-added with a better implementation.

micheles:
The moment to do this work is NOW, because it allows a big simplification of risk calculators and it makes implementing multi-risk easier.

PR #4573, opened 2019-03-27
Removed insured losses

micheles:
Closes https://github.com/gem/oq-engine/issues/4571


raoanirudh:
@micheles, this change needs to be reflected in the manual

PR #4579, opened 2019-03-29
Leveraging on the fact that each riskinput contains a single site

micheles:
This allows some simplification in the risk calculators.


micheles:
Superseded by https://github.com/gem/oq-engine/pull/4580/

PR #4581, opened 2019-03-29
Updated bt03

guyomd:
Updating of Berge-Thierry (2003) GSIM and addition of several alternatives for use with Mw.


micheles:
Guillaume, please update the file doc/sphinx/openquake.hazardlib.gsim.rst by adding the new module you introduced, then run the test openquake/hazardlib/tests/gsim/utils_test.py. If it is green, the documentation will be complete.

guyomd:
My impression is that I don't need to update the content of file "doc/sphinx/openquake.hazardlib.gsim.rst" as I did not add any module. However, I added classes to the existing "berge_thierry_2003.py" module.
I runned the "test_utils.py". No output, no error produced. I interpret this as a successful verification (?).
Thank you.

micheles:
Ah, sorry, you are right. Then LGTM for me too.

PR #4593, opened 2019-04-02
Run multirisk test with the demos

micheles:


daniviga:
You are using relative paths. This is breaking how the script works. See https://github.com/gem/oq-engine-celery/blob/master/.travis.yml#L30

micheles:
Fixed here: https://github.com/gem/oq-engine/commit/e35be62cc54e648673f8dcd0aeaa2b32edbd9e60

PR #4596, opened 2019-04-02
Changed setproctitle to display the name of the command used (i.e. oq1, oq2)

micheles:
This should work, except when `setproctitle` does not change the process title (I have seen this happening for unknown reasons).


micheles:
Let's close this since even if I fix the issues it is still not optimal.

PR #4603, opened 2019-04-03
Implement a simple way to limit number of concurrent jobs

daniviga:
This is a proof of concept of a simple queuing system to allow a limited number of jobs at the same time.

- Instead of implementing this via wrappers, bash, magic or calendars it's directly in the Engine logic, this feature has been requested also by external users in the past 
- Limit can be configured in `openquake.cfg`
- Job is put on hold (but submitted and validated) until the queue is consumed
- Order of the jobs waiting in `submitted` state is maintained
- A new state `submitted` is introduced, which is different from `created` (i.e. currently a job may refuse to start  due an error in the input file and left in `created` state stalling the queue)

The same logic works and is applied for both the cli and the webui, also when those are mixed

### Possible issues

- Jobs killed by the OOM killer are left in `executing` state, so they may stall the queue
- We must check that the dbserver reset all non `complete` jobs to failed; if it is not already done we must implement it, so a dbserver restart cleans the queue. We may also add an `oq db` command.


openquake/engine/engine.py
```diff
@@ -302,6 +302,24 @@ def run_calc(job_id, oqparam, exports, hazard_calculation_id=None, **kw):
...
+        if max_concurrent_jobs > 0:
+            executing_jobs = logs.dbcmd('get_jobs_by_status', 'executing')
+            submitted_jobs = logs.dbcmd('get_jobs_by_status', 'submitted')
+            if len(executing_jobs) + len(submitted_jobs) > max_concurrent_jobs:
+                logs.LOG.warn('wait for other jobs in queue to finish')
+
+                while (len(executing_jobs) > max_concurrent_jobs - 1 or
+                       (min(submitted_jobs) < job_id
+                        if submitted_jobs else False)):
+                    time.sleep(5)
```
daniviga: This may be increased to reduce the pressure on the dbserver

daniviga:
> * Jobs killed by the OOM killer are left in `executing` state, so they may stall the queue
> 
> * We must check that the dbserver reset all non `complete` jobs to failed; if it is not already done we must implement it, so a dbserver restart cleans the queue. We may also add an `oq db` command.

Queue will not stall also in such cases since we check that the `pid` is actually existing on the system, so if a calculation exists in `executing` but the associated `pid` does not it will not be counted

daniviga:
>     * We must check that the dbserver reset all non `complete` jobs to failed; if it is not already done we must implement it, so a dbserver restart cleans the queue. We may also add an `oq db` command.

See #4606

PR #4609, opened 2019-04-04
Imported only intensities > 0

micheles:
As requested by @CatalinaYepes


CatalinaYepes:
Some issues with the current outputs:
- The `asset_id` column is missing (we don't need it, but I if we include it we will be consistent with the output formats from other calculators)
- In the columns `taxonomy` and `tags`, where frequently the values are strings, the output file is presenting the values as integers instead of the corresponding string.
- The order of the columns is sorted based on the column headers. However I will prefer to have a more standard format. For example:
[`exposure_columns`, `tags`, `site_id`, binary_values, ash_values]

PR #4611, opened 2019-04-05
Switch Travis to Xenial, which will allow us to use Py 3.7 in the future

daniviga:
Xenial is now official supported by Travis and will become the default soon.

Let's see if we are already compatible with it. It also support Python 3.7.


daniviga:
Travis is green, I will merge it

PR #4616, opened 2019-04-10
Made it possible to run arbitrary queries on the database

micheles:
Thus bypassing the actions. Used the new feature to remove a couple of actions. Probably more will be removed  on the future, depending on the occasion.

### DIFF
--- a/openquake/server/db/actions.py
+++ b/openquake/server/db/actions.py
@@ -675,53 +675,6 @@ class List(list):
     _fields = ()
 
 
-def get_jobs_by_status(db, status):
-    """
-    :param db:
-        a :class:`openquake.server.dbapi.Db` instance
-    :param status:
-        status string
-    :returns:
-        list with id
-    """
-    fields = 'id,pid,is_running'
-    job_id = []
-
-    query = ('''-- executing jobs
-SELECT %s FROM job WHERE status=?x ORDER BY id desc''' % fields)
-    rows = db(query, status)
-    for r in rows:
-        if status != 'submitted' and status != 'executing':
-            job_id.append(r.id)
-        # if r.pid is 0 it means that such information
-        # is not available in the database
-        elif r.is_running and r.pid and psutil.pid_exists(r.pid):
-            job_id.append(r.id)
-    return job_id
-
-
-def get_executing_jobs(db):
-    """
-    :param db:
-        a :class:`openquake.server.dbapi.Db` instance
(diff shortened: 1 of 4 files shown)


openquake/server/db/actions.py
```diff
@@ -675,53 +675,6 @@ class List(list):
...
-        if status != 'submitted' and status != 'executing':
-            job_id.append(r.id)
-        # if r.pid is 0 it means that such information
-        # is not available in the database
-        elif r.is_running and r.pid and psutil.pid_exists(r.pid):
-            job_id.append(r.id)
-    return job_id
-
-
-def get_executing_jobs(db):
```
daniviga: This is needed by `cluster-status` which is now broken. Can we add it back?

daniviga: This is needed by `cluster-jobs` which is now broken. Can we add it back?

daniviga: This is needed by `cluster-jobs` which is now broken. Can we add it back?

raoanirudh: Also jobs now fail with the following error:

```sh
  File "/opt/openquake/bin/oq", line 11, in <module>
    load_entry_point('openquake.engine', 'console_scripts', 'oq')()
  File "/opt/openquake/oq-engine/openquake/commands/__main__.py", line 50, in oq
    parser.callfunc()
  File "/opt/openquake/oq-engine/openquake/baselib/sap.py", line 181, in callfunc
    return self.func(**vars(namespace))
  File "/opt/openquake/oq-engine/openquake/baselib/sap.py", line 251, in main
    return func(**kw)
  File "/opt/openquake/oq-engine/openquake/commands/db.py", line 61, in db
    res = logs.dbcmd(cmd, *convert(args))
  File "/opt/openquake/oq-engine/openquake/commonlib/logs.py", line 55, in dbcmd
    return res.get()
  File "/opt/openquake/oq-engine/openquake/baselib/parallel.py", line 348, in get
    raise etype(msg)
sqlite3.OperationalError:
  File "/opt/openquake/oq-engine/openquake/baselib/parallel.py", line 358, in new
    val = func(*args)
  File "/opt/openquake/oq-engine/openquake/server/dbapi.py", line 335, in __call__
    raise exc.__class__('%s: %s %s' % (exc, templ, args))
OperationalError: near "get_executing_jobs": syntax error: get_executing_jobs ()
```

PR #4618, opened 2019-04-10
Add support for most_likely_damage

micheles:


raoanirudh:
I'd suggest calling it `modal_damage_state` instead of `most_likely_damage`

micheles:
More work like adding a field with the total number of buildings is left for a future PR.

openquake/calculators/export/risk.py
```diff
@@ -437,16 +437,32 @@ def export_damages_csv(ekey, dstore):
     return writer.getsaved()
 
 
+def modal_damage_array(data, damage_dt):
+    # determine the damage state with the highest probability
+    A, L, MS, D = data.shape
+    dmgstate = damage_dt['structural'].names
```
raoanirudh, line 443: For calculations where `number_of_ground_motion_fields > 1`,

`dmgstate` will contain something like ('no_damage_mean', 'no_damage_stdv', 'slight_mean', 'slight_stdv', 'moderate_mean', 'moderate_stdv', 'extreme_mean', 'extreme_stdv', 'complete_mean', 'complete_stdv'). 

Instead, it should be the names of the damage states, i.e., ('no_damage', 'slight', 'moderate', 'extreme', 'complete').

PR #4627, opened 2019-04-12
More refactoring of the epsilons

micheles:
Removed ~50 lines. There is a major change of logic: now the epsilons are stored in the cache file and not sent to the workers via rabbitmq.
This is not enough to implement asset correlation in ebrisk, but it is a step forward.


raoanirudh:
Supporting `asset_correlation` would be a lower priority than simply enabling non-zero covs for vulnerability functions using the LN or BT distributions. For uncorrelated sampling of the loss ratios (i.e., the default case, when `asset_correlation = 0`), building the correlation matrix is not required.

micheles:
In event based we are not building the correlation matrix, the cases asset_correlation=0 and asset_correlation=1 are treated specially. For asset_correlation=0 the epsilons are build with the class EpsilonMatrix0: they still need to be generated on the master and transferred to the worker. The difficult part is to get the indices correctly, otherwise epsilons would be mixed between events and we would get different numbers than in event_based_risk.

PR #4631, opened 2019-04-15
Restored the old sampling logic in event based

micheles:
After the changes in https://github.com/gem/oq-engine/pull/3187/ and https://github.com/gem/oq-engine/pull/3191/ the convergency of the hazard curves was lost in event based with sampling :-(


micheles:
It turns out the convergency is still no particularly good, but the scientists like more the approach used in this branch.

PR #4408, opened 2019-02-01
Christchurch gsims

cvanhoutte:
Hi again

I have implemented the Canterbury-specific gsims required for running the Canterbury Seismic Hazard Model. I am preparing comprehensive documentation in a report that is almost completed. Will you need to see this documentation before a branch like this can be merged?

Unfortunately, none of the authors of the gsims implemented here were able to make independent code available for testing purposes, so all changes are non-verified.

Let me know if there are any questions.
Chris

PR #4656, opened 2019-04-30
New gsim class for Christchurch

cvanhoutte:
Hi there

As a follow-on from a previous pull request (#4408), I've added a new class to bradley_2013b.py, and modified a previous test to include this new class. 

This class is structured to use different predictions based on site location. We have documentation on why this has been necessary for the implementation of the Canterbury SHM
https://doi.org/10.21420/1AEM-PZ85

Let me know if you have any questions/comments/suggestions.

Thanks
Chris


openquake/hazardlib/gsim/bradley_2013b.py
```diff
@@ -476,3 +479,207 @@ def _compute_additional_sigma(self):
...
+        in the Christchurch District Plan. See Figure 4.4 of Van Houtte and
+        Abbott (2019).
+        """
+        polygon = shapely.geometry.Polygon(
+            [(172.6259, -43.5209), (172.6505, -43.5209),
+             (172.6505, -43.5399), (172.6124, -43.5400),
+             (172.6123, -43.5289), (172.6124, -43.5245),
+             (172.6220, -43.5233)]
+        )
+        points = [shapely.geometry.Point(lons[ind], lats[ind])
```
g-weatherill: For most applications it is not a good idea to be doing point-in-polygon checks for Earth geometry using geodetic coordinates, as the edge of a polygon should be a great arc and not a line in long/lat. In this case though, the area is very small so maybe the error is tolerable given the overhead in re-projection.

cvanhoutte: Agreed. The other point-in-polygon function (_check_in_cshm_polygon) was how the original Canterbury model was defined, but I don't see why that can't be changed. More than happy to amend. 

Is there anything else that jumps out?

g-weatherill: Sorry, it's a different reviewer this time around! Everything else looks good to me.

micheles:
@cvanhoutte please solve the conflicts

micheles:
It turns out I was able to solve the conflicts myself.

Issue #4584, opened 2019-03-30
Implement GMPEs for SERA Project


g-weatherill:
Several new GMPEs will be required for the SERA project (new European seismic hazard model). Mostly these adopt a "backbone" approach, taking a core GMPE(s) and applying adjustment factors that are defined by the user in the GSIM Logic tree and passed as arguments to the function. Several stages to the work:

- [x] Implement shallow crustal GMPE and adjustment factors

- [x] Implement subduction/Vrancea GMPE and adjustment factors

- [x] Add support for defining categorical site parameters (e.g. EC8/NEHRP site class, geology etc.)

- [x] Add MetaGSIM for implementing amplification using (design) code-based factors

- [x] Add experimental nonlinear amplification functions

- [x] Add NGA East site amplification model

- [x] Implement the cratonic GMPE

g-weatherill:
It is unclear if the final task of implementing geology-based site amplification for the subduction and craton GMPEs will be needed. So that task is removed now and this feature can be considered complete for the time-being and the issue closed. New issues will be raised in the future if further changes/features are required.

PR #4664, opened 2019-05-02
Adds Adjustment of BC Hydro GMPE for Europe (SERA)

g-weatherill:
Addresses point No. 2 of https://github.com/gem/oq-engine/issues/4584

For the subduction and Vrancea component of the SERA seismic hazard model a Mediterranean calibration of the Abrahamson et al. (2016) "BC Hydro" model has been created. In addition to preserving the epistemic scaling factors at high magnitudes from the original model, the new GMPEs add keyword arguments to allow the user to calibrate the anelastic attenuation coefficient (theta 6) and the statistical uncertainty (sigma mu).


micheles:
It would be nice to have also a test engine-side, i.e. like the tests in openquake/qa_tests_data/classical/

openquake/hazardlib/gsim/bchydro_epistemic.py
```diff
@@ -0,0 +1,450 @@
+# -*- coding: utf-8 -*-
```
mmpagani: The name of this file does not comply with the usual standard we used. Are you planning to change it once the GMPE will be in its final form?

debian/changelog
```diff
@@ -14,6 +17,10 @@
   * Multiple DbServer ZMQ connections are restored to avoid errors under heavy
     load and/or on slower machines
 
+<<<<<<< HEAD
+>>>>>>> master
+=======
+>>>>>>> 87b5682a07cd902a543271ff1d72ce997b7d9fc7
```
micheles: Ops!

openquake/hazardlib/gsim/bchydro_epistemic.py
g-weatherill: I had not planned to because this is a set of adjustments to an existing GMPE and not a new model with a specific publication in itself. But I should change to `bchydro_2016_epistemic.py` to clarify this applies to the 2016 BC Hydro model and not subsequent BC Hydro models.

PR #4665, opened 2019-05-02
Revert #4475

### LINKED ISSUE #4650: Connection issues from the WebUI and the QGIS plugin
Cata pointed out that on her mac she frequently has connection issues, with error messages like these (from the IRMT plugin and from the WebUI):

```python
ERROR Internal Server Error: /v1/calc/list
  File "/Users/cyepes/py36/lib/python3.6/site-packages/django/core/handlers/exception.py", line 35, in inner
    response = get_response(request)
  File "/Users/cyepes/py36/lib/python3.6/site-packages/django/core/handlers/base.py", line 128, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/Users/cyepes/py36/lib/python3.6/site-packages/django/core/handlers/base.py", line 126, in _get_response
    response = wrapped_callback(request, callback_args, *callback_kwargs)
  File "/Users/cyepes/py36/lib/python3.6/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, args, *kwargs)
  File "/Users/cyepes/openquake/oq-engine/openquake/server/views.py", line 99, in wrap
    response = func(request, args, *kwargs)
  File "/Users/cyepes/openquake/oq-engine/openquake/server/views.py", line 310, in calc_list
    utils.get_acl_on(request), id)
  File "/Users/cyepes/openquake/oq-engine/openquake/commonlib/logs.py", line 53, in dbcmd
    res = sock.send((action,) + args)
  File "/Users/cyepes/openquake/oq-engine/openquake/baselib/zeromq.py", line 147, in send
    self.zsocket.send_pyobj(obj)
  File "/Users/cyepes/py36/lib/python3.6/site-packages/zmq/sugar/socket.py", line 602, in send_pyobj
    return self.send(msg, flags=flags, **kwargs)
  File "/Users/cyepes/py36/lib/python3.6/site-packages/zmq/sugar/socket.py", line 391, in send
    return super(Socket, self).send(data, flags=flags, copy=copy, track=track)
  File "zmq/backend/cython/socket.pyx", line 727, in zmq.backend.cython.socket.Socket.send
  File "zmq/backend/cython/socket.pyx", line 774, in zmq.backend.cython.socket.Socket.send
  File "zmq/backend/cython/socket.pyx", line 249, in zmq.backend.cython.socket._send_copy
  File "zmq/backend/cython/socket.pyx", line 244, in zmq.backend.cython.socket._send_copy
  File "zmq/backend/cython/checkrc.pxd", line 25, in zmq.backend.cython.checkrc._check_rc
zmq.error.ZMQError: Operation cannot be accomplished in current state
INFO "GET /v1/calc/list?relevant=true HTTP/1.1" 200 4618
```

```python
INFO "GET /v1/calc/list?relevant=true HTTP/1.1" 200 4618
Resource temporarily unavailable (bundled/zeromq/src/signaler.cpp:301)
```

Something similar happened also to me, perhaps more frequently while using a virtual machine.

The connection usually comes back after a short while. While you use the IRMT plugin, it can be quite annoying, because the plugin closes the dialog that drives the engine when it happens, and it opens the settings to give the user the possibility to change the connection settings. I am changing the behavior of the plugin to be more tolerant when the connection fails, but without hiding the problem.

daniviga:
Revert "Keeping open the DbServer client socket [skip hazardlib][demos]"

This reverts commit a483d58.

Closes #4650 since it does not work under heavy loads and/or on slower machine

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,6 +1,6 @@
   [Daniele Viganò]
-  * Multiple DbServer ZMQ connection are restored to avoid errors under heavy load
-    on slower machines
+  * Multiple DbServer ZMQ connections are restored to avoid errors under heavy load
+    and/or on slower machines
 
   [Michele Simionato]
   * Removed the ugly registration of custom signals at import time: now they
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,5 +1,9 @@
+  [Daniele Viganò]
+  * Multiple DbServer ZMQ connection are restored to avoid errors under heavy load
+    on slower machines
+
   [Michele Simionato]
-  * Remove the ugly registration of custom signals at import time: now they
+  * Removed the ugly registration of custom signals at import time: now they
     are registered only if `engine.run_calc` is called
   * Removed the dependency from rtree
   * Removed all calls to ProcessPool.shutdown to speed up the tests and to
--- a/openquake/commonlib/logs.py
+++ b/openquake/commonlib/logs.py
@@ -34,8 +34,6 @@ LOG = logging.getLogger()
 
 DBSERVER_PORT = int(os.environ.get('OQ_DBSERVER_PORT') or config.dbserver.port)
 
-sock = None
-
 
 def dbcmd(action, *args):
     """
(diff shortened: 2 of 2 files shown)


debian/changelog
```diff
@@ -1,5 +1,9 @@
+  [Daniele Viganò]
+  * Multiple DbServer ZMQ connection are restored to avoid errors under heavy load
+    on slower machines
```
micheles: connection -> connections

PR #4669, opened 2019-05-03
Fixed calculations using NBCC2015_AA13 gsims

micheles:
This is making life hard for @kejohnso


kejohnso:
It works correctly now! LGTM

PR #4670, opened 2019-05-03
Send SIGINT in oq abort instead of SIGTERM

daniviga:
`SIGTERM` must be trapped, otherwise the `except` here https://github.com/gem/oq-engine/blob/ebdc52d2aba2e449e4b8a7bac7cdbb87e44d25ce/openquake/engine/engine.py#L349 is not processed ad computations are left in `executing` state.

With this PR when a job is `ctrl-c`'ed or aborted it is marked as 'aborted'; if it's terminated by a `SIGTERM` will be marked as 'failed'.

### DIFF
--- a/openquake/engine/engine.py
+++ b/openquake/engine/engine.py
@@ -199,7 +199,7 @@ def manage_signals(signum, _stack):
         raise MasterKilled('The openquake master process was killed manually')
 
     if signum == signal.SIGTERM:
-        raise SystemExit('SIGTERM')
+        raise SystemExit('Terminated')
 
     if hasattr(signal, 'SIGHUP'):  # there is no SIGHUP on Windows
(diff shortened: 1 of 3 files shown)


openquake/engine/engine.py
```diff
@@ -199,7 +199,7 @@ def manage_signals(signum, _stack):
         raise MasterKilled('The openquake master process was killed manually')
 
     if signum == signal.SIGTERM:
-        raise SystemExit('SIGTERM')
```
daniviga: `Terminated` is more consistent with the default unix behavior

openquake/engine/engine.py
```diff
@@ -199,7 +199,7 @@ def manage_signals(signum, _stack):
...
+        raise SystemExit('Terminated')
```
daniviga: `Terminated` is more consistent with the default unix behavior

PR #4673, opened 2019-05-05
Manage zero weights for IMTs that should be discarded

micheles:
Closes https://github.com/gem/oq-engine/issues/4426.

NB: this is probably wrong, since the feature is surprisingly difficult to implement. A lot of testing is needed, so I am scheduling this pull request for engine 3.6.


kejohnso:
@micheles I have added to the test as you requested, and in principle it works (I get the expected results when run as an oq job) but I get an error when running the test:

openquake.calculators.tests.DifferentFiles: /Users/kjohnson/GEM/oq-engine/openquake/qa_tests_data/classical/case_39/expected/hazard_curve-mean-PGA.csv /var/folders/2x/tpbvfjbs41q1hm8_47mf9bsw0000gn/T/tmp8lq1uwnm/hazard_curve-mean-PGA_669.csv

It seems the strip_id function isn't working correctly, maybe? Do you have an idea of how to fix this?

micheles:
You should push the modifications you are using, @kejohnso

PR #4674, opened 2019-05-06
Removed insured_losses from the manual

micheles:
If in the future we will restore the insured losses functionality (likely in a different way than it is now) we will re-add the documentation.


micheles:
We should not remove the attributes related to insured losses from the test files because we must make sure that an exposure that in the past worked, still work. Of course the insured losses will not be computed,  but the regular losses will. Instead we must remove the insurance parameters from the documentation and the demos, to avoid confusing the users. I am doing that right now.

PR #4698, opened 2019-05-10
Lanzano and Luzi (2019)

giovannilanzanoINGV:
Upload files for Lanzano and Luzi (2019) GMPE for volcanic zones in Italy


openquake/hazardlib/tests/gsim/data/LL19/prova
```diff
@@ -0,0 +1 @@
+
```
mmpagani: This file should be removed

micheles: I will remove it

Issue #4707, opened 2019-05-14
Windows charset encoding issue reading source model in logic tree (0x81)


pslh:
Running a model for Dom. Rep using oq-engine 3.5 on Windows both myself and Catarina see the following error message:

 File "c:\program files\openquake engine\python3.6\lib\site-packages\openquake\commonlib\logictree.py", line 920, in validate_uncertainty_value
    raise LogicTreeError(node, self.filename, str(exc)) from exc
openquake.commonlib.logictree.LogicTreeError: filename '...\full_model_OQ_DomRep\source_model_logic_tree.xml', line 9: 'charmap' codec can't decode byte 0x81 in position 214377: character maps to <undefined>

The same model runs without problem on Linux.

PR #4708, opened 2019-05-14
Use utf-8 encoding to open source model - avoid charset 0x81 problem …

pslh:
…on windows

Fixes issue #4707 on my Windows 10 installation of oq-engine 3.5. 
@micheles  I suspect there may be other places where we need to specify encoding with open() et al.

### DIFF
--- a/openquake/commonlib/logictree.py
+++ b/openquake/commonlib/logictree.py
@@ -1178,7 +1178,8 @@ class SourceModelLogicTree(object):
                 branch.child_branchset = branchset
 
     def _get_source_model(self, source_model_file):
-        return open(os.path.join(self.basepath, source_model_file))
+        return open(os.path.join(self.basepath, source_model_file),
+					encoding='utf-8')
 
     def collect_source_model_data(self, branch_id, source_model):
         """


micheles:
This happened already in the past (to Yufang, it was linux but probably their locale was not set to UTF8 or something like that) and at the time I did an effort to look at the various `open` in the code base related to XML files and changed the encoding to UTF8, but this particular occurrence was left out, there are just too many :-( I can probably figure out a way to discover all the relevant opens and raise an error in the tests if a file name ending with .xml is opened without specifing the encoding.
Thinking about that,.

micheles:
I will cherry-pick the fix in engine-3.5 too.

micheles:
Solved in https://github.com/gem/oq-engine/pull/4710: there were no other occurrencies of the issue, it was the only one.

openquake/commonlib/logictree.py
```diff
@@ -1178,7 +1178,8 @@ def apply_branchset(self, branchset_node, branchset):
                 branch.child_branchset = branchset
 
     def _get_source_model(self, source_model_file):
-        return open(os.path.join(self.basepath, source_model_file))
+        return open(os.path.join(self.basepath, source_model_file),
+					encoding='utf-8')
```
daniviga: The indentation is wrong here

micheles: There are tabs, but I have fixed them after merging.

PR #4722, opened 2019-05-17
Accepted list-like syntax for return_periods

micheles:
Apparently the manual and IPT use this syntax (which was never meant to work).

### DIFF
--- a/openquake/hazardlib/valid.py
+++ b/openquake/hazardlib/valid.py
@@ -533,7 +533,7 @@ def positivefloats(value):
     :returns:
         a list of positive floats
     """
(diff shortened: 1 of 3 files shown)


openquake/hazardlib/valid.py
```diff
@@ -533,7 +533,8 @@ def positivefloats(value):
     :returns:
         a list of positive floats
     """
-    floats = list(map(positivefloat, value.split()))
+    values = value.lstrip('[').rstrip(']').split()
```
raoanirudh: `value.strip('[]').split()` should work too

micheles: Indeed

PR #4735, opened 2019-05-22
Made the context arrays read-only

micheles:
This should solve forever the long standing and tricky mutability bugs.


rcgee:
I plotted a pair of GMPEs that were showing problems before, specifically `CauzziFaccioli2008` and `BindiEtAl2014Rhyp`, and I no longer see the problem. Previously, when plotted together, `BindiEtAl2014Rhyp` would use (incorrectly) the updated dctx generated by `cauzzi_faccioli_2008.py`. This caused `BindiEtAl2014Rhyp` to predict constant values of ground motion for distances <15km (the clipping distance specified in `cauzzi_faccioli_2008.py`). Since the distance context can no longer being overwritten, `BindiEtAl2014Rhyp` no longer predicts incorrect values when `CauzziFaccioli2008` is called before it.

LGTM

PR #4738, opened 2019-05-23
Track the work to update deps

daniviga:
See also the change in https://github.com/GEMScienceTools/gmpe-smtk/pull/106.


micheles:
I have updated numpy and scipy to the latest versions. I would keep h5py as it is, since it would require a lot of changes on our part.

PR #4746, opened 2019-05-27
Fix oq webui start --noreload argument

daniviga:
which was not passed to the popen and thus was causing issues with children


daniviga:
Backported to `engine-3.5`

PR #4753, opened 2019-05-28
Fixed regression in gsim parser

micheles:
```xml
<uncertaintyModel gmpe_table="../gm_tables/Woffshore_low_clC.hdf5">
    GMPETable
</uncertaintyModel>```

is now valid again.


micheles:
Backported to engine-3.5

PR #4763, opened 2019-05-30
added sitewriter to write collection of sites to nrml file

abhineetgupta:


micheles:
We thank you for your pull request, however since the site model can be described in CSV format nowadays, there is no much point in an XML writer for it. You can see an example of site model in CSV in our event based risk demo:
https://github.com/gem/oq-engine/blob/master/demos/risk/EventBasedRisk/site_model.csv
risk/EventBasedRisk/site_model.csv

abhineetgupta:
Dear Michele

Thank you for letting me know. This is very useful feature but is not
provided anywhere in the OpenQuake Engine User Manuals
<https://docs.openquake.org/manuals/OpenQuake%20Manual%203.5.pdf>. The only
file-based options that are shown in the documentation are either using the
sites_csv parameter that seems to accept a different format csv, or a
site_model_file parameter that only accepts xml files. As a result, it was
necessary to provide functionality to write a site NRML. It's unfortunate
that this is not clearer in the manual, and I am sure other researchers
have encountered the problem of creating a valid Sites NRML file. Hence, I
believe that the sitewriter will be useful to them.

Thanks
Abhineet

micheles:
I have changed the manual by making clear that the CSV format is the preferred one (see https://github.com/gem/oq-engine/pull/4772) and I am going to close this PR.

PR #4771, opened 2019-06-03
Move later the check on the number of realizations

micheles:
This was requested by @raoanirudh but we have to check how it works in practice.


raoanirudh:
The current behavior is that when using full enumeration in an event based calculation with a logic tree that has more than 100 terminal branches, the engine stops the job with a `Value Error: There are too many potential logic tree paths, use sampling instead of full enumeration`.

However, this also stops calculations involving full enumeration where the actual number of "effective realizations" is far below 100.

micheles:
We decided with Anirudh to keep the current behavior. I am changing the error message to say
```
There are too many potential logic tree paths (XXX): use sampling instead of full enumeration or reduce the source model with oq reduce_sm
```

PR #4775, opened 2019-06-03
Check engine versions on the nodes

micheles:
As discussed with Paul


daniviga:
Didn't we diecided to make it a command because it was slowing down the startup of calculations on a cluster?

micheles:
Yes, I just forgot to close it.

Issue #4630, opened 2019-04-15
Allow a local file path to be passed to /v1/calc/run


daniviga:
We should add the ability to pass a local path relative to the server to the API.

This will allow the WebUI to execute files (zip or ini) that are already existing on the server, without having users to upload files every time. This is very convenient for use cases like our cluster.

The WebUI will need also to be adapted: 'Run a calculation' should now open a modal with two options. One option would be a zip upload and a second option a text field where to put the file path relative to the server running the webui

micheles:
I like more the idea of writing in the job.ini file something like `source_model_logic_tree_file = ftp:///cluster/dir/ssmLT.xml` or `source_model_logic_tree_file = http:///cluster/dir/ssmLT.xml` which is already possible.

Issue #4648, opened 2019-04-24
Add an oq engine --submit


daniviga:
Now that we have a queue it would be good to have an `oq engine --submit` which behaves as `--run` but it just send the job and then detach from the terminal.

micheles:
It is not so urgent given that we already have `utils/postzip`, so I am closing for the moment.

PR #4781, opened 2019-06-06
Added a command `oq submit`

micheles:
Closes https://github.com/gem/oq-engine/issues/4630 and https://github.com/gem/oq-engine/issues/4648.

### DIFF
--- a/openquake/commands/submit.py
+++ b/openquake/commands/submit.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+# vim: tabstop=4 shiftwidth=4 softtabstop=4
+#
+# Copyright (C) 2019 GEM Foundation
+#
+# OpenQuake is free software: you can redistribute it and/or modify it
+# under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# OpenQuake is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


daniviga:
Hippie! Wouldn't it better to have an `oq engine --submit` ? It looks to me syntactically more correct

micheles:
But it would be too difficult to implement, no way. Also think of the difference in semantic: `oq submit` assumes there is a WebUI already active, `oq engine --run` does not.

daniviga:
I see you are using the WebUI: how does it play when authentication is enabled? (Like on a cluster)

openquake/commands/submit.py
```diff
@@ -0,0 +1,37 @@
...
+from openquake.baselib import sap, config
+from openquake.hazardlib import valid
+
+
+@sap.script
+def submit(job_ini, hc=None):
+    """Post a local file to the WebUI"""
+    job_ini = os.path.abspath(job_ini)
+    dic = dict(job_ini=job_ini, hazard_job_id=hc)
+    # NB: there is no WebUI port in openquake.cfg for the moment
```
daniviga: We need to add it it to the `openquake.cfg` but in a 'client' sections: in real cases the 'client' port (i.e. 443) is different from the daemon port (i.e. 8800) because a reverse proxy is used. We also need to specify if we need to make the connection via HTTPS or plain HTTP.

micheles: Can you add that, Daniele?

micheles: An alternative would be to pass on the command line the server URL.
It looks better to me, what do you think?

micheles:
Closing since it is not so important.

PR #4802, opened 2019-06-12
Rename outdated parameter in a job file

raoanirudh:
`complex_rupture_mesh_spacing` is now `complex_fault_mesh_spacing`


micheles:
Merge if the tests are green

daniviga:
Why does a change in a variable in the `job.ini` produces different numbers?

micheles:
Because it is discretization parameter @daniviga , so it is normal that the numbers change.

PR #4819, opened 2019-06-17
Switch packager.sh to pytest

### LINKED ISSUE #4818: packager.sh is still using nose
We have switched to `pytest`, so we should migrate the `packager.sh` and parallelize hazard tests as we did in travis.

daniviga:
Fixes #4818

Coverage has been removed as agreed with @micheles


daniviga:
https://ci.openquake.org/job/zdevel_oq-engine/2837/

PR #4826, opened 2019-06-18
Extended `oq db` to run generic queries

micheles:
This is not much secure (anybody on wilson can use `oq db`, we may change that) but convenient.
Now one can do things like
```
   $ oq db 'SELECT count(*) FROM job WHERE status=?x' complete
   ========
   count(*)
   ========
   205     
   ========
```


daniviga:
A bit risky... could we limit it to `SELECT`? Why we don't use directly `sqlite3`?

```
$ echo 'SELECT count(*) FROM job WHERE status="complete"' | sudo -u openquake sqlite3 /var/lib/openquake/db.sqlite3
[sudo] password for dvigano: 
4672

```

does the same, it's more safe and can be easily converted into a bash alias.

micheles:
+ 1 for restricting this to SELECT unless the username is "openquake"

PR #4834, opened 2019-06-19
Make possible to abort submitted jobs

micheles:
Another thing we forgot.

### DIFF
--- a/openquake/commands/abort.py
+++ b/openquake/commands/abort.py
@@ -32,7 +32,7 @@ def abort(job_id):
     if job is None:
         print('There is no job %d' % job_id)
         return
-    elif job.status not in ('created', 'executing', 'running'):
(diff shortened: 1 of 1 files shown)


openquake/commands/abort.py
```diff
@@ -32,7 +32,7 @@ def abort(job_id):
     if job is None:
         print('There is no job %d' % job_id)
         return
-    elif job.status not in ('created', 'executing', 'running'):
+    elif job.status not in ('created', 'submitted', 'executing', 'running'):
         print('Job %d is %s' % (job.id, job.status))
         return
```
micheles: Perhaps we should have written `if job.status in ('failed', 'aborted')`

daniviga: yes, I think we have only those two states left

micheles: And 'complete'

Issue #4837, opened 2019-06-20
Delete command is not deleting the cache file


daniviga:
When deleting a calculation the corresponding `calc_NNNN.hdf5` is removed (when it's possible), however the `cache_NNNN.hdf5` file is not removed as it should be.

PR #4840, opened 2019-06-21
Removed also the cache file when removing a calculation

micheles:
Should close https://github.com/gem/oq-engine/issues/4837


micheles:
I meant `[skip hazardlib][demos]`!

daniviga:
So then update the changelog and let travis run (with `[skip hazardlib][demos]`)

Issue #4841, opened 2019-06-21
Add checks on the beta distribution parameters for vulnerability


raoanirudh:
If a vulnerability function specifies the Beta distribution for modeling the loss ratios for a set of intensity measure levels (IML), the standard deviation specified for each IML must satisfy the following condition: 

```
0 < σ < sqrt[μ × (1 - μ)]
```

where μ is the mean loss ratio for the IML, and `σ = cov × μ` is the corresponding standard deviation. 

Currently, there are no checks on the parameters specified in the vulnerability functions, so if the provided parameters are outside the above mentioned bounds, the calculation either terminates with `ValueError: a <= 0` if the cov is not positive, or runs to completion but with NaN values for the losses if the cov exceeds the upper bound.

PR #4848, opened 2019-06-25
Added checks for vulnerability functions with the Beta distribution

micheles:
Closes https://github.com/gem/oq-engine/issues/4841

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,6 @@
   [Michele Simionato]
+  * Added an early check on the coefficients of variation and loss ratios of
+    vulnerability functions with the Beta distribution
   * Made sure that `oq engine --dc` removes the HDF5 cache file too
   * Removed the flag `optimize_same_id_sources` because it is useless now
   * Introduced a soft limit at 65,536 sites for event_based calculations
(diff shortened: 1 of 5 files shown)


micheles:
@raoanirudh please note that this PR will break the mosaic. For instance South America has vulnerability functions with zeros in the loss ratios,

micheles:
No, I am wrong, the South America works because the zero ratios are associated to zero CoVs. Please ignored what I said.

micheles:
Actually I was right, the South America is broken, but not due to the zeros. The error is:
```python
ValueError: node vulnerabilityFunction: The coefficient of variation 0.0786413 > 0.07643251879290494 is too large in <VulnerabilityFunction(CR_LFINF-DUL_H4, SA(0.6))>, line 306 of /home/michele/oq-risk-tests/test/event_based_risk/inputs/taxmap/vulnerability_total.xml
```
See https://gitlab.openquake.org/openquake/oq-risk-tests/-/jobs/6033

micheles:
Anirudh suggest to change BT -> LN in the vulnerability functions to disable the check. Will do.

PR #4859, opened 2019-06-28
Adds in finite rupture option to Germany-adjusted Cauzzi and Derras GMPEs

g-weatherill:
In the original PSHA implementation for Germany (made in FRISK) hypocentral distances were used for the GMPEs, with Derras et al. (2014) and Cauzzi et al. (2015) adjusted to their respective finite rupture distances by conversion formulae. However, for the few active faults in the model (only on one branch) the finite rupture distances are used. To avoid requiring a new logic tree branching level for the finite rupture sources, a hack is made onto the adapted GMPE for Germany that will use hypocentral distance when the source is a point, and the correct finite rupture distance in the case that the finite rupture is defined. The check is a simple switch based on the rupture width, using the point approximation if rupture width < 0.001 and the finite distances otherwise.


openquake/hazardlib/gsim/germany_2018.py
```diff
@@ -71,8 +72,13 @@ def _compute_mean(self, C, rup, dists, sites, imt):
         """
         Returns the mean ground motion acceleration and velocity
         """
-        # Convert rhypo to rrup
-        rrup = rhypo_to_rrup(dists.rhypo, rup.mag)
+        print(rup.__dict__, self.REQUIRES_RUPTURE_PARAMETERS)
```
micheles: You forgot a print here

micheles:
Please update the changelog and I will merge

PR #4866, opened 2019-07-02
Restored the grid site model associations

micheles:
As @VSilva asked, the engine is again associating the site model to the hazard grid, if any. However a warning is printed, saying that it would be better to use `oq prepare_site_model` to perform the associations once and for all.


openquake/commonlib/readinput.py
```diff
@@ -420,15 +421,19 @@ def get_site_collection(oqparam):
         except ValueError:
             # this is the normal case
             depth = None
-        sitecol = site.SiteCollection.from_points(
-            sm['lon'], sm['lat'], depth, sm, req_site_params)
-        if oqparam.region_grid_spacing:
-            logging.info('Reducing the grid sites to the site '
-                         'parameters within the grid spacing')
+        if grid_spacing:
+            logging.warning('Having a grid and a site model at the same time '
```
raoanirudh: There's already a similar warning for the case where a grid, a site model, and an exposure model are provided: https://github.com/gem/oq-engine/blob/engine-3.5/openquake/commonlib/oqvalidation.py#L307-L312

Can the two warnings be combined?

Another note: `oq prepare_site_model` works with exposure and sites files, but not yet with a region + region grid spacing. Won't this advice be confusing to the user?

openquake/qa_tests_data/event_based_risk/case_miriam/job2.ini
```diff
@@ -5,7 +5,7 @@ calculation_mode = event_based
 
 [geometry]
 
-region_grid_spacing = 1
+region_grid_spacing = 1.25
 exposure_file = exposure_model.xml
```
raoanirudh: A scenario hazard job (or any other hazard job) with a `region` and `region_grid_spacing` might serve as a better test case for this functionality, rather than a risk job with an exposure and `region_grid_spacing`.

openquake/commonlib/readinput.py
micheles: Damn, I forgot that I had already the warning in place, so I will only keep the original warning.

`oq prepare_site_model` works with a `--grid-spacing` option: if you will pass it, it will grid the convex hull of the exposure.

openquake/qa_tests_data/event_based_risk/case_miriam/job2.ini
micheles: True, but the case with also an exposure is the trickiest one and it deserves to be tested more. Also, most calculations are event based risk calculations, so I want to test this case more than any other.

openquake/commonlib/readinput.py
raoanirudh: Yes, though this case is different, in that there is no exposure. The user has only a `region` (defined by WKT polygon vertices) and a `region_grid_spacing` for which they wish to compute ground motion fields. In addition they have site parameters on a separate regular grid / irregular set of points.

micheles: This case should work already. Can you add a test in oq-risk-tests?

openquake/qa_tests_data/event_based_risk/case_miriam/job2.ini
micheles: And you can always add a test in oq-risk-tests

PR #4873, opened 2019-07-04
move constant b4 into COEFFS table

rcgee:
In the GMPE sharma_2009.py, constant b4 is defined outside of the COEFFS table then added via `coeffs.update(self.CONSTS)`. This results in a key error if the imls specified in the job.ini include periods that require interpolation (e.g. 0.15s and 0.7s)

The key error occurs during the [interpolation between the closest higher and closest lower period ](https://github.com/gem/oq-engine/blob/52a0d4889230bc919a3374fcd5b7e9cad5c2fdc4/openquake/hazardlib/gsim/base.py#L751) because the values of `self.sa_coeffs[max_below]` and `sa_coeffs[min_above]` do not have the same keys (one has b4, the other does not)

In this PR I move constant b4 into the COEFFS table, and there is no longer a key error when using this GMPE with periods that require interpolation.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,7 +1,10 @@
+  [Robin Gee]
+  * Moves b4 constant into COEFFS table for GMPE Sharma et al., 2009 
+
   [Graeme Weatherill]
   * Adds functionality to Cauzzi et al. (2014) and Derras et al. (2014)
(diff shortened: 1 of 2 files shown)


micheles:
LGTM, but please update the changelog

PR #4875, opened 2019-07-04
Added some support for computing the mean MFD from the generated ruptures

micheles:
And a way to compare with the expected MFD. This is preliminary and the plotting works only in the case where there is a single MFD, like in our event based demo which I have enhanced to have 2 realizations with full enumeration, otherwise the trickiness of the multiplicity would not be visible. Here is the plot I get:
mfd


rcgee:
LGTM

PR #4881, opened 2019-07-09
What's New for engine 3.6

micheles:
See https://github.com/gem/oq-engine/blob/whatsnew/doc/whats-new.md


rcgee:
I changed the link to the `minimum_distance` description. It was previously pointing to the  `zero weights` description. All of the links are currently broken, but I'm assuming you are aware of this, and they will work when 3.6 is released. Otherwise LGTM

PR #4882, opened 2019-07-09
Removed extract/agg_curves

micheles:
It is trying to aggregate the individual curves which is wrong. The right way is to extract the aggregate curves generated by the ebrisk calculator. @ptormene should comment out the feature in the plugin.


ptormene:
Here is the plugin-side companion: https://github.com/gem/oq-irmt-qgis/pull/595

PR #4886, opened 2019-07-11
Minor edits to whats-new for v3.6

raoanirudh:


doc/whats-new.md
```diff
@@ -139,42 +145,45 @@ quite significant, for instance the generation of ruptures for a large
...
 
-We fixed a small bug in the rupture XML exporter that was failing
+We fixed a small bug in the rupture XML exporter, which was failing
 if the user did not specify the hazard sites.
 
 We added the ability to generate hazard curves without storing the GMFs,
 simply by setting the flags
 ```
-  hazard_curves_from_gmfs=true
-  ground_motion_fields=true
```
raoanirudh: Is this correct? Shouldn't it be 
`hazard_curves_from_gmfs = true`
`ground_motion_fields = false` ?

doc/whats-new.md
```diff
@@ -293,27 +307,27 @@ the engine that generated it, and some relevant parameters, like the
...
 
 In this release, we restored the ability to perform the association
-directly in the engine.  This is less efficient than using `oq
-prepare_site_model`, since the same associations will be recomputed at
+directly in the engine. This is less efficient than using `oq
+prepare_site_model`, since the same associations will be recomputed during
 each run. It is still useful for people wanting to experiment with the
 grid spacing: they can run several calculations and when they are
 happy with the grid spacing, run `oq prepare_site_model` and fix the
-site model once and for all with the good grid spacing.
```
raoanirudh: The use case that isn't yet covered by `oq prepare_site_model` is the one where, say, the user has a rupture model and would like to compute ground motion fields on a grid around the rupture. The key inputs in such cases would be:
1. A `region` polygon, defined by lon lat co-ordinates of its vertices
2. A `region_grid_spacing`
3. A `site_model_file`

doc/whats-new.md
micheles: Yep, my mistake

doc/whats-new.md
micheles: In that case the user should generate a file with the hazard sites (by gridding the polygon) and pass it to `oq prepare_site_model` together with vs30.csv file:

`$ oq prepate_site_model -s sites.csv vs30.csv`

PR #4893, opened 2019-07-17
Moved `max_sites_disagg` from openquake.cfg into the job.ini

micheles:
It was still not working with `--config openquake.cfg` because the workers were reading the original configuration.


daniviga:
LGTM

PR #4908, opened 2019-07-20
Avoid the error for POE_TOO_BIG

micheles:
The error makes a lot of sense if there is a single site (you cannot run a disaggregation for a PoE larger than the PoEs that can be generated by the model), less so in the multi-site situation. So I changed it to a warning and I am skipping the sites affected by the problem. If no sites are left, a `SystemExit('Cannot do any disaggregation')` is raised. This is necessary, otherwise @julgp calculation for Colombia cannot be run.


julgp:
LGTM

PR #4912, opened 2019-07-22
Huge improvement on multi-sites disaggregation

micheles:
I have measured an improvement of 40x for memory occupation, data transfer and data storage in a calculation for Colombia with ~1200 sites given to me by @julgp . Basically now it can be done even without a cluster.


julgp:
That sound really good! Thanks, Michele! LGTM

Issue #4930, opened 2019-07-23
Incorrect seismicity rates/total moment release for YC1985 MFD from_total_moment_rate


cossatot:
The `YoungsCoppersmith1985MFD` produces incorrect rates if seismicity when it is initialized with the `from_total_moment_rate` class method constructor.  After some investigation, I believe that the error results from the rounding of the characteristic magnitude, and the discretization of the characteristic 'boxcar' of a finite width into several magnitude bins (this is done in the `get_min_mag_and_num_bins` method). The problem is that the characteristic rate is not adjusted to account for the changes in the number and central value of the bins.  Because the seismic energy release is 10^(1.5 * `mag` + 9.05), even very small changes to `mag` values yield very large changes in seismic energy release.

As a result, the calculated total moment release rate (which is the sum of all of the seismic energy released over one year) is not equal to the total moment rate that is used to initialize the class; i.e. the energy budget is not maintained. 

The magnitude and sign of the error is dependent on the `bin_width` and `char_mag` of the MFD (and probably the `min_mag` as well).

Here is a figure illustrating how the calculated total moment release rate (in blue) varies from the input total moment rate (1.0e16 N m /yr, in orange) based on the `char_mag` parameter:

![image](https://user-images.githubusercontent.com/2835718/61719180-48e0f200-ad19-11e9-8bc0-36264c27b622.png)

The rapid increases in calculated moment rate occur when a new bin is added to the characteristic boxcar (e.g. when the characteristic magnitude bins go from [7.1, 7.5, 7.9] to [7.2, 7.4, 7.6, 7.8] or something based on the rounding and discretization).

Similarly by changing the `bin_width`, the calculated moment rate varies as a function of bin width, though it should be constant. The variability increases with `bin_width`, to the point where it can be a factor of 2 higher or lower than it should be for a fault when `bin_width` is close to 0.5 magnitude units which is the maximum allowed for this MFD:
![image](https://user-images.githubusercontent.com/2835718/61719202-539b8700-ad19-11e9-91d7-cb15bcd3bb0f.png)

I have a fix in a branch on my `oq-engine` fork, which I will make into a pull request when I finish testing, hopefully today.

PR #4931, opened 2019-07-23
Youngs and Coppersmith 1985 from_total_moment_rate fix

cossatot:
This is in response to issue #4930 . 

The error has to do with a rounding and discretization of a 'boxcar' section of the MFD.

My fix is simple and either clever or terrible, depending on your perspective.

Because of the complexity of estimating the magnitude of the correction before the rounding etc. has occurred, I simply make a temporary class within the `YoungsCoppersmith1985MFD.from_total_moment_rate()` class method, calculate the total moment rate from the MFD, and then adjust that based on the misfit in between it and the input `total_moment_rate` arugment, and the re-initialize the class (finally) with the adjusted `total_moment_rate` argument.

If there is another, deeper fix then I would be happy to use that instead, but for the time being this solves the problem and produces the correct seismicity rates.


micheles:
LGTM too

PR #4937, opened 2019-07-24
Moved the CompositeRiskModel inside riskmodels

micheles:
And also renamed `riskmodel`->`cr_model` and other changes. This should make life easier for readers of the code like @raoanirudh


openquake/calculators/classical_bcr.py
```diff
@@ -44,7 +44,7 @@ def classical_bcr(riskinputs, riskmodel, param, monitor):
     R = riskinputs[0].hazard_getter.num_rlzs
     result = AccumDict(accum=numpy.zeros((R, 3), F32))
     for ri in riskinputs:
-        for out in riskmodel.gen_outputs(ri, monitor):
+        for out in ri.gen_outputs(crmodel, monitor):
```
raoanirudh: @micheles, can you confirm this change is correct? The variables 'ri' and 'crmodel' seem switched compared to the previous version.

raoanirudh: NM, I see that the method has been moved too.

Issue #4898, opened 2019-07-18
Improve the management of vulnerability functions with beta distribution


micheles:
There is no need to generate the epsilons in that case. Also, we need to set the seed properly and to add tests.

TODO: see if we can unskip `oq-risk-tests/test/event_based_risk/inputs/skip_case_1f`

PR #4938, opened 2019-07-24
Setting the seed for VulnerabilityFunction with Beta Distribution

micheles:
Closes https://github.com/gem/oq-engine/issues/4898. I am also setting the seed to the master_seed for VulnerabilityFunctionWithPMF (before the seed was hard-coded to 42!)


raoanirudh:
The following test cases in oq-risk-tests also exercise the Beta distribution if more tests are needed:
Classical Risk: case_1b, case_1d
Scenario Risk: case_1d, case_1g
Event Based Risk: case_1e

micheles:
The tests in oq-risk-tests are running and I have unskipped the case_1e and case_1f. Can you have a look at the numbers to see if they make sense?

Issue #4541, opened 2019-03-15
Try to unify the losses_by_event exporters


micheles:
It should work the same in scenario_risk, event_based_risk and ebrisk.

raoanirudh:
The `losses_by_event` output for the `event_based_risk` calculator includes the fields `event_id`, `rlz_id`, `rup_id`, and `year`, followed by the respective portfolio losses for each loss category. 

In comparison, the `losses_by_event` output for the `ebrisk` calculator is currently missing the `rlz_id`, `rup_id`, and `year` fields, which would be very useful to include in the output csv, as it is difficult for a user to obtain these fields through alternate means.

PR #4948, opened 2019-07-26
Losses by event

micheles:
Part of https://github.com/gem/oq-engine/issues/4541. Now the losses_by_event exporter is the same for ebrisk, scenario_risk and scenario_damage.


openquake/qa_tests_data/scenario_damage/case_4b/expected/losses_by_event.csv
```diff
@@ -1,201 +1,201 @@
...
-91,1,9.74514E+03,1.82478E+04,3.92422E+03
-92,1,1.00673E+04,1.77096E+04,3.44821E+03
-93,1,8.59188E+03,1.53098E+04,2.38836E+03
-94,1,4.07984E+03,6.92272E+03,6.09233E+02
-95,1,1.20164E+04,2.14793E+04,5.16137E+03
-96,1,1.55286E+04,2.94459E+04,9.23570E+03
-97,1,7.83625E+03,1.38717E+04,1.81385E+03
-98,1,1.25997E+04,2.57561E+04,7.72237E+03
-99,1,1.22807E+04,2.60105E+04,7.98402E+03
+event_id,contents,nonstructural,structural
```
raoanirudh: @micheles, we lost the realization id from the output!

micheles:
I know, it is on purpose. Since now the exporters are unified when it will be added to ebrisk it will be added to scenario_risk and scenario_damage too. It is just temporary. Now I am thinking if we should generate one file per realization or not. What is the most convenient thing for you?

raoanirudh:
I think a single file would work better. While the scenarios would typically have a manageable number of realizations, the event based calculations might have dozens or even hundreds of realizations, which would lead to a proliferation of output files if we split by realization.

PR #4955, opened 2019-07-29
Hide but do not delete removed calcs

daniviga:
Instead of removing calculation from the DB we are marking it as 'deleted' and those are filtered out in outputs.

I also removed the unused `relevant` field. I left it into the DB since `drop column` is not supported by SQLite and the table needs to be recreated, which may be inconvenient for large databases.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,5 @@
   [Daniele Viganò]
+  * Deleted calculations are not removed from the database anymore
   * Removed the 'oq dbserver restart' command since it was broken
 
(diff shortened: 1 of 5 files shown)


micheles:
Please update the changelog

micheles:
We forgot to notify @ptormene about this.

daniviga:
> We forgot to notify @ptormene about this.

He should not care about it: the API isn't changed. Deleted calcs are not exposed via the API. The `deleted` status is only internal to the DB.

PR #4957, opened 2019-07-29
July 2019 Update of Kotha et al. (2019)

g-weatherill:
Updates the Kotha et al. (2019) GMPE with new coefficients using a robust linear mixed regression analysis.

Adds three new subclasses for modelling site amplifications:
1. 2nd order polynomial site amplification model based on Vs30
2. 2nd-order polynomial site amplification model based on slope
3. Two-segment linear model based on slope conditional upon geological era (Addresses last task of https://github.com/gem/oq-engine/issues/4584)

Further updates may still be possible, so the ExperimentalWarning is retained, but will eventually removed in a future PR


openquake/hazardlib/gsim/kotha_2019.py
```diff
@@ -276,60 +279,133 @@ def get_stddevs(self, C, stddev_shape, stddev_types, sites):
...
+    def get_site_amplification(self, C, sites, imt):
+        """
+        Defines a second order polynomial site amplification model
+        """
+        # Render with respect to 0.1 m/m reference slope
+        sref = np.log(sites.slope / 0.1)
+        return C["g0_slope"] + C["g1_slope"] * sref +\
+            C["g2_slope"] * (sref ** 2.)
+
+    def get_stddevs(self, C, stddev_shape, stddev_types, sites):
```
mmpagani: You are not using sites in this method.

openquake/hazardlib/gsim/kotha_2019.py
```diff
@@ -276,60 +279,133 @@ def get_stddevs(self, C, stddev_shape, stddev_types, sites):
...
+    linear function. This form of the GMPE defines the
     site in terms of a measured or inferred Vs30, with the total
     aleatory variability adjusted accordingly.
     """
 
     #: Required site parameter is not set
     REQUIRES_SITES_PARAMETERS = set(("vs30", "vs30measured"))
 
-    def get_site_amplification(self, C, sites):
+    def get_site_amplification(self, C, sites, imt):
```
mmpagani: Here you do not use the `imt` parameter

openquake/hazardlib/gsim/kotha_2019.py
```diff
@@ -350,10 +426,109 @@ def get_stddevs(self, C, stddev_shape, stddev_types, sites):
...
+        geol_units = np.unique(sites.geology)
+        t_slope = np.copy(sites.slope)
+        t_slope[t_slope > 0.1] = 0.1
+        for geol_unit in geol_units:
+            idx = sites.geology == geol_unit
+            if geol_unit in self.GEOLOGICAL_UNITS:
+                # Supported geological unit - use the random effects model
+                v1 = C_AMP_FIXED["V1"] + C_AMP_RAND_INT[geol_unit.decode()]
+                v2 = C_AMP_FIXED["V2"] + C_AMP_RAND_GRAD[geol_unit.decode()]
+            else:
```
mmpagani: Are you sure you want this? If the geology of a site is misspelt you will apply this without knowing it. Did you consider using an `unknown` geological category?

g-weatherill: It's difficult to decide where to put the check. I think it is more likely that the user might map an existing geological database to the input file, which could contain many units besides the ones listed here and not explicitly be classed as "unknown" per se, and in these cases the same fixed effects should be used as is the true "unknown" case. Rather than require the users to create such a mapping in which everything but the supported units is classed as "unknown" I would rather let it run on the broadest set geological units and leave it to the users to ensure that their data are correct.

openquake/hazardlib/gsim/kotha_2019.py
g-weatherill: In other variants of the `.get_site_amplification(...)` method used in the subclasses one or both of sites and/or imt are needed. So I keep the call the the function general for all cases in the main class even if not all the terms are used in the subclasses.

openquake/hazardlib/gsim/kotha_2019.py
g-weatherill: As previous comment

PR #4969, opened 2019-08-01
Cleanly decouple `gsim.get_mean_std` from `get_poes`

micheles:
This breaks the signature of `get_poes` but does not affect the gmpe-smtk. Also, made `_disaggregate_pne` accepts directly a matrix of (M, P) intensities.


openquake/calculators/disaggregation.py
```diff
@@ -123,9 +123,10 @@ def compute_disagg(dstore, slc, cmaker, iml2s, trti, bin_edges, monitor):
         oq.investigation_time)
     pne_mon = monitor('disaggregate_pne', measuremem=False)
     mat_mon = monitor('build_disagg_matrix', measuremem=False)
+    gmf_mon = monitor('computing mean_std', measuremem=False)
```
mmpagani: `gmf` is a bit misleading. What about gms?

openquake/hazardlib/gsim/nshmp_2014.py
```diff
@@ -111,94 +133,38 @@ def get_weighted_poes(gsim, sctx, rctx, dctx, imt, imls, truncation_level,
...
-        for spec of input and result values.
-        """
-        # Get original mean and standard deviations
-        mean, stddevs = super().get_mean_and_stddevs(
-            sctx, rctx, dctx, imt, stddev_types)
-        # Return mean, increased by the adjustment factor,
-        # and standard devation
-        return mean + nga_west2_epistemic_adjustment(rctx.mag, dctx.rrup),\
-            stddevs
+    get_mean_and_stddevs = get_mean_and_stddevs
```
mmpagani: I do not understand this.

micheles: This is used to convert a function external to the class into a method of the class

openquake/calculators/disaggregation.py
micheles: I am not sure what to use here. I will think of something.

PR #4987, opened 2019-08-07
Add a link to our code mirror on Gitea [skip ci]

daniviga:
Maybe we can add something in other project too, like in IRMT


pslh:
LGTM

Issue #4979, opened 2019-08-02
NGA East GMPEs not instantiating correctly


g-weatherill:
It seems as though as change in the instantiation of the `GMPETable` class hasn't made it's way into the NGA East GMPEs that inherit from it. The NGAEast base class runs `__init__` from the superclass (GMPETable), which now calls `init(fle=None)` instead. This means that the required attributes are not read from the hdf5 tables.

The following change to lines 429-467 seems to make this work on my machine:
```python
    def __init__(self, gmpe_table, tau_model="global", phi_model="global",
                 phi_s2ss_model=None, tau_quantile=None,
                 phi_ss_quantile=None, phi_s2ss_quantile=None):
        """
        ...
        """
        self.kwargs = {"gmpe_table": gmpe_table}
        self.tau_model = tau_model
        self.phi_model = phi_model
        self.phi_s2ss_model = phi_s2ss_model
        self.TAU = None
        self.PHI_SS = None
        self.PHI_S2SS = None
        if self.phi_s2ss_model:
            self.ergodic = True
        else:
            self.ergodic = False
        self.tau_quantile = tau_quantile
        self.phi_ss_quantile = phi_ss_quantile
        self.phi_s2ss_quantile = phi_s2ss_quantile
        self._setup_standard_deviations(fle=None)
        super().init(fle=None)
```

If this fix is correct then I can open a PR with it.

micheles:
It it is broken while all the tests in hazardlib are green, it means that they are bad tests not testing the real use case. Please include a good test in your PR, @g-weatherill

PR #4990, opened 2019-08-15
Fix bug in NGA East GMPE Instantiation

g-weatherill:
Fixes a potential issue in the use of the NGA East GMPEs raised here: https://github.com/gem/oq-engine/issues/4979

When instantiating the NGA East GMPEs in an environment like ipython/Jupyter the data normally read from the hdf5 files were not being parsed, as such the GMPE would not be defined for any IMTs, REQUIRES_DISTANCES would be empty etc.). This is because the instantiation was only calling `.__init__()` from the superclass and not the more recently added `.init()`. This was not being found in in the tests because they call `.init()` automatically. A test case PSHA calculation using the NGA East GMPEs was working correctly without the fix, so this would seem to only affect openquake.hazardlib usage w/o the main engine. A test has been added to ensure that all arguments are parsed from the hdf5 at instantiation.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,6 @@
+  [Graeme Weatherill]
+  * Fixed bug in NGA East instantiation and added test
+  
   [Michele Simionato]
   * In the agglosses output of scenario_risk the losses were incorrectly
     multiplied by the realization weight
--- a/openquake/hazardlib/gsim/nga_east.py
+++ b/openquake/hazardlib/gsim/nga_east.py
@@ -465,6 +465,7 @@ class NGAEastBaseGMPE(GMPETable):
         self.phi_s2ss_quantile = phi_s2ss_quantile
         self._setup_standard_deviations(fle=None)
         super().__init__(gmpe_table=gmpe_table)
+        super().init()
 
     def _setup_standard_deviations(self, fle):
         # setup tau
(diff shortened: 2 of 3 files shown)


micheles:
I see. This is not a bug then, it is the intended behavior, the mistake is in not having documented it properly. I am going to add some notes to https://docs.openquake.org/oq-engine/advanced/parametric-gmpes.html
The point is that GMPEs use the double-step initialization pattern in which objects are partially initialized until you call ``.init()`` and ``__init__`` *must not call* ``init`` because the two operations are meant to happen in different contexts. This was introduced when there was no shared directory requirement in the engine, so the workers could not call ``.init`` (having no access to the file system) and only the controller node had to call it. If ``__init__`` could call ``init`` then the workers would have given an error. Right now the workers can access the file system but it is still better to keep the two-step initialization procedure, with ``__init__`` being the fast and easy part of the initialization and ``init`` being the slow and hard part, possibly requiring more permissions.
Users of hazardlib must call the ``.init`` method manually, this is the intention and I am going to document it.

g-weatherill:
So this means that the call to `.init()` in the `.__init__()` as implemented above would be incorrect? If so, then for hazardlib only use-cases (including the gmpe-smtk) is there any downside to always instantiating models via `gsim = MyModel().init()`?

micheles:
> So this means that the call to .init() in the .__init__() as implemented above would be incorrect? 

Yes.

> If so, then for hazardlib only use-cases (including the gmpe-smtk) is there any downside to always instantiating models via gsim = MyModel().init()?

`.init()` returns `None` so it would not work. There are several ways to get the behavior you want (i.e. instantiating a fully initialized GMPE in hazardlib without calling ``.init` manually), I am thinking what it is the best way. Unfortunately the implementations of GMPETable, NGEEast and  NBCC2015_AA13 classes is not particularly clean and it looks like I will have to refactor it first.

PR #5001, opened 2019-08-20
avg_losses-rlzs and avg_losses-stats in ebrisk

micheles:
Removing the special case for avg_losses at the cost of more data transfer.


micheles:
As it is now, this is a no go. I measured a slow down in "computing risk" of a factor 34x in some simple calculation and it can be much worse, so I am putting this on hold.

PR #5007, opened 2019-08-22
Stopping the DbServer must stop the zworkers

micheles:
This could also solve the problem with zmq zombies, since now `popen.terminate()` is called explicitly at the end of each calculation.


micheles:
Apparently, the zombies are still there :-(

PR #5028, opened 2019-08-31
Do not sort site coordinates if site_ids are provided by user

raoanirudh:
Calculations starting from ground motion fields input by the user require at least two input files related to the gmf data:
1. A sites.csv file, listing {site_id, lon, lat} tuples
2. A gmfs.csv file, listing {event_id, site_id, gmv[IMT1], gmv[IMT2], ...} tuples

The site coordinates defined in the sites file do not need to be in sorted order; the only requirement should be to ensure there is no duplication of either the site_ids or the coordinates.

However currently, when creating the site mesh from the site coordinates read from the csv file, the coordinates always get sorted and are assigned new site ids (if the input coordinates were not in sorted order), breaking the link with the user-specified site_ids and thus also with the gmf data.

This patch will avoid sorting the coordinates when the site_ids are explicitly provided in the user input sites file.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,13 @@
-  [Michele Simionato]
+[Anirudh Rao]
+  * Fixed a bug in the gmf CSV importer: the coordinates were being
+  sorted and new site_ids assigned even though the user input sites
+  csv file had site_ids defined
+  * Replaced the nonzero covs in test_case_2 in gmf_ebrisk_test to
+  ensure replicability of numbers
+  * Corrected the expected number for test_case_2 in gmf_ebrisk_test
+  based on hand calculations
+
+[Michele Simionato]
   * Fixed a bug in the rupture CSV exporter: the boundaries of a GriddedRupture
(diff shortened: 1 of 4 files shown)


micheles:
There was a reason why the coordinates were sorted. I do not remember now, but I suspect that the same sites in different orders will give you different hazard and/or risk. If that is the case, we should add a check that the sites provided by the user are ordered by lon, lat and raise an error otherwise.

raoanirudh:
The interesting lines are:

```python
c = (coords[start:stop] if header[0] == 'site_id'
     else sorted(coords[start:stop]))
return geo.Mesh.from_coords(c)
```

The first line does the correct thing and checks if the file has a `site_id` field in the header, and if so, doesn't sort the coordinates. The geo.Mesh.from_coords method, however, sorts the coordinates by default if the `sort` parameter is not passed, negating the check on the first line.

micheles:
Please update the changelog

PR #5036, opened 2019-09-03
Fixed the seed management for the beta distribution again

micheles:
See https://github.com/gem/oq-engine/issues/4898. The seed was lost after unpickling and reset to None. Now the problem cannot happen anymore since I added an `assert self.seed is not None` in `set_distribution`.


raoanirudh:
Could we add a test case for this?

micheles:
Let's add it in a separated PR (or you can it it yourself into oq-risk-tests).

Issue #5045, opened 2019-09-05
Have a way to collect and confront all the CSV outputs


micheles:
In particular the same fields should have the same names in different outputs, and there should be an easy way to compare with previous versions of the engine.

PR #5046, opened 2019-09-05
Changed write_csv and read_csv to operate on real CSV files

micheles:
The trick is to add the right number of commas in the pre-header row and to quote the last field. As a bonus, now github can display properly the outputs of the engine.
This is part of https://github.com/gem/oq-engine/issues/5045


openquake/qa_tests_data/classical/case_1/expected/hazard_curve-PGA.csv
```diff
@@ -1,3 +1,3 @@
-# generated_by='OpenQuake engine 3.5.0-gitd83c42cda6', start_date='2019-05-12T07:12:05', checksum=141718627, kind='rlz-000', investigation_time=1.0, imt='PGA'
+#,,,,,"generated_by='OpenQuake engine 3.7.0-git5e91b20b88', start_date='2019-09-05T06:13:25', checksum=141718627, kind='rlz-000', investigation_time=1.0, imt='PGA'"
```
raoanirudh: Why all the commas at the beginning?

mmpagani:
Michele, all, can we have a meeting (perhaps next week) to consolidate the
output formats?
Grazie
Marco

micheles:
Absolutely @mmpagani , we are going to have a meeting with the risk guys and the hazard team should be involved too, particularly with outputs like the event based ruptures.

openquake/qa_tests_data/classical/case_1/expected/hazard_curve-PGA.csv
micheles: I originally put the commas at the beginning but then it looks bad on GitHub :-(

raoanirudh:
There are some decent recommendations about managing metadata for csv files and groups of related tabular files from the CSV working group: https://github.com/w3c/csvw

Particularly:
[CSV on the Web: A Primer](http://www.w3.org/TR/2016/NOTE-tabular-data-primer-20160225/)
[Model for Tabular Data and Metadata on the Web](https://www.w3.org/TR/2015/REC-tabular-data-model-20151217/)
[Metadata Vocabulary for Tabular Data](http://www.w3.org/TR/2015/REC-tabular-metadata-20151217/)

micheles:
This is very interesting, @raoanirudh, but for me having a separate metadata file is a no-go. We already too many files. The pre-header looks better IMO.

PR #5077, opened 2019-09-13
Now the engine automatically sorts the user-provided intensity_measure_types

micheles:
As requested by @CatalinaYepes.

### DIFF
--- a/openquake/hazardlib/valid.py
+++ b/openquake/hazardlib/valid.py
@@ -651,7 +651,7 @@ def intensity_measure_type(value):
 def intensity_measure_types(value):
     """
     :param value: input string
-    :returns: non-empty list of Intensity Measure Type objects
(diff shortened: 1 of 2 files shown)


openquake/hazardlib/valid.py
```diff
@@ -664,9 +664,7 @@ def intensity_measure_types(value):
       ...
     ValueError: Duplicated IMTs in SA(0.1), SA(0.10)
     >>> intensity_measure_types('SA(1), PGA')
-    Traceback (most recent call last):
-    ...
-    ValueError: The IMTs are not sorted by period: SA(1), PGA
+    ['PGA', 'SA(1.0)']
```
raoanirudh: Could add 'PGV' here to illustrate how it will get sorted wrt the 'PGA' and 'SA' strings.

PR #5082, opened 2019-09-16
Ameri et al. (2017) GMPE

guyomd:
Implementation of Ameri et al. (2017) GMPE, for stress drop-independent (heteroscedastic variability) and stress drop-dependent (homoscedastic variability) versions, and taking into account two distance metrics (Rjb, Repi).


micheles:
@guymod, I see that the directory data/AMERI2017 contains over 8 MB of data and 17,000+ lines of CSV. If it was possible to reduce the size of the test files it would help a lot. We do not what to bloat the size of the engine package.

guyomd:
Hi Michele,

What is a typical directory size for gmpe test files?
I would simply remove a fraction of test files based on the ratio (desired size / actual size), however with a quick check that test still span a sufficient range of parameters values...

G.

De: notifications@github.com
Envoyé: 18 septembre 2019 2:45 PM
À: oq-engine@noreply.github.com
Répondre à: reply@reply.github.com
Cc: guillaume.daniel@outlook.com; author@noreply.github.com
Objet: Re: [gem/oq-engine] Ameri et al. (2017) GMPE (#5082)


@guymod, I see that the directory data/AMERI2017 contains over 8 MB of data and 17,000+ lines of CSV. If it was possible to reduce the size of the test files it would help a lot. We do not what to bloat the size of the engine package.

micheles:
17,000 lines are too much, if you can reduce to say 1,700 lines (by summing the lines all the CSV files) it would be fine.

Issue #5094, opened 2019-09-19
Enable the SWMR mode


micheles:
This is a major project that will allow us to get rid of the cache files. Several intermediate steps
will be needed:

- [x] collapse the task_info table into a single table (1)
- [x] remove the ability to create datasets from the hdf5.extend3 function (2)
- [x] write "~" files in SWMR mode, so that oq show performance does not break the calculation (3)
- [x] remove the obsolete autoflush and pass hdf5path to the Monitor (4)
- [x] make sure libver='latest' and swmr=True as set when instantiating hdf5.File (5) 
- [x] move the task_sent information from attributes to a proper dataset (6)
- [x] pass an open HDF5 file object to the Starmap, not a file path (7)
- [x] pass an open HDF5 file object to the Monitor, not a file path (8)
- [x] remove `set_nbytes` functionality since we cannot set attributes in SWMR mode (9)
- [x] remove the `hdf5.extend3` functionality altogether (10)
- [x] avoid closing the datastore in the .run method; it means that the tests with a parent calculation
       both (engine and oq-risk-tests) must close the parent datastore explicitly (11)
- [x] remove the `poes` cache file in classical calculations (12)
- [x] remove the `ruptures` cache in event based calculations (13)
- [x] remove the cache from ebrisk (14)
- [x] remove the `DataStore.extend` method (15)
- [x] remove the `rupgeoms` cache in event_based calculations (16)
- [x] remove the `~` files and store the performance info directly (17)
- [x] remove the `sitecol` from the cache file (18)
- [x] remove the dataset `rlzs_by_grp` from the cache file (19)
- [x] make sure the SWMR mode is not set too early in classical (20)
- [x] remove the dataset `weights` from the cache file (21)
- [x] remove the epsilons from the cache file

PR #5096, opened 2019-09-19
Removed the extend3 functionality

micheles:
Part of https://github.com/gem/oq-engine/issues/5094. Opening a file, extending a dataset, and closing it gives an error if at same time the file object is being opened in SWMR mode by the same process, so this approach cannot be used. Surprisingly, it was the only thing that worked to store the performance info in regular mode.


micheles:
Superseded by https://github.com/gem/oq-engine/pull/5097/

PR #5129, opened 2019-09-26
Adds configurable nonergodic option to BC Hydro and SERA GMPEs

g-weatherill:
For the Abrahamson et al (2015 - really 2016) BC Hydro GMPEs and the Kotha et al. (2019) GMPEs used for the SERA project this PR makes it possible to run the calculations using the nonergodic standard deviation if configured by the user. The ergodic use case is retained as the default.


micheles:
LGTM on my side

PR #5141, opened 2019-09-30
Fix the WebUI start and improve SIG management

daniviga:
In multi-user mode the DbServer isn't started in background anymore.

Also some more improvements are made:
- make the propagation of signals working
- log output to stdout/err so it can be collected by docker logs (only when in non interactive mode)


daniviga:
A test has been added too: https://github.com/gem/oq-engine-celery/commit/f22a2b0e441c8276118a68bf82c3cf0e982afc79

micheles:
LGTM

micheles:
LGTM

PR #5155, opened 2019-10-02
Add a subprocess.wait() after terminate() in workerpool

daniviga:
This avoids zombies after workers have been stopped in `zmq` mode.


daniviga:
Backported to https://github.com/gem/oq-engine/tree/engine-3.7

PR #5160, opened 2019-10-03
Replaced `pointsource_distance` with `collapse_distance`

micheles:
The problem with the `pointsource_distance` concept is that it does not depend from the magnitude
(while it should) and that it is not clear how to set it. Here I propose to use the rupture radius as a measure of the finite-size effects and to introduce a `collapse_distance` as a pure number in terms of the rupture radius. For sites more distant than the collapse distance the nodal plane distribution and hypocenter distribution are collapsed to the first value. The approximation is only enabled if there are many sites (more than `max_sites_disagg`) but not in the case of few sites (disaggregation) while one wants to do things properly and also the speedup would be minor.
After several experiments it looks like a default of 2 for the collapse_distance is a good number, with a big speedup and a minor loss of precision in the hazard maps. A test case_43 has been added
to play with the collapse_distance (actually
I measured a real time speedup of 3-4 times for a calculation by Julio for CCA without any visible change in the hazard maps.


micheles:
Merging, otherwise the branch will get out of sync with master. We have two months to test this feature and it easy to disable it.

PR #5191, opened 2019-10-14
Added an API /v1/calc/validate_zip

micheles:
For use with IPT, which should upload an archive job.zip file under the name "archive".

### DIFF
--- a/openquake/server/views.py
+++ b/openquake/server/views.py
@@ -282,19 +282,20 @@ def validate_nrml(request):
 @require_http_methods(['POST'])
 def validate_zip(request):
     """
-    Leverage the engine libraries to check if a given job.zip file is valid
+    Leverage the engine libraries to check if a given zip archive is a valid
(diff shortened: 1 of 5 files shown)


openquake/server/views.py
```diff
@@ -276,6 +277,33 @@ def validate_nrml(request):
...
+        a `django.http.HttpRequest` object containing a zip archive
+
+    :returns: a JSON object, containing:
+        * 'valid': a boolean indicating if the provided job.zip is valid
+        * 'error_msg': the error message, if any error was found
+                       (None otherwise)
+    """
+    archive = request.FILES.get('archive')
+    if not archive:
+        return HttpResponseBadRequest('Missing job.zip file')
```
nastasi-oq: ``job.zip`` is too specific without a real reason.

micheles: Changed to archive

PR #5195, opened 2019-10-15
Morikawa 2 the revenge

mmpagani:
This adds new versions based on an updated version of this model as described in a .doc file kindly shared with us. Unfortunately, we do not have full coverage.


rcgee:
LGTM

PR #5206, opened 2019-10-16
Restore a css class in engine/includes/footer.html

### LINKED ISSUE #5205: WebUI footer is broken
Probably a regression in https://github.com/gem/oq-engine/pull/5180/

Screenshot from 2019-10-16 17-54-19

daniviga:
We must retain the class `footer` to keep compatibility with `base.css`

The change also fixes `oq-platform-standalone` which has indeed the same issue.

Fixes #5205


micheles:
LGTM

PR #5209, opened 2019-10-17
Shift hypo

mmpagani:
This PR adds a new (alternative) logic controlling the way the OQ Engine models hypocenters in distributed seismicity. 

For each point source, the OpenQuake Engine places the ruptures in the seismogenic layer using the corresponding hypocentral depth distribution and assuming that the hypocenter is located in the centre of the rupture. For hypocentral depths close to the upper and lower seismogenic depth and relatively large magnitudes the ruptures do not fit anymore the seismogenic layer. In these cases, the OQ engine shifts the ruptures back within the seismogenic layer breaking the initial assumption that the hypocenters are placed at the centre of the rupture. This PR adds to the OQ Engine the option of shifting hypocenters consistently with the ruptures. 

In order to use this option, all the user must do is to set in the `.ini` file the `shift_hypo` parameter to true i.e. `shift_hypo = true`


daniviga:
Please do a refresh with master :)

openquake/hazardlib/calc/disagg.py
```diff
@@ -337,7 +337,7 @@ def disaggregation(
              'filter_distance': filter_distance, 'imtls': {str(imt): [iml]}})
         contexts.RuptureContext.temporal_occurrence_model = (
             srcs[0].temporal_occurrence_model)
-        rdata = contexts.RupData(cmaker).from_srcs(srcs, sitecol)
+        rdata = contexts.RupData(cmaker, **kwargs).from_srcs(srcs, sitecol)
```
micheles: This looks wrong to me, RupData does not accept kwargs. Strange that no test broke.

micheles:
Apart the minor comment, LGTM

Issue #5211, opened 2019-10-18
Revisit the site model associations


micheles:
They currently too complex and actually wrong in the case of an implicit grid, since they are discarding sites. We also need to document in the advanced manual the algorithm used and add missing tests (for instance one with a large grid for the site params and a fine implicit grid).

PR #5214, opened 2019-10-18
Getting the sites from the exposure in precedence over the site model

micheles:
This is how it should be, but it was implemented the other way around. For instance for case_4 this was the situation before (hazard sites from the site model)
from-site-model
and this is the situation now (hazard sites over the assets)
from-exposure

Part of https://github.com/gem/oq-engine/issues/5211.

**CAVEAT**: there is a reason for the current behavior. Originally the exposure had the precedence over the site model, as documented in @raoanirudh original specs, but then users had exposures with thousands and thousands of assets on distinct sites. When performing event based risk calculations with lots of sites it is inevitable to run out of memory and the engine was dying all the time. Therefore years ago the logic was changed to give the precedence to the site model. This essentially was putting the assets on the site model grid, discarding sites with no assets and reducing substantially the number of distinct sites. If we go back to the original behavior, as proposed here, we need to raise an early error for calculations with too many sites, telling people that they should use the implicit grid feature or to aggregate the exposure in some way. Since now we have the implicit grid feature active by default on the engine and since we have the ebrisk calculator which works also with many sites it is now probably possible to go back to the original specs.


micheles:
Superseded by https://github.com/gem/oq-engine/pull/5217

PR #5226, opened 2019-10-21
Fixed GriddedSurface.get_surface_boundaries

micheles:
@julgp discovered that the geometries exported by the engine for nonparametric surfaces were broken. The problem was that I was exporting the bounding box, not the associated polygon.
The good news is that there is no need to repeat any calculation, the bug was purely in the exporter.


micheles:
Daniele discovered at least another bug with simple fault ruptures, but it will be fixed in a different PR.

PR #5231, opened 2019-10-22
Solved the problem of the NaNs in scenario_damage

micheles:
The solution is to treat a missing `noDamageLimit` attribute of the fragility file as a zero.
Closes https://github.com/gem/oq-engine/issues/3583.


openquake/risklib/read_nrml.py
```diff
@@ -162,7 +162,7 @@ def ffconvert(fname, limit_states, ff, min_iml=1E-10):
     with context(fname, ff):
         ffs = ff[1:]
         imls = ff.imls
-    nodamage = imls.attrib.get('noDamageLimit')
+    nodamage = imls.attrib.get('noDamageLimit', 0)
     if nodamage == 0:
         # use a cutoff to avoid log(0) in GMPE.to_distribution_values
         logging.warning('Found a noDamageLimit=0 in %s, line %s, '
```
raoanirudh: This warning may not be correct if the nodamage == 0 is not coming from the fragility file, but is being set to zero in the previous line.

micheles: But if the `noDamageLimit` is missing in the fragility file the user should still know that the cutoff is used, don't you think? It is just the warning that it is misleading saying `Found a noDamageLimit=0`` when it is not there. I could just keep the note about using a cutoff.

micheles: On the other hand the user can get hundreds of warnings that she cannot do anything about because the fragility functions come from a third party. So I am reducing the warning to a debug log.

PR #5235, opened 2019-10-23
Fixed duplication in the np/hd distributions

micheles:
It was there in the Canada model once, IIRC.


micheles:
NB: the change affects the rupture seeds in event based, so the test for British Columbia in oq-risk-tests broke.

PR #5259, opened 2019-10-31
Implemented IMT-dependent weights with sampling

micheles:
As requested by @rcgee


rcgee:
LGTM

micheles:
But it is giving the wrong numbers for Arabia, so this is blocked until I fix it or find a better solution, see https://github.com/gem/oq-engine/issues/5261.

micheles:
After much thought, it is impossible to do that. If the weights for different IMTs are different, a calculation for PGA and a calculations for SA(1.0) (for instance) will produce different realizations, so how can they be combined?

PR #5263, opened 2019-11-02
Added `require_file` in the job.ini

micheles:
Once upon a time, all files required by a calculation were listed in the job.ini file. However, after the introduction of parametric GMPEs, it is possible to list files in the gsim_logic_tree file. This is bad for two reasons:

1. it breaks `oq zip` and the datastore will miss files, thus making the calculation not transferable to another machine
2. if there is a missing file the error will be discovered when reading the logic tree, while it should be
discovered earlier, when reading the job.ini

To solve these issues it is now required to specify the required files in the job.ini with a dictionary-like
syntax, like the following:
```
require_file = {
  'b14': 'kappa_factors_b14.csv',
  'cy14': 'kappa_factors_cy14.csv',
  're19-5': 'kappa_factors_re19-5.csv',
  'c14': 'kappa_factors_c14.csv',
  're19-10': 'kappa_factors_re19-10.csv',
  'ya15': 'kappa_factors_ya15.csv'}
```

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,5 @@
   [Michele Simionato]
+  * Added `require_file` in the job.ini to check the existence of required files
   * Made it possible to disaggregate on multiple realizations
     with the parameters `rlz_index` or `num_rlzs_disagg`
   * Fixed downloading the ShakeMaps (again)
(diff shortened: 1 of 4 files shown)


micheles:
Closing since there is better way to to this by extending `readinput.get_input_files`

micheles:
First attempt towards https://github.com/gem/oq-engine/issues/5269.

PR #5294, opened 2019-11-13
Revises SERA BCHydro Epistemic + fix to Kotha 2019

g-weatherill:
Adjustments to SERA ground motion models:
1. Revises the stress parameter epistemic uncertainty scaling to replace magnitude-dependent factor with imt-dependent factors of Abrahamson et al. (2019)
2. Minor fix to Kotha et al. (2019) SERA model to cap the vs30 at 1100 m/s


micheles:
LGTM for me too

PR #5300, opened 2019-11-19
Made it possible to import a GMF.csv file with more IMTs than needed

micheles:
As requested by @CatalinaYepes who also provided the test case.


openquake/calculators/base.py
```diff
@@ -983,13 +983,20 @@ def import_gmfs(dstore, fname, sids):
     if names[0] == 'rlzi':  # backward compatbility
         names = names[1:]  # discard the field rlzi
     imts = [name[4:] for name in names[2:]]
-    gmf_data_dt = dstore['oqparam'].gmf_data_dt()
-    arr = numpy.zeros(len(array), gmf_data_dt)
-    col = 0
+    oq = dstore['oqparam']
+    if set(oq.imtls) > set(imts):
+        raise ValueError('%s contains %s but the risk model needs %s' %
```
raoanirudh: Might be more informative to list the missing IMTs in this case.

micheles: Done

raoanirudh: Would `set(oq.imtls) - set(imts)` give the missing imts?

PR #5301, opened 2019-11-19
Report (and use?) the number of visible cores, not total system cores

bebosudo:
In a cgroups limited environment, like a SLURM cluster, the number of processors available is different from the number of procs available in the whole system.

Example of a "limited" env:
```bash
$ srun -N 1 --ntasks-per-core=1 -c 10 --pty bash
$ grep 'model name' /proc/cpuinfo |wc -l
36    # Total number of cores in the whole system
$ nproc
10    # Cores usable
$ python3 -c 'import multiprocessing as mp; import os; print(mp.cpu_count(), len(os.sched_getaffinity(0)))'
36 10
```

This is only a "graphical" fix in the logs; right now oq spawns a number of processes matching the whole system cores available, while it should use only the ones available, as reported in the multiprocessing doc:
https://docs.python.org/3.8/library/multiprocessing.html#multiprocessing.cpu_count

This causes a tremendous overhead because processes are fighting for resources.
I can't find the place where the number of processes to use for multiprocessing is defined, any help would be appreciated.
Let me know if you think this may be useful for oq.

### DIFF
--- a/openquake/engine/engine.py
+++ b/openquake/engine/engine.py
@@ -335,8 +335,9 @@ def run_calc(job_id, oqparam, exports, hazard_calculation_id=None, **kw):
         return
     try:
         if OQ_DISTRIBUTE.endswith('pool'):
+            # Report the number of "visible" cores, not the total system cores
             logs.LOG.warning('Using %d cores on %s',
-                             parallel.Starmap.num_cores, platform.node())
+                             len(os.sched_getaffinity(0)), platform.node())
         if OQ_DISTRIBUTE == 'zmq' and config.zworkers['host_cores']:
             logs.dbcmd('zmq_start')  # start the zworkers
             logs.dbcmd('zmq_wait')  # wait for them to go up
--- a/openquake/engine/engine.py
+++ b/openquake/engine/engine.py
@@ -335,7 +335,6 @@ def run_calc(job_id, oqparam, exports, hazard_calculation_id=None, **kw):
         return
     try:
         if OQ_DISTRIBUTE.endswith('pool'):
-            # Report the number of "visible" cores, not the total system cores
             logs.LOG.warning('Using %d cores on %s',
                              parallel.Starmap.num_cores, platform.node())
         if OQ_DISTRIBUTE == 'zmq' and config.zworkers['host_cores']:
--- a/openquake/engine/engine.py
+++ b/openquake/engine/engine.py
@@ -337,7 +337,7 @@ def run_calc(job_id, oqparam, exports, hazard_calculation_id=None, **kw):
         if OQ_DISTRIBUTE.endswith('pool'):
             # Report the number of "visible" cores, not the total system cores
             logs.LOG.warning('Using %d cores on %s',
-                             len(os.sched_getaffinity(0)), platform.node())
+                             parallel.Starmap.num_cores, platform.node())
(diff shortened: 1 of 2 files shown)


micheles:
A pull request involving functionalities not available on windows/mac is a no go. Since engine 3.7 you can set the parameter `num_cores` in the job.ini and the engine will use at maximum that number of cores. This is a workaround, but it should solve the contention issue for the existing version of the engine. For the future, we will have to think of something. If the number of cores is fixed in could go as a configuration parameter in the openquake.cfg file. This is easier than trying to figure it out dynamically which requires a proper integration with SLURM.

bebosudo:
For windows and mac we could use psutil (I didn't want to add a new dependency initially):
https://psutil.readthedocs.io/en/latest/#psutil.Process.cpu_affinity

This feature could be relevant not only to SLURM or other job scheduler users, but also to users of containers.
In my case this is not an issue, since I'll run SLURM requesting whole nodes each time.

daniviga:
> For windows and mac we could use psutil (I didn't want to add a new dependency initially):
> https://psutil.readthedocs.io/en/latest/#psutil.Process.cpu_affinity

`psutil` is already a dependency: see https://github.com/gem/oq-engine/blob/master/setup.py#L63
We are providing binary wheels for all the supported OS/arch combinations (upstream provides binary wheels only for Windows).

I confirm that it works just fine on Windows and macOS too and for sure on Linux (we use it in a different project that heavily relies on Docker and cgroups).

```bash
docker run --rm -it --cpuset-cpus="1,3" python:3.6 /bin/bash
root@6270beced1b5:/# pip install http://cdn.ftp.openquake.org/wheelhouse/linux/py36/psutil-5.4.3-cp36-cp36m-manylinux1_x86_64.whl
Collecting psutil==5.4.3
  Downloading http://cdn.ftp.openquake.org/wheelhouse/linux/py36/psutil-5.4.3-cp36-cp36m-manylinux1_x86_64.whl (250kB)
     |████████████████████████████████| 256kB 995kB/s 
Installing collected packages: psutil
Successfully installed psutil-5.4.3
root@6270beced1b5:/# python3 -c 'import psutil; p = psutil.Process(); print(p.cpu_affinity())'
[1, 3]
```

openquake/engine/engine.py
```diff
@@ -335,8 +335,9 @@ def run_calc(job_id, oqparam, exports, hazard_calculation_id=None, **kw):
         return
     try:
         if OQ_DISTRIBUTE.endswith('pool'):
+            # Report the number of "visible" cores, not the total system cores
             logs.LOG.warning('Using %d cores on %s',
-                             parallel.Starmap.num_cores, platform.node())
+                             len(os.sched_getaffinity(0)), platform.node())
```
daniviga: It would be better to fix the cpu count in https://github.com/gem/oq-engine/blob/master/openquake/baselib/parallel.py#L595 directly (maybe using `psutil` as discussed) and not here, but @micheles has the final word on this

bebosudo: Oh great, there it is!
I can fix it properly using psutil, if you find this useful. In case: should I simply revert the commit, rebase my branch or create a new PR?

micheles: Just add a new commit that changes the right line and restore the currently changed line as it was before.

bebosudo: Done!

micheles:
LGTM

Issue #4093, opened 2018-10-18
Allow export-dir to be specified on the command line


daniviga:
It would be great to be able to specify the output_dir via command line when '--run' is used. This should override the output folder set in the job.ini. This was suggested by @vpoggi.

@micheles proposed `--export-dir=EXPORTDIR`

micheles:
Let's close this, I do not want to change the manual and for the moment we have https://github.com/gem/oq-engine/pull/4092

PR #5305, opened 2019-11-20
Change export dir default behavior to current working dir

klunk386:
This pull request modified the default path used to export the calculation results, so that files are saved in the current directory.

### DIFF
--- a/openquake/commonlib/oqvalidation.py
+++ b/openquake/commonlib/oqvalidation.py
@@ -789,8 +789,7 @@ class OqParam(valid.ParamSet):
         and the user must have the permission to write on it.
         """
         if self.export_dir and not os.path.isabs(self.export_dir):
-            self.export_dir = os.path.normpath(
-                os.path.join(self.input_dir, self.export_dir))
+            self.export_dir = os.getcwd()
         if not self.export_dir:
             self.export_dir = os.path.expanduser('~')  # home directory
             logging.warning('export_dir not specified. Using export_dir=%s'


openquake/commonlib/oqvalidation.py
```diff
@@ -789,8 +789,7 @@ def is_valid_export_dir(self):
         and the user must have the permission to write on it.
         """
         if self.export_dir and not os.path.isabs(self.export_dir):
-            self.export_dir = os.path.normpath(
-                os.path.join(self.input_dir, self.export_dir))
+            self.export_dir = os.getcwd()
```
daniviga: I would add a warning is the `export_dir` is not set (like it was in the past):


```suggestion
            self.export_dir = os.getcwd()
            logging.warn('export_dir not specified. Using export_dir=%s'
                         % self.export_dir)
```

micheles:
This could break some of our scripts, so we need to check.

daniviga:
We'll check with @rcgee and @raoanirudh if any internal workflow is relying on the current behavior, if none, I guess we can merge this.

daniviga:
It's funny, because this change was actually requested by you, @vpoggi :D

daniviga:
Travis (with forced demos execution) is happy: https://travis-ci.org/gem/oq-engine/jobs/614549344

raoanirudh:
> We'll check with @rcgee and @raoanirudh if any internal workflow is relying on the current behavior, if none, I guess we can merge this.

None I can think of

bebosudo:
We wanted to propose this PR because we didn't find a way to set the export dir on the command line along with `--run`: we'd like to completely _control_ oq from the CLI so we don't need to remember each time to add `export_dir` inside the job.ini file.

Using this patch we don't need to "grep" on the stderr of oq to find the calculation id to export later, since we can set the working dir to the output dir and let oq save data in the desired location, but it's still a bit of a hack.
We'd like to propose maybe a new CLI flag to use in combination with `--run` to override the export directory; this way we could let the default behavior as it is, and simply use that flag in our wrapper around oq that we use to run simulations in a HPC cluster.

Where should I go in the source code to define a new flag for the export directory?

cc @klunk386

daniviga:
> We wanted to propose this PR because we didn't find a way to set the export dir on the command line along with `--run`: we'd like to completely _control_ oq from the CLI so we don't need to remember each time to add `export_dir` inside the job.ini file.

We were already thinking about it, but at that time we were in hurry: https://github.com/gem/oq-engine/issues/4093

klunk386:
It is quite funny it was me proposing this last year. I could not even 
remember it... :D

Il 21/11/19 12:15, Daniele Viganò ha scritto:

micheles:
Can't you use the combination
```bash
$ oq engine --run job.ini && oq engine --eos -1 /my/export/dir
```
`-1` means the last computation.

bebosudo:
> Can't you use the combination
> 
> ```shell
> $ oq engine --run job.ini && oq engine --export-outputs -1 /my/export/dir
> ```

No unfortunately using a 2-steps export is prone to race conditions in a distributed environment where multiple simulations are fired at the same time, like how we are doing with our wrapper.
Let me illustrate it with an example:
```
--run simulation1  # id=1
--run simulation2  # id=2
.. time passing by
end of simulation 2 & export -1 => export last simulation in database => export id=2
end of simulation 1 & export -1 => export last simulation in database => export id=2
```

This is why we need to have a way to either directly export, or at least be sure to extract the correct id for each simulation.

micheles:
Let's approve your PR, since it is going back to the traditional behavior of the engine that worker for 7 years until Valerio asked to change it. If we still have some script relying on the export_dir to be relative to the input_dir and it breaks we will fix it.

openquake/commonlib/oqvalidation.py
micheles: I am adding it

micheles:
Unfortunately some important scripts by @mmpagani are breaking, plus we may be breaking things for other users without us knowing, so I am reverting this change. You have better ways of doing what you want to do, like writing your own wrapper.

PR #5309, opened 2019-11-20
cpu_affinity is not available on macOS

daniviga:
I was too optimistic: `cpu_affinity` is not available on macOS (it is on FreeBSD, _but_ not on mac, sigh!)
Jenkins spotted it: https://ci.openquake.org/job/macos/job/master_macos_engine/label=mojave,python=python3.7/5220/console
Now it runs: https://ci.openquake.org/job/macos/job/zdevel_macos_engine/label=mojave,python=python3.7/2/console

I also updated the changelog and added @bebosudo to contributors.

Regression introduced by #5301

### DIFF
--- a/openquake/baselib/parallel.py
+++ b/openquake/baselib/parallel.py
@@ -594,9 +594,9 @@ class Starmap(object):
     running_tasks = []  # currently running tasks
     # use only the "visible" cores, not the total system cores
     # if the underlying OS supports it (macOS does not)
-    if hasattr(psutil.Process(), 'cpu_affinity'):
+    try:
         num_cores = len(psutil.Process().cpu_affinity())
-    else:
+    except AttributeError:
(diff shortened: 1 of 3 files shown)


openquake/baselib/parallel.py
```diff
@@ -592,8 +592,12 @@ def getargnames(task_func):
 class Starmap(object):
     pids = ()
     running_tasks = []  # currently running tasks
-    # Use only the "visible" cores, not the total system cores.
-    num_cores = len(psutil.Process().cpu_affinity())
+    # use only the "visible" cores, not the total system cores
+    # if the underlying OS supports it (macOS does not)
+    if hasattr(psutil.Process(), 'cpu_affinity'):
+        num_cores = len(psutil.Process().cpu_affinity())
```
micheles: A `try .. except AttributeError` would be more pythonic

PR #5333, opened 2019-11-28
Make WebUI ACL more flexible

daniviga:
As requested by @micheles :

- `ACL_ON` is `False` by default: all users can have access to all outputs/exports
- List of calculations is always filtered (on users or groups) unless a request comes from a `superuser`
- Abort can be fired only when request comes from the owner or `superuser`: this prevents other users to kill jobs even when `ACL_ON = False`
- There was a bug in the failed abort error message (the wrong user was reported)


micheles:
LGTM!

PR #5347, opened 2019-12-03
Implemented IMT-dependent weights with sampling by cheating

micheles:
The cheat is to ignore the IMT-dependent weights: an error is logged but the computation continues.
The test we have in case_30 uses the following GMPE logic tree:
```xml
                    <logicTreeBranch branchID="YoungsEtAl1997SInter">
                        <uncertaintyModel>YoungsEtAl1997SInter</uncertaintyModel>
                        <uncertaintyWeight>0.33</uncertaintyWeight>
                        <uncertaintyWeight imt="SA(1.0)">0.3</uncertaintyWeight>
                    </logicTreeBranch>
                    <logicTreeBranch branchID="AtkinsonBoore2003SInter">
                        <uncertaintyModel>AtkinsonBoore2003SInter</uncertaintyModel>
                        <uncertaintyWeight>0.33</uncertaintyWeight>
                        <uncertaintyWeight imt="SA(1.0)">0.4</uncertaintyWeight>
                    </logicTreeBranch>
                    <logicTreeBranch branchID="ZhaoEtAl2006SInter">
                        <uncertaintyModel>ZhaoEtAl2006SInter</uncertaintyModel>
                        <uncertaintyWeight>0.34</uncertaintyWeight>
                        <uncertaintyWeight imt="SA(1.0)">0.3</uncertaintyWeight>
                    </logicTreeBranch>
```
Here is the effect of the change of weight in the test; calculation 33700 uses weights .33, .33, .34 for all IMTs and calculation 33701 uses weights .3, .4, .3 for all IMTs.
PGA
SA(1 0)


micheles:
An alternative would be to raise an error and force the user to change the gsim_logic_tree file or to force him to set a parameter like `ignore_imt_dependent_weights=true`. If desired I can do that, but for the moment the easiest thing for the final user is to have the engine perform the calculation anyway (especially for risk users who have no idea about how to change the logic tree).

Issue #5343, opened 2019-12-02
Can we move the calculation of aggregated event loss tables in postprocessing?


micheles:
Probably not for event based, but certainly yes for scenario_risk, and it is worth testing the feasibility of the approach.

PS: after 11 days of work, it turns out that the approach is viable even for ebrisk, if we discard the smallest losses, which affect only the low period portion of the loss curves.

micheles:
The trick to make this work is to store only the highest losses per asset, i.e. the tail of the distribution instead of the full distribution.

raoanirudh:
Can we consider adding pandas as a dependancy? Some of the functionality pandas provides that could be very useful for dealing with large damage tables and loss tables, from the [package overview](https://pandas.pydata.org/pandas-docs/stable/getting_started/overview.html):

- Powerful, flexible group by functionality to perform split-apply-combine operations on data sets, for both aggregating and transforming data
- Robust IO tools for loading data from flat files (CSV and delimited), Excel files, databases, and saving / loading data from the ultrafast HDF5 format
- Intelligent label-based slicing, fancy indexing, and subsetting of large data sets
- Intuitive merging and joining data sets
- Flexible reshaping and pivoting of data sets

micheles:
I actually support this, independently from the event_based_damage project, it makes sense for us to have a "batteries included" openquake distribution.

g-weatherill:
+1 to this. Several additional toolkits building on OQ use Pandas, so there are few situations where one may use OQ in a distribution without Pandas. Any possibility there could be speed-ups in, for example, the site collection object if it worked as a Pandas dataframe?

micheles:
I sincerely doubt that using Pandas will speedup the site collection. Providing Pandas will make it easier to explore the datastore (see https://github.com/gem/oq-engine/pull/5357) and to perform post-processing analysis, but for the time being I would refrain from using it in the core engine. I do not trust it as much as I trust numpy, especially performance-wise.

PR #5349, opened 2019-12-04
Storing only the highest losses in the asset loss table

micheles:
First step towards https://github.com/gem/oq-engine/issues/5343. The `highest_losses` parameter may disappear, in the sense that in the future it may be decided by the engine and not by the user.
 Also, I introduced a `minimum_asset_loss` dictionary in the job.ini.

NB: in this first attempt the asset loss table is kept fully in memory. This will likely change later on,
to avoid things like `storing asset loss table 2,177s`.


micheles:
Agreed. Actually the actually highest_losses has been removed a few weeks ago, this PR is nearly two months old ;-)

PR #5350, opened 2019-12-04
Small refactor of AvgSA GMPE

g-weatherill:
Refactors the GenericGMPEAvgSa to fix some usage bugs and cleanup:

1. Builds the correlation matrix upon instantiation of the GMPE rather than on each call

2. Allows for interpolation of the Akkar coefficient matrix - so can now specify any period within the range supported by the GMPE

3. Can be run with GMPEs that require additional arguments

4. Ensures that the GenericGmpeAvgSA class is always recognised as returning total standard deviation regardless of the GMPEs it calls (was causing an error in the event-based calculation)


micheles:
LGTM on my side

PR #5358, opened 2019-12-08
Changed logic tree sampling by ordering the branches by TRT

micheles:
This is yet another ordering bug. `gsim_lt.branches` and `gsim_lt.values` are dictionaries that should be ordered the same way but the first was ordered by TRT and the other by the order of the branches in the XML file. The consequence was that in the case of sampling the logic tree read from the XML file was different from the logic tree read from the datastore, even when they should have been the same. Also, in the test case_master with sampling ebrisk was seeing a different logic tree than event_based.


micheles:
This is also affecting the British Columbia calculation in oq-risk-tests, https://gitlab.openquake.org/openquake/oq-risk-tests/-/jobs/12734

PR #5362, opened 2019-12-10
Change how wheels are cached by Travis

daniviga:
Now pip is capable of caching also wheels downloaded from our requirements, so the hack we were using to cache `whl` can be removed. It was very fragile when a wheel was updated/removed.


daniviga:
I also cherry-picked the commit  to 3.7 and 3.6 (I did not went beyond)

Issue #5354, opened 2019-12-06
Add the ability to collapse the GSIM logic tree


micheles:
This was requested by Crescenzo and it is actually trivial to implement on our side. The way I envision the feature, the user will be able to write in the job.ini something like

```
collapse_gsim_logic_tree = bs1 bs2
```
and then the branchset bs1 and bs2 of th full GSIM logic tree will be collapsed, i.e. the multiple gsims for the corresponding tectonic region type will be replaced by a single average gsim.
The use case is event based: a model like the one for Pakistan would change from 800 realizations for source_model_1 + 800 realizations for source_model_2 (or something like that) to 1 + 1 realizations, thus saving 800 times more disk space in the GMFs and making it possible the ingestion into a financial software.

mmpagani:
I think it's key to specify what you mean by 'single average gsim'

micheles:
It will be implemented as a geometric mean, similarly to what we did for the AvgSA GMPE.

PR #5367, opened 2019-12-11
AvgGMPE

micheles:
The AvgGMPE computes the geometric mean of the underlying GMPEs, similarly to AvgSA. It is here for experimental purposes. When I will have time, I plan to see if it is usable for risk purposes (it it were, the problem of IMT-dependent weights would be solved, since we could use full enumeration after collapsing the gsim logic tree). This would solve https://github.com/gem/oq-engine/issues/5354.


openquake/qa_tests_data/classical/case_19/gmpe_logic_tree.xml
```diff
@@ -27,25 +27,17 @@
...
-                </logicTreeBranch>
-                
-                <logicTreeBranch branchID="b24"> <uncertaintyModel>ToroEtAl2002SHARE</uncertaintyModel>
-                    <uncertaintyWeight>0.20</uncertaintyWeight>
-                </logicTreeBranch>
-                
-                <logicTreeBranch branchID="b25"> <uncertaintyModel>Campbell2003SHARE</uncertaintyModel>
-                    <uncertaintyWeight>0.20</uncertaintyWeight>
-                </logicTreeBranch>
+              <logicTreeBranch branchID="b2">
```
mmpagani: Why are you changing an existing test? Wouldn't it be better to add a new one?

micheles: The interesting thing is that in this test using the AvgGMPE produces the *same mean curves* as before, up to the numeric precision, even if in general the AvgGMPE does NOT produce the same mean curves. I am reusing an existing tests because I wanted to check the difference between the hazard curves obtained from the original logic tree and from the collapsed one.

PR #5398, opened 2020-01-08
adding pointsource_distance to view_params 

kejohnso:
This PR adds the pointsource_distance specified in the job.ini (or otherwise "None") to the view_params in openquake/calculators/views.py so that it appears in report_<id>.rst.


micheles:
LGTM

PR #4852, opened 2019-06-26
Tabular GMPEs of MPS04

giovannilanzanoINGV:

PR #5407, opened 2020-01-10
Removed ita04 classes

micheles:
They were merged by mistake, coming from https://github.com/gem/oq-engine/pull/4852.


mmpagani:
LGTM

PR #5414, opened 2020-01-15
Added a mechanism to overridde the job.ini parameters

micheles:
And reverted the export_dir to be relative to the input_dir, as before https://github.com/gem/oq-engine/pull/5305. Here is an example of usage:

 $ oq engine --run demos/hazard/AreaSourceClassicalPSHA/job.ini --exports csv --param export_dir=/tmp/pippo


klunk386:
ok, LGTM

PR #5427, opened 2020-01-20
Update fault_models.py

AvinashSingh786:
When a non-default bin width is given the collapsed version does not use the specified bin width instead it uses the default. This commit uses the bin width specified from the recurrence model.

### DIFF
--- a/openquake/hmtk/faults/fault_models.py
+++ b/openquake/hmtk/faults/fault_models.py
@@ -461,7 +461,7 @@ class mtkActiveFault(object):
             if np.max(model.magnitudes) > mmax:
                 mmax = np.max(model.magnitudes)
         if collapse:
-            self.mfd = ([self.collapse_branches(mmin, bin_width, mmax)],
+            self.mfd = ([self.collapse_branches(mmin, model.recurrence.bin_width, mmax)],
                         [1.0],
                         [rendered_msr])
         else:


AvinashSingh786:
At the moment I can't provide any automated tests, but a manual check can be done, when building a model from a YML file when a non-default bin_width is given it does not use the provided bin_width from the YML instead it uses the default of 0.1. A test would be running a FaultYmltoSource on a YML file and then build_fault_model and the serialise_to_nrml using collapsed branching. When doing this the bin_width doesn't change from the default 0.1 to the value specified in the YML.

Code:
reader = FaultYmltoSource(yml_file)
fault_model, tectonic_region = reader.read_file(mesh_spacing)
fault_model.build_fault_model(collapse=False, rendered_msr=WC1994())
fault_model.source_model.serialise_to_nrml(output_file, use_defaults=False)

Hope this helps.
Thanks.

mmpagani:
Okay. Let's merge it as is for the time being. I would be grateful if you could provide a simple .yml file exercising this. Thank you again.

AvinashSingh786:
Hi, okay so I am attaching a ZIP, in the YML under MDF_Spacing for the recurrence model it is set to 0.05. I wrote a quick python script and generated the xml files before and after the patch.

Hope this helps!
[Recurrence_test.zip](https://github.com/gem/oq-engine/files/4139985/Recurrence_test.zip)

g-weatherill:
I think I understand the bug that @AvinashSingh786  has raised, but I don't think the fix proposed is the correct one. From the example shown the fix should be made here: https://github.com/gem/oq-engine/blob/master/openquake/hmtk/faults/active_fault_model.py#L114

It seems as though the bin_width argument doesn't get passed to the generate_recurrence_models function, so the default is never overwritten

AvinashSingh786:
Thanks @g-weatherill, I think this solution is cleaner, I didn't take the time to trace exactly how everything was flowing. However, with your proposed solution that would require the bin_width to be passed in with the function but instead, the bin_width is specified within the YML file, so should there not be a helper function to read it out of the file or to extract if out of the MFD_config when a fault is built from a YML file?  I see in the YML Parser https://github.com/gem/oq-engine/blob/master/openquake/hmtk/parsers/faults/fault_yaml_parser.py#L187 that reads the MFD config and appends it to the fault. Is there a way to extract it to pass this bin_width to build_fault_model because the bin_width for each recurrence_model can be different?

g-weatherill:
The suggested fix has been implemented in a new PR (#5567). Thank you @AvinashSingh786 for bringing this to our attention and for providing the test files. From my side, this PR can now be closed.

AvinashSingh786:
Thanks @g-weatherill for finding the elegant way to fix this issue.

Issue #5339, opened 2019-11-29
Event based damage calculator


raoanirudh:
## Introduction
Heretofore the engine has provided five primary risk calculators: `classical_damage`, `classical_risk`, `scenario_damage`, `scenario_risk`, and `event_based_risk`. A stochastic `event_based_damage` calculator has been the critical missing piece. The rationale and specifications for building such an event based damage calculator are proposed in this issue.

## Rationale
The classical PSHA based damage/risk approach is useful if the user is mainly interested in estimating damage-state probabilities or loss exceedance probabilities within a time-window for individual assets. This approach is also conducive for the computation of time-averaged metrics such as the _average annual collapses_ (or 50-year collapse probabilities) and the _average annual loss_.

While the average annual loss values for different assets that lie within the same target accumulation zones (or any other arbitrary groups of assets that share a set of attributes) can be simply added to yield the aggregate average annual loss for the target accumulation zones (or for the groups of assets), the same doesn't hold for the loss exceedance curves or probabilistic loss maps outputs. For instance, simply adding up the 2,500-year return period losses for all of the assets within an accumulation zone (or any other arbitrary aggregation group) will not necessarily yield the 2,500-year return period loss for the zone (or group of assets). For computing almost all kinds of aggregated loss statistics for a distributed portfolio of assets, the `event_based_risk` calculator provides a much more viable solution. The `event_based_damage` calculator would serve a similar function, allowing users to compute **aggregated damage statistics** for a distributed portfolio of assets

In addition, the `event_based_damage` calculator would also enable the calculation of probabilistic damage metrics such as **damage-state exceedance curves** and **damage-state maps** for different return periods, which the `classical_damage` calculator does not support. Similar to the event based risk calculator, the `event_based_damage` calculator would be able to compute these metrics at both the asset level and also at the aggregated level for groups of assets sharing a set of attributes.

Finally, the central result computed by the `event_based_damage` calculator, viz the _asset-level event damage table_ – would serve as the starting input for estimating a wide set of **consequences** including but not limited to direct economic costs of repairing the damaged buildings, estimates of casualties, displaced households, shelter requirements, loss of use of essential facilities, amount of debris generated, costs associated with loss of function (e.g., business interruption losses), and repair and recovery times. While the computation of consequences would be coupled with the `event_based_damage` calculator, specifications for a general consequences calculator will be proposed in a separate issue and this issue focuses solely on the `event_based_damage` calculator.

## Specifications

[...]

PR #5428, opened 2020-01-21
Moved from floating point damage distributions to integer damage distributions

micheles:
And forced the "number" attribute in the exposure to be an integer in the range 1..65535 (extrema included). Part of https://github.com/gem/oq-engine/issues/5339.


micheles:
The ` numpy.bincount(numpy.random.choice(D, n, p=frac/frac.sum()), minlength=D)` function is terribly slow and will have to go for large event based damage calculations. We already checked that is making the calculation 12x slower in the Messina example. But this is left to a future PR.

PR #5430, opened 2020-01-21
Supported floating point asset numbers in scenario_damage/event_based_damage

micheles:
~For when Murray will not be able to run his calculation, it will be enough to set `ddd_fast=true` and we will gain an order of magnitude of speedup (or  more) by getting similar numbers anyway.~
In that case we will use the old algorithm `fractions * n` (rounded to uint32 integers) instead of `numpy.bincount(numpy.random.choice(D, n, p=frac/frac.sum()), minlength=D)`, and a warning will be printed.


raoanirudh:
Supporting two options will require documentation in the manual to explain both options and may cause confusion down the line. Instead, if a vectorized implementation of the sampling step can be found, it would most likely solve the speed issue. 

A promising solution is listed in this answer: https://stackoverflow.com/questions/34187130/fast-random-weighted-selection-across-all-rows-of-a-stochastic-matrix

micheles:
After the recent discussion, I think that we should change this as follows:
if there is at least one floating point number attribute in the exposure (or at least one number > 65536) use the old algorithm and compute floating point damage distributions which are then not stored and print a warning; otherwise, use the new algorithm.

PR #5433, opened 2020-01-22
Compute the max_intensity amongst all IMTs

micheles:
Instead of using the intensity with the highest period. Notice that all tests pass, so the changes in the numbers are negligible. This is also fast (70 seconds for the GMPEs of the Canada model on my workstation).


kejohnso:
LGTM

Issue #5434, opened 2020-01-23
Try again the idea magnitude-dependent maximum_distance


micheles:
We abandoned it because it was changing the numbers too much, but there is still hope, if we are smarter. It still has the potential to improved the speed of our calculations by 10x, but a lot of work and analysis will be required.

PR #5435, opened 2020-01-23
Store effect_by_mag_dst_trt always if there are enough sites

micheles:
For transparency (before it was stored only if the pointsource_distance was set). It will be even more useful in the future, in particular for https://github.com/gem/oq-engine/issues/5434.


kejohnso:
LGTM

PR #5436, opened 2020-01-24
Lowered the limit on the maximum disagg matrix size to 1E6

micheles:
To stop @rcgee ;-) In her case the problem was a coordinate_bin_width = 0.2, which was too small for the problem at hand, generating a disaggregation matrix too large with over 5M elements.


rcgee:
LGTM

PR #5437, opened 2020-01-24
Fixed bug with MMI

micheles:
As reported here: https://groups.google.com/d/msg/openquake-users/eLxSDJCRzw4/8N4FH0QYEQAJ


openquake/qa_tests_data/classical/case_45/expected/hazard_map-mean-PGA.csv
```diff
@@ -0,0 +1,3 @@
+#,,,"generated_by='OpenQuake engine 3.9.0-git4635b29844', start_date='2020-01-24T09:26:25', checksum=3719331686, kind='mean', investigation_time=50.0"
```
mmpagani: The name of the file is a bit misleading

micheles: Good catch, fixed!

PR #5442, opened 2020-01-25
Extended `minimum_intensity` to classical/disaggregation calculators

micheles:
This actually closes https://github.com/gem/oq-engine/issues/5434.

If the user sets in the job.ini something like `minimum_intensity=0.01` a magnitude-dependent maximum distance is automatically computed for the given `minimum_intensity` and used instead of the normal `maximum_distance`, which is used only if  `minimum_intensity` is missing, which means `minimum_intensity=0`.

In short: by reducing the precision requirement on the low intensity part of the hazard curves one can get an order of magnitude speedup or so. It should be noticed that the low intensity part of the hazard curves is imprecise anyway, since it is sensitive to very far away ruptures which contribution is subject to significant errors (for instance at a distance of 1000 km the tectonic region type has probably changed so you cannot trust the GMPE anyway).

Here are the figures for some calculations on cluster2 for 1 site in Canada:
```
calc_id min_int max_dist time   eff_ruptures
======= ======= ======== ====== ============
34437   0.00    600-600  728_272 40_110_297
34443   0.02    134-600  545_567 30_027_995
34444   0.05     56-600  399_214 21_862_258
34445   0.10     35-516  278_938 15_071_724
```
There is nearly a 3x speedup for minimum_intensity=0.1g, but then of course you canno trust the values of the curve below 0.1g.
canada1


micheles:
I am merging this since the change is safe i.e. by default minimum_intensity is not set and is ignored, nothing changes.

PR #5470, opened 2020-02-03
Improved the error message in case of CSV exposures with wrong headers

micheles:
This happened to Tiegan.


openquake/risklib/asset.py
```diff
@@ -860,14 +860,16 @@ def _read_csv(self):
             with open(fname, encoding='utf-8') as f:
                 fields = next(csv.reader(f))
                 header = set(fields)
+                missing = expected_header - header - {'exposure', 'country'}
```
raoanirudh: Is 'country' a reserved name? Then we should document it.

openquake/risklib/asset.py
```diff
@@ -860,14 +860,16 @@ def _read_csv(self):
...
                 if len(header) < len(fields):
                     raise InvalidFile(
                         '%s: The header %s contains a duplicated field' %
                         (fname, header))
-                elif expected_header - header - {'exposure', 'country'}:
-                    raise InvalidFile(
-                        'Unexpected header in %s\nExpected: %s\nGot: %s' %
-                        (fname, sorted(expected_header), sorted(header)))
+                elif missing:
+                    msg = ('Unexpected header in %s\nExpected: %s\nGot: %s\n'
```
raoanirudh: Here, it would be helpful to list _only_ the field names that do not match, perhaps using set subtract.

openquake/risklib/asset.py
micheles: Yes, you should document it in the manual, Anirudh! ;-)

openquake/risklib/asset.py
micheles: `missing = expected_header - header - {'exposure', 'country'}` is using set subtract and it is displaying such names

Issue #5460, opened 2020-01-31
See if it is possible to use `numpy.load(allow_pickle=False)` in the QGIS plugin


micheles:
At the moment it is not possible. The investigation is to assess how much work it is to change the engine to make it possible. The benefit would be security and most importantly *independence from the Python version*. Right now an engine with Python 3.8 would not work with a QGIS with Python < 3.8 because they changed the pickle protocol.

NB: this is hard, but worth the effort.

PR #5477, opened 2020-02-05
Allow to use `numpy.load(allow_pickle=False)` in the QGIS plugin

ptormene:
Fixes https://github.com/gem/oq-engine/issues/5460


openquake/server/utils.py
```diff
@@ -130,5 +131,23 @@ def check_webserver_running(url="http://localhost:8800", max_retries=30):
...
+    bytes (and therefore modifying its dtype). This is useful in order to
+    prevent the array from being pickled.
+    It returns the unmodified object if it is not a numpy array.
+    """
+    if not isinstance(arr, numpy.ndarray):
+        return arr
+    if arr.dtype == numpy.dtype('O'):
+        try:
+            return numpy.array([s.encode('utf-8') for s in arr])
+        except AttributeError:
```
micheles: You do not need the try.. except. Just use the second formula always.

openquake/server/utils.py
```diff
@@ -130,5 +131,23 @@ def check_webserver_running(url="http://localhost:8800", max_retries=30):
 
     if not success:
         logging.warning('Unable to connect to %s within %s retries'
-                     % (url, max_retries))
+                        % (url, max_retries))
     return success
+
+
```
micheles: I would change the name of the function to `array_of_strings_to_bytes`

PR #5479, opened 2020-02-05
Implements Forearc/Backarc Taper in SERA BC Hydro Model

g-weatherill:
The use of the forearc/backarc switch in the Abrahamson et al. (2016) BC Hydro model can be problematic in national/regional hazard mapping, as the backarc term is a boolean site property. In practice this means that for a hazard map, sites must be assigned as either forearc or backarc, usually in proximity to the volcanic front. The problem is that as the attenuation is faster in the case of backarc sites you get a sudden "cliff-edge" drop in seismic hazard between sites on either side of the front. This behaviour is obviously unphysical! However, whilst the original forearc/backarc scaling term (F_FABA) in the model is a binary term that takes 0 (for forearc) and 1 for backarc and multiplies by the backarc attenuation factor, we can turn this into a continuous term in the range 0 <= F_ABA <= 1, dependent on the proximity to the volcanic front. This, in effect, gives a smooth transition from forearc attenuation to backarc attenuation.

This PR implements several tapering models (`Step`, `Linear`, `Sigmoid`, `SFunc` and `Gaussian`) to be applied to the SERA adaptation of the BC Hydro model (not the original Abrahamson et al., 2016 BC Hydro model itself). These models define F_FABA as a continuous function depending on a new site-specific property called `backarc_distance`, which is a measure of the proximity of the site to the volcanic front and takes negative values in the forearc side, positive values on the backarc side and 0 on the front itself. The selection and shape of the taper can be controlled by the user as arguments passed to the GMPE.

Includes:
1. Implementation of a set of tapers
2. Adjustment to the SERA BC Hydro models to allow the tapering model and controlling parameters to be set by the user. Default is a `Step` model, which requires no additional parameters and keeps the behaviour similar to the binary state in which `backarc_distance <= 0` -> `F_FABA = 0`, and `backarc_distance > 0` -> `F_FABA = 1`.
3. Updates test tables
4. Adds backarc_distance to the site properties
5. Updates classical test case 38 for compatibility with this model


openquake/hazardlib/gsim/bchydro_2016_epistemic.py
```diff
@@ -67,12 +68,174 @@ def get_stress_factor(imt, slab=False):
...
+class FABATaperStep(object):
+    """
+    General class for a tapering function, in this case
+    a step function such that the backarc scaling term takes 0 for
+    forearc sites (negative backarc distance), and 1 for backarc sites
+    (positive backarc distance)
+    """
+    def __init__(self, **kwargs):
+        pass
+
```
micheles: @g-weatherill I guess you are using the `**kwargs` signature here because I said that it was okay to use it in the GMPETable case. Actually it was a lesser of two evils situation: in a situation when you do not know the parameters in advance and you cannot anticipated what in the future will go in the subclasses (as it happens for subclasses of GMPETable) using `**kwargs` is acceptable. If this is the situation here I will not protest. But if you know that the parameters can be at most a, b,c, sigma, etc, it is a  lot better to be explicit and to list them in the signature. The important thing is that the signature must be the same among all subclasses, i.e. it is okay to include parameters that are not used. Unless the number of parameters is huge or it is not known.

g-weatherill: I'm not sure which is the better option here then. As you can see, each subclass has different inputs and even though I use the same parameter names (a, b etc.) some of them have different meanings and therefore different default values. Of course I could re-name the parameters so that each is more specific to the subclass but I'm not sure that is more elegant than using `**kwargs`. I don't anticipate the addition of more subclasses than those present here.

micheles: Then it is easy, Graeme: leave things as they are, but add a comment explaining the choice. A good point where to add the comment is near the lines
```python
    def __init__(self, **kwargs):
        pass
```
which are a bit surprising for somebody looking at the code.

PR #5512, opened 2020-02-12
Export

rcgee:
Looking at the sigma_epsilon_XX.csv file computed from a scenario or evt-based calculation, it was not clear what values were being exported. This PR changed the headers to indicate that the values correspond to inter event sigma. I also add a sentence to implementing-new-gsim.md to clarify the units.


micheles:
Please update the changelog, Robin. LGTM

PR #5519, opened 2020-02-14
Hmtk plotting error

kejohnso:
the catalogue density plotters were missing the last bin on both the x- and y-axes


kejohnso:
Here is an example for an interface segment, for which the original plot does not include the most recent or highest magnitude occurrences. 

**Catalogue:**
[PI_cat_dec_int_kerton_uh.csv.zip](https://github.com/gem/oq-engine/files/4204695/PI_cat_dec_int_kerton_uh.csv.zip)

**Original:**
original

**Corrected:**
corrected

micheles:
Please add a line in the changelog, Kendra.

kejohnso:
Okay, added!

Issue #5513, opened 2020-02-13
Add a demo containing nonparametric sources


micheles:
Part of the CGS project. We need it to test the extract/visualization features.

PR #5520, opened 2020-02-14
Classical PSHA Non-parametric sources Demo

ptormene:
Fixes https://github.com/gem/oq-engine/issues/5513


demos/hazard/NonparametricSourcesClassicalPSHA/README.txt
```diff
@@ -0,0 +1,17 @@
...
+This calculation demo exercises the Classical hazard caclculator with a reduced
+source model logic tree, containing a single model with a non-parametric sources
+with magnitude larger Mw7.0.
+
+In this example, we first compute hazard curves for all IMTs and sites of
+interest (see below). From each curve set we interpolate hazard maps at 10% and
+2% probabilities of exceedance. Finally, we also extract Uniform Hazard Spectra
+(UHS) for the same probabilities from the PGA and SA hazard maps.
+
+Expected runtime:
```
ptormene: We should add the expected runtime (what's the reference kind of machine for this kind of estimate?)

demos/hazard/NonparametricSourcesClassicalPSHA/job.ini
```diff
@@ -0,0 +1,57 @@
...
+truncation_level = 3
+# integration distance in km (J&B distance)
+maximum_distance = {'Subduction IntraSlab': 300.,'default': 300.}
+
+
+[output]
+
+export_dir = /tmp
+mean_hazard_curves = True
+quantile_hazard_curves =
```
ptormene: Is this an expected format?

demos/hazard/NonparametricSourcesClassicalPSHA/README.txt
micheles: Your laptop is a good reference machine. Approximate per excess, I would say in this case the runtime is < 20 s

demos/hazard/NonparametricSourcesClassicalPSHA/job.ini
micheles: You can remove `mean_hazard_curves = True quantile_hazard_curves =` since it is the default.

Issue #5517, opened 2020-02-14
Unify source model logic tree and gsim logic tree


micheles:
Currently - after a lot of work in the last 6 months - they are nearly the same thing; however, for historical reasons they are implemented in two completely different ways causing a big maintenance burden. We have to solve this before doing any serious new work on the logic trees. The unification is also relevant for the risk logic tree project, since it would share the same implementation.

micheles:
Here are the reasons why the unification program cannot succeed:

1. IMT-dependent weights are a feature peculiar to GSIM logic trees
2. logic tree reduction is peculiar to GSIM logic trees
3. GMPETables are treated specially in GSIM logic trees (in __toh5__/__fromh5__)
4. the collapse method used in AvgGMPE is special

The best we can do is to clean up the current implementation of SourceModelLogicTree, but GsimLogicTree has to stay separate.

PR #5546, opened 2020-02-21
Some cleanup of the imports

micheles:
And removed some lines of code. Part of https://github.com/gem/oq-engine/issues/5517.


micheles:
This is no good.

PR #5554, opened 2020-02-24
Fixed the case of duplicated nodal planes

micheles:
This was affecting the Italy model with engine 3.8.


micheles:
But that is tested in the other tests we have, specific for the nodal plane/hypocenter distribution; here I am just adding another case.

PR #5567, opened 2020-02-27
Fixes bin_width bug in hmtk fault model collapse

g-weatherill:
Fixes the bug described here: https://github.com/gem/oq-engine/pull/5427

1. Adds the `bin_width` keyword input to `openquake.hmtk.faults.mtkActiveFaultModel.build_fault_model` 

2. Incorporates a test for collapsed branches with a bin width different from 0.1, using the test case files supplied by @AvinashSingh786


micheles:
LGTM!

PR #5562, opened 2020-02-26
[Doc] Minor updates to implementing-new-gsim.md [skip CI]

raoanirudh:
- Add a note that tuples containing a single element **must** use a trailing comma during assignment
- Link to current PEP 8 guide instead of legacy
- Replace 'GMPE' with 'GSIM' where appropriate

PR #5586, opened 2020-03-02
Define several sets as {'smth'} instead of set(('smth',))

ptormene:
Referring to https://github.com/gem/oq-engine/pull/5562


micheles:
@ptormene please also change the docs in https://github.com/gem/oq-engine/pull/5562.

PR #5587, opened 2020-03-02
Added a sanity check on the total number of assets for each event

micheles:
@CatalinaYepes discovered a bug there, introduced in master.


openquake/calculators/scenario_damage.py
```diff
@@ -213,6 +212,14 @@ def post_execute(self, result):
         # damage by event
         eid_dmg_dt = self.crmodel.eid_dmg_dt()
         d_event = numpy.array(sorted(result['d_event'].items()), eid_dmg_dt)
+        # sanity check on the total number of assets, same for each event
+        num_sums = len(set(d['dmg'].sum() for d in d_event))
+        if self.param['approx_ddd'] and num_sums > 1:
+            logging.warning('Due to numeric errors the sum of assets in each '
+                            'damage state is not the same for each event')
+        elif num_sums > 1:
+            raise ValueError('The sum of assets in each damage state is not '
```
raoanirudh: It won't be clear to the user why they are seeing this error as the only way this can happen is if there's a bug in the damage state simulation.

openquake/calculators/scenario_damage.py
```diff
@@ -213,6 +212,14 @@ def post_execute(self, result):
         # damage by event
         eid_dmg_dt = self.crmodel.eid_dmg_dt()
         d_event = numpy.array(sorted(result['d_event'].items()), eid_dmg_dt)
+        # sanity check on the total number of assets, same for each event
+        num_sums = len(set(d['dmg'].sum() for d in d_event))
+        if self.param['approx_ddd'] and num_sums > 1:
+            logging.warning('Due to numeric errors the sum of assets in each '
```
raoanirudh: "Due to rounding errors, the total number of structural units may not exactly identical for each event in the dmg_by_event table."

openquake/calculators/scenario_damage.py
micheles: This was meant as a sanity check, an error that the final user will never see. However, since yesterday I have changed approach (see https://github.com/gem/oq-engine/pull/5588) and this check is not needed anymore.

PR #5592, opened 2020-03-04
Added an utility to plot logic trees

micheles:
Plotting logic trees in general is a large project that we do not want to start; however plotting small logic trees for teaching purposes is simple and I did it in 30 minutes by using the ete3 library. It is added here in the utils directory, where non-documented, experimental, subject to disappear utilities external to the engine go. @ptormene may want to have a look at it.
Here is an example of usage:
```
# pip install ete3
$ ~/oq-engine/utils/plot_lt classical/case_20/source_model_logic_tree.xml 
```

### DIFF
--- a/utils/plot_lt
+++ b/utils/plot_lt
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# vim: tabstop=4 shiftwidth=4 softtabstop=4
+#
+# Copyright (C) 2020 GEM Foundation
+#
+# OpenQuake is free software: you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


ptormene:
It looks like the option "Show branch info" might be enabled by default

kejohnso:
Ciao, when I first tried this new functionality, I got the following error:     
from ete3 import Tree, TreeStyle, TextFace, add_face_to_node
ImportError: cannot import name 'TreeStyle

I was able to fix this by[ installing PyQt5](https://github.com/etetoolkit/ete/issues/379), which seems to be a dependency of ete3.

PR #5594, opened 2020-03-04
Add a warning to reduce_sm command if duplicate source ids are found

ptormene:
...and filter by `info['eff_ruptures'] > 0` instead of `info['calc_time'] > 0`
Fixes https://github.com/gem/oq-engine/issues/5593


openquake/commands/reduce_sm.py
```diff
@@ -29,13 +31,22 @@ def reduce_sm(calc_id):
...
         oqparam = dstore['oqparam']
         info = dstore['source_info'][()]
-        ok = info['calc_time'] > 0
-        source_ids = set(info[ok]['source_id'])
+        ok = info['eff_ruptures'] > 0
+        source_ids_list = info[ok]['source_id']
+        duplicate_source_ids = [
+            source_id
+            for source_id, count in Counter(source_ids_list).items()
+            if count > 1]
```
micheles: No, use numpy or pandas.

openquake/commands/reduce_sm.py
```diff
@@ -29,7 +31,15 @@ def reduce_sm(calc_id):
     with datastore.read(calc_id) as dstore:
         oqparam = dstore['oqparam']
         info = dstore['source_info'][()]
-        ok = info['calc_time'] > 0
+        source_ids, source_counts = np.unique(
+            info['source_id'], return_counts=True)
```
micheles: This is no good. The check should be the following:
```python
to_remove= info[info['eff_ruptures'] == 0]
source_ids, source_counts = np.unique(to_remove, return_counts=True)
```

micheles:
I have fixed the check and I am merging. The difficult part (i.e. adding a test) is still left to @ptormene

PR #5595, opened 2020-03-04
Improved oq reduce_sm to distinguish sources with same ID but different codes

micheles:
Part of https://github.com/gem/oq-engine/issues/5593. For instance, if a model contains the source "100" with code b"A" within the maximum distance. and also the source "100" with code b"S" outside the maximum distance, before neither of them could be discarded by `oq reduce_sm`, while now b"S" can be discarded correctly.There is still an issue with duplicated source IDs with the same code, one within the maximum distance and one without, but I will do nothing for that case, since it would be too difficult to address.


openquake/commonlib/readinput.py
```diff
@@ -833,11 +834,33 @@ def get_pmap_from_csv(oqparam, fnames):
...
+
+
+# used in oq reduce_sm and utils/extract_source
 def reduce_source_model(smlt_file, source_ids, remove=True):
     """
-    Extract sources from the composite source model
+    Extract sources from the composite source model.
+
+    :param smlt_file: path to a source model logic tree file
+    :param source_ids: dictionary source_id -> source_code
```
ptormene: `source_ids` can also be a list of source IDs as in extract_source (see line 860)

PR #5617, opened 2020-03-09
Small cleanup of the code in sample_cluster

micheles:
A block `if sources.cluster:` can be removed since it is alway True.


micheles:
Closing because it would conflict with https://github.com/gem/oq-engine/pull/5637

PR #5627, opened 2020-03-11
create scaled bindi11 and tests

giovannilanzanoINGV:


micheles:
The tests do not run https://ci.openquake.org/job/zdevel_oq-engine/2867/console, probably because your fork is ancient. Please refresh it from a recent version of the engine (i.e. current master).

PR #5637, opened 2020-03-12
20200310 cluster

mmpagani:
This PR adds the ability to generate stochastic events sets using a cluster with mutually exclusive sources. In a following PR we will add a working example.


openquake/hazardlib/nrml.py
```diff
@@ -204,6 +203,18 @@ def get_source_model_04(node, fname, converter=default):
...
+        itime = valid.positivefloat(itime)
+    stime = node.get('start_time')
+    if stime is not None:
+        stime = valid.positivefloat(stime)
+    # adding this to make sure the converter has the proper investigation
+    # time even when we read just the model .xml file
+    if itime is not None:
+        converter.investigation_time = itime
+    # reading groupss
     groups = []  # expect a sequence of sourceGroup nodes
```
micheles: Mispelling in groupss

openquake/hazardlib/tests/calc/stochastic_test.py
```diff
@@ -55,3 +59,34 @@ def test_nankai(self):
...
+            nr = src.num_ruptures
+            src.serial = start + seed
+            start += nr
+        # Counting the number of occuurences of the three ruptures admitted
+        # by this test model
+        rups, tme = sample_cluster(group, sf, num_ses=1000, param=param)
+        nocc = sum([r.n_occ for r in rups])
+        # The number of occurrences must be close to 10 since the occurrence
+        # rate of the cluster is 0.01 events/year
+        assert nocc == 9
```
micheles: Just use `self.assertEqual(nocc, 9)`

openquake/hazardlib/nrml.py
mmpagani: done

openquake/hazardlib/tests/calc/stochastic_test.py
mmpagani: done

micheles:
There is some amount of duplication among the two branches of the `if`: an helper function would be recommended, but it can go later.

mmpagani:
> There is some amount of duplication among the two branches of the `if`: an helper function would be recommended, but it can go later.

Yes, I noticed that as well. The is also a comprehensive test missing at the moment.

micheles:
After 14 months this PR got spoiled. We well reopen it again when needed, with the new ContextMaker.

PR #5638, opened 2020-03-12
[Doc] Edits to the seeds section in faq-risk [skip CI]

raoanirudh:


micheles:
Why don't you copy the text of the FAQ in the manual too?

PR #5644, opened 2020-03-13
Fixed mosaic/EUR

micheles:
Fixes https://gitlab.openquake.org/hazard/mosaic/EUR/-/jobs/16347. Then issue is for source model files which are duplicated inside the logic tree file.


rcgee:
Thanks Michele. This shouldn't impact the results, right?

micheles:
Yes, no impact on the results. The fix was for a minor regression introduced yesterday while refactoring the logic trees.

PR #5661, opened 2020-03-18
Computed the mean disaggregation if multiple realizations are requested

micheles:
This was requested by @kejohnso .


kejohnso:
Grazie Michele, LGTM

PR #5630, opened 2020-03-11
create bindi et al 2011 scaled

giovannilanzanoINGV:
Now, it is in sync with the new repository. I did the test and should be ok.

PR #5682, opened 2020-03-26
Added cauzzi_2014scaled, bindi_2014scaled, bindi_2011scaled

micheles:
Contributed by the INGV. Supersedes https://github.com/gem/oq-engine/pull/5630.

~@mmpagani let me know if this purely cut and paste from the base classes and could be solved with inheritance (PS: 90% of it is duplication, indeed).~

I have removed all the duplication.


openquake/hazardlib/gsim/bindi_2014scaled.py
```diff
@@ -0,0 +1,82 @@
...
+    D.Bindi, M. Massa, L.Luzi, G. Ameri, F. Pacor, R.Puglia and P. Augliera
+    (2014), "Erratum to Pan-European ground motion prediction equations for the
+    average horizontal component of PGA, PGV and 5 %-damped PSA at spectral
+    periods of up to 3.0 s using the RESORCE dataset", Bulletin of
+    Earthquake Engineering, 12(1), 431 - 448. The erratum notes that the
+    printed coefficients tables were in error. In this implementation
+    coefficients tables were taken from the Electronic Supplementary
+    material of the original paper, which are indicated as being unaffected.
+    """
+    def _get_site_amplification_term(self, C, vs30):
```
mmpagani: This method is identical to https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/gsim/bindi_2014.py#L234 so I would remove it

openquake/hazardlib/gsim/cauzzi_2014scaled.py
```diff
@@ -0,0 +1,74 @@
...
+    spectra and peak ground motions, based on a global dataset of digital
+    acceleration records", Bulletin of Earthquake Engineering, In Press
+
+    Spectral acceleration (SA) values are obtained from displacement response
+    spectrum  (DSR) values (as provided by the original equations) using the
+    following formula ::
+
+        SA = DSR * (2 * π / T) ** 2
+    """
+    def _get_site_amplification_term(self, C, vs30):
```
mmpagani: This method seems to me identical to https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/gsim/cauzzi_2014.py#L471

micheles:
The scaled GMPEs cause a higher hazard. Here is what happens on site (13.3, 42.6) for PGA
PGA
and for SA(4.0)
SA4

openquake/hazardlib/gsim/bindi_2014scaled.py
micheles: Removed

openquake/hazardlib/gsim/cauzzi_2014scaled.py
micheles: Removed

PR #5688, opened 2020-03-26
Implements possibility to use GMPEs with Gaussian Mixture Models in the Standard Deviation

g-weatherill:
Gaussian Mixture Models are a tool for combining two distributions (in this case log-normal) in order to modify the tails and change the probability of rare motions. This has been adopted in some site-specific PSHAs for nuclear facilities in order to model the heavy tailed distribution of very rare ground motions. A more comprehensive description is found in the PEER Probabilistic Seismic Hazard Code Validation Project (Hale et al., 2018 - https://peer.berkeley.edu/sites/default/files/2018_03_hale_final_8.13.18.pdf)

The conditional probability of exceeding a level of ground motion can be determined by summing the probability of exceedence for a set of distributions each with its own standard deviation:

mixture_model_formula

The mixture model is defined in this case by the user from a set of scaling `factors` to be applied to the standard deviation, and their corresponding `weights`. For example, a mixture model of two equally weighted sigmas, one of 0.8 times the original GMPE sigma, the other of 1.2 times the original GMPE sigma, this would be set in the logic tree via:

```
[NameOfGMPE2020]
[NameOfGMPE2020.mixture_model]
factors = [0.8, 1.2]
weights = [0.5, 0.5]
```
There is no limit to the number of distributions in the model, but the computational cost is greater with more.

A complete test of the mixture model was included in the PEER Tests (Test Case 2.5), and this example is included here both as a direct test and as a QA test. Agreement of the logarithm of the hazard curves is to within 0.04 % between the PEER benchmark and the OQ implementation.


micheles:
Notice that due to the way TOML works you can also write the GSIM as
```
[NameOfGMPE2020.mixture_model]
factors = [0.8, 1.2]
weights = [0.5, 0.5]
```
which is one line shorter. LGTM.

PR #5689, opened 2020-03-27
Substantial improvement of the pointsource_distance approximation

micheles:
We now avoid calling  `src._get_rupture_surface` and we returns directly `PointRupture` objects that have no surface. For the moment, the change affects only the case of many sites.

Here are the figures for a reduced SHARE calculation on cluster1:
```
calc_37860                   time_sec memory_mb counts   # master
============================ ======== ========= =========
get_poes                     91_006   0.0       4_124_359
computing mean_std           65_644   0.0       4_124_359
composing pnes               37_990   0.0       4_124_359
iter_ruptures                15_165   0.0       373_516  
make_contexts                14_790   0.0       3_815_554


calc_37862                   time_sec memory_mb counts   # new
============================ ======== ========= =========
get_poes                     87_354   0.0       4_130_971
computing mean_std           63_751   0.0       4_130_971
composing pnes               36_174   0.0       4_130_971
make_contexts                9_010    0.0       3_835_614
iter_ruptures                667      0.0       91_063   
```
The spectacular improvement is in iter_ruptures and make_contexts.


openquake/hazardlib/source/point.py
```diff
@@ -171,6 +175,18 @@ def iter_ruptures(self, **kwargs):
                         surface, occurrence_rate,
                         self.temporal_occurrence_model)
 
+    def point_ruptures(self):
+        """
+        Generate one point rupture for each magnitude
+        """
+        _, np = self.nodal_plane_distribution.data[0]
```
mmpagani: This is the only thing I do not like. I think we should take the hypocenter and nodal plane with the highest weight.

micheles: At long distances it should not matter much, but it is easy to change.

PR #5694, opened 2020-03-28
Enabled the pointsource_distance approximation even in the case of few sites

micheles:
This gives huge speedups as seen in the MultiPointSource demo (from  9,106 to 690 effective ruptures, a 13x reduction. I have also introduce a `collapse_ctxs` method that gives even better collapsing, but it is disables by default, until I understand why it is not working in the Italy model (and working perfectly in all other cases, including the Italy model with 1 realization).


micheles:
PS: the issue with the Italy model was a too low `pointsource_distance`: we need to raise it. The safest value is pointsource_distance=maximum_distance so that finite-size effects for the highest magnitudes are fully considered. I must ask @rcgee to raise `pointsource_distance` in the mosaic, now the engine should be fast enough to make that possible.

PR #5696, opened 2020-03-29
[Doc] Updates to faq-risk [skip CI]

raoanirudh:
Answered the very faq:

What values should I use for `investigation_time`, `ses_per_logic_tree_path`, and `number_of_logic_tree_samples` in my calculation? And what does the `risk_investigation_time` parameter for risk calculations do?


micheles:
LGTM and you could copy the explanation in the official manual too.

PR #5706, opened 2020-04-01
[Doc] Fix some typos [skip CI]

raoanirudh:
pytqt5 → PyQt5
nighlty → nightly

Windows has a built in uncompressor, no need to recommend installing 7-zip

Add silx version to unix and mac doc

### DIFF
--- a/doc/installing/development-windows.md
+++ b/doc/installing/development-windows.md
@@ -84,7 +84,7 @@ To uninstall the OpenQuake development environment remove the folder where it ha
 It is possible to install, as an example, the [Silx HDF5 viewer](http://www.silx.org/) in the same environment as the OpenQuake Engine. To make that happen run the following commands via the `oq-console.bat` prompt:
 
 ```cmd
-pip install PyQt5==5.14.1 silx==0.12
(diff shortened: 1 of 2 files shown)


doc/installing/development-windows.md
```diff
@@ -86,7 +84,7 @@ To uninstall the OpenQuake development environment remove the folder where it ha
 It is possible to install, as an example, the [Silx HDF5 viewer](http://www.silx.org/) in the same environment as the OpenQuake Engine. To make that happen run the following commands via the `oq-console.bat` prompt:
 
 ```cmd
-pip install pytqt5==5.7.1 silx==0.10
+pip install PyQt5==5.14.1 silx==0.12
 ```
```
micheles, line 88: Pay attention @raoanirudh: silx is at version 0.10 because the version 0.12 does not work with the numpy we have, 1.16. But we are going to upgrade to numpy 1.18. Getting the versions right is always tricky.

raoanirudh: Switched back to 0.10 for now

PR #5725, opened 2020-04-08
Fixed sorting bug in /extract/rupture_info

micheles:
In /extract/rupture_info we were sorting the rupture parameters by rup_id but not the geometries!


ptormene:
Before giving my LGTM I forgot to suggest to add a test for this.

PR #5729, opened 2020-04-09
Sanity check on agg_losses/tot_losses

micheles:
The sum of the agg_losses over all the tags must be equal to the tot_losses. It was not so in engine 3.8 as discovered by an user, see https://groups.google.com/d/msg/openquake-users/voHp5eH6Rqg/MWvVMQNnBAAJ

### DIFF
--- a/openquake/calculators/post_risk.py
+++ b/openquake/calculators/post_risk.py
@@ -113,7 +113,7 @@ class PostRiskCalculator(base.RiskCalculator):
         """
         oq = self.oqparam
         stats = oq.hazard_stats().items()
-        S = len(stats)
+        S = self.S = len(stats)
         P = len(builder.return_periods)
         loss_types = oq.loss_names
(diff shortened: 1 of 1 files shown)


openquake/calculators/post_risk.py
```diff
@@ -198,9 +198,31 @@ def execute(self):
...
+        Sanity check on tot_losses
+        """
+        logging.info('Sanity check on agg_losses/tot_losses')
+        for kind in 'rlzs', 'stats':
+            agg = 'agg_losses-' + kind
+            tot = 'tot_losses-' + kind
+            if agg not in self.datastore:
+                return
+            if kind == 'rlzs':
+                kinds = ['rlz-%d' % rlz for rlz in range(self.R)]
```
raoanirudh, line 214: Just a note on the variable names: kind and kinds are referring to different entities, may possibly lead to confusion down the line.

PR #5732, opened 2020-04-09
Implements Pitilakis et al., (2020) Site Amplification Model

g-weatherill:
Pitilakis et al., (2020) have published a small update to their 2018 design code based amplification model. This small update changes the coefficients only, so a little re-adjustment is made to both the `openquake.hazardlib.gsim.sera_amplification_models.PitilakisEtAl2018` class to allow the `PitilakisEtAl2020` class to change only the coefficient set.

The 2020 model is intended to be used as a potential alternative to, rather than replacement of, the 2018 model, so both models are retained at the author's request.


micheles:
There is a minor conflict now. For the rest, LGTM.

PR #5734, opened 2020-04-10
Add a second trt to the EventBasedPSHA demo

ptormene:
I just copied/pasted the existing source group, changing the tectonicRegion, the ids and the posList.


micheles:
Please update the changelog

PR #5739, opened 2020-04-14
Regenerated the outputs for engine 3.9

micheles:
Here are the problems discovered and solved.

- [x] added files for tot_losses and tot_curves
- [x] fixed mean disaggregation output (had the wrong shape)
- [x] the sigma_epsilon dataset in scenario was stored differently than in event based!


openquake/calculators/tests/outputs/sigma_epsilon.csv
```diff
@@ -1,3 +1,3 @@
-event_id,sig,eps
-0,2.600000E-01,-1.842593E+00
-1,2.600000E-01,3.478994E-01
+event_id,rlz_id,sig_inter_PGA,eps_inter_PGA
+10,1,NAN,NAN
```
raoanirudh, line 2: Are the NAN values expected?

openquake/calculators/tests/outputs/realizations.csv
```diff
@@ -1,2 +1,3 @@
 rlz_id,branch_path,weight
-0,"[FromShakeMap]",1.0000000e+00
+0,"[LinLee2008SSlab]\nminimum_distance = 10",6.0000002e-01
```
raoanirudh, line 2: Is the newline character expected here in the GMPE name?

openquake/calculators/tests/outputs/sigma_epsilon.csv
micheles: Yes, this happens for GMPE not defined for "inter" but only for "total".

openquake/calculators/tests/outputs/realizations.csv
micheles: Yes, because inside the realization output we do not store only the GMPE name but also the GMPE arguments. Think for instance to the GMPEs for Canada, which depends on an HDF5 file: they are all instances of GMPETable but with different arguments.

Issue #5748, opened 2020-04-16
Change the pointsource_distance behavior


micheles:
After discussion with the hazard team it was agreed that the current behavior is not optimal. So I propose the go back to the old syntax for the pointsource_distance
```
pointsource_distance = {trt: [(mag1, dist1), ..., (magN, distN)], ...}
```
However, there will be lots of shortcut notations, in particular
```
pointsource_distance = {trt: dist, ...}
```
will be used when the distances are all equal and independent from the magnitude. Moreover
```
pointsource_distance = dist
```
will mean use the same distance for all TRTs. We could also allow a
```
pointsource_distance = *
```
to let engine decide the magnitude-distance distribution to use, but that would NOT be the default.
The behavior would be as in the past (i.e. how @rcgee likes):
```
pointsource_distance = 100
```
would mean that beyond 100 km the point ruptures will be collapsed, independently from the magnitude. Instead something like this (that @mmpagani likes)
```
pointsource_distance = [(4, 20), (5.5, 50), (6.5, 100), (8, 200)]
```
would be a magnitude-dependent pointsource_distance function, defined in terms of `numpy.interp`.
For instance the distance at magnitude 6 would be
```
numpy.interp(6, [4, 5.5, 6.5, 8], [20, 50, 100, 200]) = 75 km
```

PR #5750, opened 2020-04-17
Changed pointsource_distance

micheles:
Closes https://github.com/gem/oq-engine/issues/5748.


openquake/qa_tests_data/classical/case_43/job.ini
```diff
@@ -34,7 +34,7 @@ investigation_time = 1.0
 intensity_measure_types_and_levels = {"PGA": logscale(0.005, 3.00, 20)}
 truncation_level = 3
 maximum_distance = 200
-pointsource_distance = 100
+pointsource_distance = *
```
mmpagani, line 37: We will need to document this  Michele otherwise the users will be lost.

micheles: Of course, but let us first see how this approach works out.

PR #5765, opened 2020-04-22
Save memory in large classical calculations

micheles:
The issue was that lots of contexts were kept in memory without need, even in the case of many sites. It is most likely that this will solve the issue with Australia.


mmpagani:
Bene!

PR #5766, opened 2020-04-22
Added test with pointsource_distance

micheles:
So that one can play with the magnitude-dependent pointsource_distance


openquake/qa_tests_data/classical/case_13/job.ini
```diff
@@ -3,6 +3,7 @@
 description = Classical PSHA QA test
 calculation_mode = classical
 random_seed = 23
+split_sources = false
```
mmpagani, line 6: Maybe you can add a short  comment on this parameter

micheles: It is a parameter introduced years ago used to avoid splitting the sources (in this test it reduces the number of spawned tasks and makes the test a bit faster).

PR #5780, opened 2020-04-28
Adding a new gsim by Zalachoris & Rathje (2019)

jasongrig:
Hi, 
I added the GMM by Zalachoris and Rathje (2019) [ZR19] titled "Ground Motion Model for Small-to-Moderate Earthquakes in Texas, Oklahoma, and Kansas". This GMM essentially modified, through a multiplication factor, the GMM by Hassani & Atkinson (2015 [HA15], which in turn had modified the BSSA14 GMM [Boore et al. 2014]. So in essence, ZR19 requires the other two GMM to run. Unfortunately when Zalachoris and Rathje were calibrating their GMM, they used one fewer decimal in some of the coefficients of BSSA14, so I had to modify the original coeff. table of BSSA14 to match the results of the verification tables provided by ZR19. I tried to improve the code using flake8 but I could not solve any more issues. It already took forever to understand how to solve the ones that I did solve. I will not be able to do more in terms of the syntax. 
Thanks,
Iason


mmpagani:
This PR is closed. Iason, is this intentional?

jasongrig:
I closed it because the request was made through the master and not through my fork. 
I am confused with how to use my fork with the terminal commands.

If you can please consider it open

thanks,
Iason

Issue #5786, opened 2020-04-30
Binary format for gridded sources


micheles:
The model for South America (SAM) contains 2 GB of gridded sources (i.e. nonparametric sources with gridded ruptures) in XML format. It would be better to use a binary format in this case. The format should be HDF5 for familiarity (it is already used by UCERF). The expected improvement in disk space is from 2 GB to 150 MB.

NB: even if we will introduce the new binary format, the old XML format will stay forever for backward compatibility and it will never be deprecated, so converters are needed.

Here is an idea for a mixed format:
```xml
<?xml version="1.0" encoding="utf-8"?>
<nrml xmlns="http://openquake.org/xmlns/nrml/0.5" xmlns:gml="http://www.opengis.net/gml">
    <sourceModel investigation_time="1." name="example source model" ">
        <sourceGroup tectonicRegion="Subduction IntraSlab">
            <nonParametricSeismicSource id="src_col" name="Gridded src_col" />
        </sourceGroup>
    </sourceModel>
</nrml>
```
Implementing this feature has impact on

- [x] sourceconverter
- [x] sourcewriter
- [x] oq zip

PR #5793, opened 2020-05-04
HDF5 format for gridded sources

micheles:
Closes https://github.com/gem/oq-engine/issues/5786. A few notes:

1. we are totally backward compatible: the old format will be read just fine
2. when creating new source models the sourcewriter will use the new format .xml+.hdf5
3. oq zip will keep working without missing the .hdf5 files

The disk space saving for the SAM model is 13x, while the speedup in reading is only 30% (it looks like the time was not spent in XML parsing/validation).


julgp:
LGTM too

Issue #5789, opened 2020-04-30
Amplification of GMFs: add check


micheles:
We want to avoid this error:
```python
  File "/opt/openquake2/oq-engine/openquake/calculators/base.py", line 795, in _read_risk_data
    self.amplifier = Amplifier(
  File "/opt/openquake2/oq-engine/openquake/hazardlib/site_amplification.py", line 118, in __init__
    l_indices = digitize('level', self.midlevels, imls)
  File "/opt/openquake2/oq-engine/openquake/hazardlib/site_amplification.py", line 66, in digitize
    if max(values) > max(edges):
ValueError: max() arg is an empty sequence
```
See calculation 38173 on cluster2.

micheles:
I cannot reproduce the error, @CatalinaYepes : you changed the files and now running /home/cyepes/scenarios_colombia/Jobs/job_sgc_ampl_Bog.ini works!

CatalinaYepes:
To reproduce the error the files are in the `nonlinear_factors` branch.

PR #5813, opened 2020-05-08
Raised a clear error for amplification calculations with missing intensity measure types and levels

micheles:
Closes https://github.com/gem/oq-engine/issues/5789.


CatalinaYepes:
I don't think the `intensity measure types and levels` field should be required for event based risk calculations. See https://github.com/gem/oq-engine/issues/5816

PR #5832, opened 2020-05-14
Add disagg to manual [skip CI]

kejohnso:
adding new disaggregation parameters to the manual


doc/manual/oqum/hazard/01d_config_disaggregation.tex
```diff
@@ -40,6 +40,8 @@
 coordinate_bin_width = 1.5
 num_epsilon_bins = 3
 disagg_outputs = Mag_Dist_Eps Mag_Lon_Lat
+num_rlzs_disagg = 3
+rlz_index = 22,23
 \end{minted}
```
micheles, line 45: This is wrong, @kejohnso !
The options `num_rlzs_disagg` and `rlz_index` are mutually exclusive, you cannot set them both!

kejohnso: Yes, I explain this in line 103. 

I will move the explanation of rlz_index to a separate block, like is done for iml_disagg

micheles:
BTW, the travis tests are red in this branch. I think the reason is that you did not refresh it with the current master, please try that.

Issue #5802, opened 2020-05-05
Event based damage outputs


raoanirudh:
The new `event_based_damage` calculator is using exporters designed for the `scenario_damage` calculator. However, in the event based calculators, the average must be taken over the effective investigation time. In the scenario calculators instead, the average is taken over the `number_of_ground_motion_fields`. The `scenario_damage` post-processor and exporters cannot thus be used directly for the `event_based_damage` calculator.

* [x] `dmg_by_asset` output:
  * [x] This output should be renamed to `avg_damages` instead of `dmg_by_asset`, in line with the `avg_losses` of `event_based_risk` and `ebrisk`.  Pseudocode for calculating the `avg_damages` output is listed in the specifications issue for the `event_based_damage` calculator: https://github.com/gem/oq-engine/issues/5339
  * [x] Should not contain the stdv fields. The header in this case should simply be: `asset_id,city,county,state,taxonomy,tract,zip,lon,lat,contents~no_damage,contents~ds1,contents~ds2,contents~ds3,contents~ds4,nonstructural~no_damage,nonstructural~ds1,nonstructural~ds2,nonstructural~ds3,nonstructural~ds4,structural~no_damage,structural~ds1,structural~ds2,structural~ds3,structural~ds4`.
  * [x] This output is missing the metadata header line. 
* [x] `losses_by_asset` output:
  * [x] This output should be renamed to `avg_losses` instead of `losses_by_asset`.
  * [x] This output is essentially the same as the `avg_losses-rlz` output for the `event_based_risk` calculator. So the fields such as `structural~mean,structural~stddev` which are relevant for `scenario_damage` and `scenario_risk` should not appear in this output. Instead, the fields should simply be `structural`, as in the `avg_losses-rlz` output.

micheles:
I would propose instead of making the outputs of scenario_damage more similar to the outputs of event based. It would be enough to remove the stddev fields in scenario_damage. The user can still study the stability of the means by changing the seed.

PR #5848, opened 2020-05-19
Computed the averages correctly in even based damage

micheles:
Closes https://github.com/gem/oq-engine/issues/5802.


openquake/calculators/scenario_damage.py
```diff
@@ -200,11 +200,19 @@ def post_execute(self, result):
         events_per_asset = (indices[:, 1] - indices[:, 0]).mean()
         logging.info('Found ~%d dmg distributions per asset', events_per_asset)
 
+        # avg_factor = factor used when computing the averages
+        oq = self.oqparam
+        if oq.investigation_time is None:  # scenario_damage
```
raoanirudh, line 205: If the job file for a scenario_damage calculation includes the `investigation_time` inadvertently, do we ignore it when reading the oq params?

micheles: I would raise an error in that case, what do you think?

raoanirudh: That should work

> I would raise an error in that case, what do you think?

That should work

micheles: Done

PR #5859, opened 2020-05-22
Collapsing nonparametric ruptures

micheles:
I finally understood how to collapse nonparametric ruptures. The feature is implemented and tested in CollapseTestCase, however it is disabled by default (as all collapsing features). I have also changed the storage of `probs_occur` to use 64 bit floats, since numeric errors are pretty noticeable in disaggregation (like the negative probabilities of two days ago).

NB: the impact on the SGC disaggregation is a reduction of the time from 6_876s to 4_666s.


mmpagani:
I checked the code, particularly the `combine_po` function and I am happy with it and the underlying logic. I guess that the most costly component of collapsing np ruptures will be the selection of ruptures with a similar context.

PR #5878, opened 2020-05-28
Optimized the storage of the disaggregation outputs (less, bigger datasets)

micheles:
There are now 32,500 times less datasets in the SGC disaggregation. That improves the storing speed by a factor of 10 :-)


julgp:
LGTM, really, really nice...

PR #5884, opened 2020-05-29
Changed disagg_layer to work in terms of return periods in the field names

micheles:
Notice that the hazard curves/UHS are still in terms of PoEs as before, so I am not sure if we want to do this.


micheles:
Let's close this for the moment, to not add entropy.

Issue #5885, opened 2020-05-29
See if we can fix the minimum_magnitude feature for UCERF


micheles:
Both in event based and classical.

PR #5886, opened 2020-05-30
Fixed minimum_magnitude in UCERF

micheles:
Closes https://github.com/gem/oq-engine/issues/5885.
The `minMag` in the XML file was taking precedence over the `minimum_magnitude` in the job.ini and therefore the feature was not working, even for background sources :-(


mmpagani:
Grazie Michele. LGTM!

PR #5910, opened 2020-06-10
Restored scenario_damage with fractional asset number

micheles:
As requested by @raoanirudh


micheles:
Thanks Anirudh, I never remember about the ``%g`` formatting.

PR #5934, opened 2020-06-17
Add demo for event_based_damage calculator #5933

CatalinaYepes:
A demo for Nepal featuring the recent event_based_damage calculator is included.

Before merging, checks are needed in the outputs!


micheles:
We can merge now because the no damage bug has been fixed in https://github.com/gem/oq-engine/pull/5936

PR #5954, opened 2020-06-25
Update minimum_intensity documentation in manual

### LINKED ISSUE #2110: Apparently minimum_intensity does not work as explained in the manual
So change the way it works, if needed.

NB: the manual says that the first non-zero level is taken from the vulnerability functions, instead right now the engine is taking the first level as minimum_intensity, even if it is zero.

raoanirudh:
Bring the documentation in the manual for the default value of the `minimum_intensity` parameter in line with how the engine currently handles it. Closes #2110


micheles:
Ah, good, thank you @raoanirudh

PR #5964, opened 2020-07-02
Create f19

S-Hadi:


S-Hadi:
[farajpour_etal_2019_2.zip](https://github.com/gem/oq-engine/files/4864980/farajpour_etal_2019_2.zip)

PR #5973, opened 2020-07-07
Added command `oq renumber_sm`

micheles:
Solves the issue of different sources with the same ID. The usage is `oq renumber_sm ssmLT.xml`. It is fast, less than 10 minutes to renumber the Australia model.


mmpagani:
Thanks Michele. This is very useful. LGTM

One question. Does this tool renumber the sources only if it's needed or it does it regardless? i.e. say that we have a PSHA input with 3 initial SSMs. Just one of them with 'wrong' IDs. Does the tool renumber everything?

micheles:
As it is now it renumbers everything regardless.

PR #5974, opened 2020-07-07
20200525 kernel

mmpagani:
This PR adds a preliminary and experimental version of the kernel method for computing PSHA using amplification functions. A following phase of more comprehensive testing is needed, but the scheme for integrating this method into the engine is almost complete. I prefer to merge the code to avoid that this branch diverts from the master.

This implementation computes automatically the IMLs on rock required to compute reliably hazard on soil using the IMTs and IMLs provided by the user in the `.ini` file. It also introduces a new parameter for the configuration file called `amplification_method` that can take two values `convolution` and `kernel`. See test cases 37, 49 and 50 in  https://github.com/gem/oq-engine/tree/20200705_fix_between/openquake/qa_tests_data/classical. 

The amplification function used by the `kernel` method requires two additional parameters `from_mag` and `from_rrup` (see example 50 and example below). Say that you have a rupture of magnitude 6.1 at 12 km. The amplification code searches for the magnitude-distance combination in the amplification function where `from_mag` and `from_rrup`are just below the values in question.

Amplification function example:
```
#,,,,,,,"vs30_ref=760"
ampcode,from_mag,from_rrup,level,PGA,SA(1.0),sigma_PGA,sigma_SA(1.0)
A,      4.0,      0.0,     0.001,1.0,1.0,    0.3,      0.2
A,      4.0,      0.0,     0.010,1.1,1.0,    0.3,      0.2
```


openquake/hazardlib/gsim/base.py
```diff
@@ -212,6 +220,148 @@ def _get_poes(mean_std, loglevels, truncation_level, squeeze=False):
     return _truncnorm_sf(truncation_level, out)
 
 
+from scipy.stats import norm
```
micheles, line 223: This looks like a duplication of hazardlib.site_amplification.norm_cdf. Why not importing that?

openquake/hazardlib/tests/gsim/base_site_test.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def test01(self):
+        import time
+
+        fname = gettemp(ampl_func)
+        df = read_csv(fname, {'ampcode': ampcode_dt, None: numpy.float64},
+                      index='ampcode')
+        df.reset_index(drop=False, inplace=True)
+        af = AmplFunction.from_compact(df)
+
```
micheles, line 79: The name `from_compact` is not clear to me. Why not `from_dframe`?

openquake/hazardlib/tests/gsim/base_site_test.py
```diff
@@ -0,0 +1,110 @@
...
+        # Shape: 2 x 4 (distances) x 2 (IMTs) x 2 (GMMs)
+        self.meastd = get_mean_std(sites, rup, dists, imts, [gmmA, gmmB])
+
+    def test01(self):
+        import time
+
+        fname = gettemp(ampl_func)
+        df = read_csv(fname, {'ampcode': ampcode_dt, None: numpy.float64},
+                      index='ampcode')
+        df.reset_index(drop=False, inplace=True)
```
micheles, line 77: Why it is necessary to reset the index?

openquake/hazardlib/tests/gsim/base_site_test.py
```diff
@@ -0,0 +1,110 @@
...
+
+        truncation_level = 3
+        sitecode = b'A'
+
+        imls_soil = numpy.log([0.012, 0.052, 0.12, 0.22, 0.52])
+        imls_soil = numpy.log(numpy.logspace(-2, 0, num=20))
+        imtls_soil = DictArray({'PGA': imls_soil, 'SA(1.0)': imls_soil})
+
+        # The output in this case will be (1, x, 2) i.e. 1 site, number
+        # intensity measure levels times 2 and 2 GMMs
```
micheles, line 88: I would just use the good old `time.time()` since there are no calls to `time.sleep()` below.

openquake/hazardlib/gsim/base.py
mmpagani: I have a circular import problem if I do:
```
from openquake.hazardlib.site_amplification import norm_cdf
```

openquake/hazardlib/tests/gsim/base_site_test.py
mmpagani: Done

openquake/hazardlib/gsim/base.py
micheles: The solution is to move norm_cdf in another module (perhaps in hazardlib.stats, since it is a statistical function?)

openquake/hazardlib/tests/gsim/base_site_test.py
mmpagani: I cleaned this part. Since tracking execution time is probably not needed in the test.

openquake/hazardlib/tests/gsim/base_site_test.py
mmpagani: Leaving the index as is, it is giving me problems i.e. when I call `from_dframe` at line 70 of `site_amplification.py` (see https://github.com/gem/oq-engine/blob/20200525_kernel/openquake/hazardlib/site_amplification.py#L70) I get ```
KeyError: 'ampcode' ```

openquake/hazardlib/gsim/base.py
mmpagani: I have a circular import problem.

openquake/hazardlib/tests/gsim/base_site_test.py
micheles: I see, then just add a comment explaining this.

openquake/hazardlib/gsim/base.py
mmpagani: Done

openquake/hazardlib/gsim/base.py
```diff
@@ -212,6 +221,123 @@ def _get_poes(mean_std, loglevels, truncation_level, squeeze=False):
...
+    clm = 0
+    for m, imt in enumerate(loglevels):
+
+        # Here we set automatically the IMLs that will be used to compute
+        # the probability of occurrence of GM on rock within discrete
+        # intervals
+        sigma = ampfun.get_max_sigma()
+        ll = numpy.linspace(min(soillevels)-sigma*4.,
+                            max(soillevels)+sigma*4., num=nsamp)
+
```
micheles, line 283: There lines can go outside the loop, right?

openquake/hazardlib/site_amplification.py
```diff
@@ -16,36 +16,121 @@
...
+
+        return median, std
+
+    def get_max_sigma(self):
+        """
+        :returns:
+            The maximum sigma value in the amplification function
+        """
+        # TODO - this must be improved
+        # return max(self.df['std'])
```
micheles, line 132: Not clear why the 0.3 is hard coded here and why `max(self.df['std'])` is commented out

mmpagani: Fixed

openquake/hazardlib/gsim/base.py
mmpagani: I fixed this part since the IMLs can vary depending on the IMT. We can move out just the call to get_max_sigma.

openquake/hazardlib/site_amplification.py
```diff
@@ -131,15 +261,32 @@ def amplify_one(self, ampl_code, imt, poes):
...
+        #
+        # Amplify, for each site i.e. distance
         for g in range(G):
+            # Compute the probability of occurrence of GM within a number of
+            # intervals
             p_occ = -numpy.diff(poes[:, g])
+            #
+            # Setting the amplification function
+            if mag is not None and dst is not None:
+                1/0  # ERROR: it never goes here
```
micheles, line 283: It never enters here because in openquake/calculators/classical.py the call is `pcurves = amplifier.amplify(ampcode[sid], pcurves)` without passing magnitude and rrup (which is correct for the convolution).

PR #5975, opened 2020-07-08
20200705 fix between

mmpagani:
This PR introduces a generalised implementation of a modifiable GMPE. The only method added so far is a method that fixes the between-event variability and replaces the total standard deviation with the within-event one. It can be used for scenario risk analysis when there is a will to better constrain ground-motion, ideally using information from recorded shaking. 

An example for the definition of this modifiable GMPE is available below (see also case_51 in https://github.com/gem/oq-engine/tree/master/openquake/qa_tests_data/classical).
```
<uncertaintyModel>
    [ModifiableGMPE]
        gmpe_name="AkkarEtAlRjb2014"
        params.mod1.meth="set_between_epsilon"
        params.mod1.params.epsilon_tau=0.5
</uncertaintyModel>
```

The parameter `gmpe_name` defines the original GMPE. The first modification is specified with these two lines:
```
params.mod1.meth="set_between_epsilon"
params.mod1.params.epsilon_tau=0.5
``` 
where `params.mod<modification_number>.meth` is mandatory (it specifies the modification requested) while 
```
params.mod1.params.<name_variable_1> = 1.0
params.mod1.params.<name_variable_2> = 1.0
```
etc. will contain all the parameters required by this specific modification.


micheles:
The problem here is that I do not like the syntax. Why not something like this?
```

<uncertaintyModel>
    [ModifiableGMPE]
        gmpe_name="AkkarEtAlRjb2014"
        apply = [{meth="set_between_epsilon", epsilon_tau=0.5}]
</uncertaintyModel>
```
`apply` is a list of dictionaries with a key `meth` (the name of the method to apply) and optionally other keys (the parameters to pass).

micheles:
Having slept on that, I remember how it works. The JSON-syntax for a callable with arguments (including a GMPE) is `{'callable': {'arg1': val1, 'arg2': val2, ...}}` which nicely translates to TOML. Therefore it would be
```xml
<uncertaintyModel>
    [ModifiableGMPE]
gmpe = "AkkarEtAlRjb2014"
set_between_epsilon = {epsilon_tau=0.5}
</uncertaintyModel>
```
with GMPE without arguments and

```xml
<uncertaintyModel>
    [ModifiableGMPE]
gmpe.GMPETable = {gmpe_table="fname.hdf5"}
set_between_epsilon = {epsilon_tau=0.5}
</uncertaintyModel>
```
for a GMPE with arguments like the GMPETable. If you like this, I can implement the necessary changes.

PR #5976, opened 2020-07-08
Fixed terrible regression in classical_damage

micheles:
The fix in https://github.com/gem/oq-engine/pull/5946 did not really solve the problem with the minIML feature due to a missing copy of the intensity measure levels :-( This was terribly hard to find, the effect was having different damages with OQ_DISTRIBUTE=no in case_master. Fortunately the bug affected only master.
There is also some small refactoring of the calculator and a simplification of the test case_master.


openquake/qa_tests_data/classical_damage/case_master/job_haz.ini
```diff
@@ -28,7 +28,7 @@ random_seed = 24
 truncation_level = 3
 maximum_distance = 200.0
 investigation_time = 1
-number_of_logic_tree_samples = 0
+number_of_logic_tree_samples = 1
```
raoanirudh, line 31: The case_master needs to have multiple branches in order to generate the statistics outputs. We should preferably keep the full enumeration.

PR #5990, opened 2020-07-13
Extracting portions of the site collection

micheles:
Continuation of https://github.com/gem/oq-engine/pull/5980. You should call /extract/42/sitecol?field=zipcode

### DIFF
--- a/openquake/calculators/extract.py
+++ b/openquake/calculators/extract.py
@@ -713,8 +713,8 @@ def extract_agg_curves(dstore, what):
 def extract_agg_losses(dstore, what):
     """
     Aggregate losses of the given loss type and tags. Use it as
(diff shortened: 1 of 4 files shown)


openquake/calculators/extract.py
```diff
@@ -380,8 +380,11 @@ def extract_sitecol(dstore, what):
     """
     Extracts the site collection array (not the complete object, otherwise it
     would need to be pickled).
-    Use it as /extract/sitecol
+    Use it as /extract/sitecol?param=vs30
```
ptormene, line 383: I guess 'zipcode' should be written here instead of vs30

micheles: Yes, but zipcode normally is not available, while the vs30 is always there, so it makes a good example.

PR #6041, opened 2020-08-22
Added Stewart et al. (2016) GMPE

ftbernales:
Implements the SBSA15 GMPE by Stewart et al. (2016) for vertical-component ground motions from the PEER NGA-West2 Project (2 of 4).

This is largely similar in form to the BSSA14 horizontal-component model by Boore et al. (2014), so I called some methods therein, as needed.

I used the FORTRAN code published by Dave Boore in his website to generate the suite of test data for verification. I shall send to you these files in a separate email for your reference.


mmpagani (approved):
As I suggested for the PR I reviewed earlier today, for this implementation it might be useful to extend the BSSA14 rather than the `GMPE` class. Other than this, I am fine with this implementation.

mmpagani:
Pls add this model in the .rst file https://github.com/gem/oq-engine/blob/master/doc/sphinx/openquake.hazardlib.gsim.rst

mmpagani:
Pls also update the changelog so that your contribution is recognized! https://github.com/gem/oq-engine/blob/master/debian/changelog#L12

PR #6072, opened 2020-09-16
Implementation of Kuehn et al 2020 NGA-Sub model

cvanhoutte:
Doesn't yet include the regional models where there is a 'fraction of backarc' term, but the NZ model is included.


cvanhoutte:
Sorry, merging into the wrong master!

PR #6123, opened 2020-10-05
Psutil fix2

nastasi-oq:
Add fallback documentation to install missing, already built, dependencies if wheels are not provided on pypi for a specific system.


pslh:
See that [https://github.com/gem/oq-builders/pull/136](https://github.com/gem/oq-builders/pull/136 ) now resolves relative links in README.html for inclusion in packages, LGTM.

PR #6126, opened 2020-10-05
secondary perils: fixed critical accel formula

cossatot:
This PR fixes a bug I had in a formula for the Newmark critical acceleration. The bug was fixed on a branch I had originally in the MBTK but did not get merged before the code was pulled into the engine.


openquake/sep/landslide/newmark.py
```diff
@@ -2,6 +2,7 @@
 
 import numpy as np
 
+g: float = 9.81
```
raoanirudh, line 5: There's a `g` available through [scipy.constants](https://docs.scipy.org/doc/scipy/reference/constants.html)

micheles (approved):

PR #6160, opened 2020-10-15
Unified avg_damages and damages

micheles:
And fixes master that was broken by https://github.com/gem/oq-engine/pull/6158.


raoanirudh:
@micheles, recall https://github.com/gem/oq-engine/pull/5851#pullrequestreview-415537915

PR #6183, opened 2020-10-22
Chao20

viktor76525:
Implemented Chao et al. 2020


openquake/hazardlib/gsim/chao_2020.py
```diff
@@ -0,0 +1,307 @@
...
+        const.StdDev.TOTAL,
+        const.StdDev.INTER_EVENT,
+        const.StdDev.INTRA_EVENT
+    ])
+
+    REQUIRES_DISTANCES = {'rrup'}
+
+    REQUIRES_RUPTURE_PARAMETERS = {'mag', 'ztor'}
+
+    REQUIRES_SITES_PARAMETERS = {'vs30', 'vs30measured', 'z1pt0'}
```
mmpagani, line 54: Would be good adding also the reference rock conditions as done here: https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/gsim/abrahamson_2014.py#L82

mmpagani (approved):
Apart from a minor comment LGTM

openquake/hazardlib/gsim/chao_2020.py
viktor76525: fixed

PR #6373, opened 2020-12-18
Update openquake.hazardlib.gsim.rst

hasanmousavi:
I have updated the openquake.hazardlib.gsim.rst
I wish the problem has been solved


mmpagani:
As previous one.

PR #6374, opened 2020-12-18
Update openquake.hazardlib.gsim.rst

hasanmousavi:


mmpagani:
This PR is not needed. Please move these changes in the PR that's already open.

PR #6481, opened 2021-01-22
Updated the documentation about installing the engine for development

micheles:
By suggesting to use install.py


vot4anto:
LGTM

PR #6526, opened 2021-02-09
20210209 multi fault source opt2

mmpagani:
This is an alternative option to what's implemented in a companion branch `20210209_multi_fault_source`. 
Todo:
- [x] Add missing methods
- [ ] Add modifications required to support a model a la UCERF
- [ ] Discuss modifications that might be useful for Thomas


mmpagani:
See https://github.com/gem/oq-engine/pull/6723

Issue #6602, opened 2021-03-11
Error in post_risk when running event_based_risk from a previous event_based calculation


nicolepaul:
When using `oq engine --run job.ini --hc <hazard_calc_id>` with the **event_based_risk** calculator and a previously computed **event_based** hdf5, I got the following error:

```
[2021-03-11 12:44:26 #40996 INFO] Stopping the workers
Traceback (most recent call last):
  File "/Users/npaul/openquake/bin/oq", line 8, in <module>
    sys.exit(oq())
  File "/Users/npaul/openquake/lib/python3.6/site-packages/openquake/commands/__main__.py", line 44, in oq
    sap.run(commands, prog='oq')
  File "/Users/npaul/openquake/lib/python3.6/site-packages/openquake/baselib/sap.py", line 225, in run
    return _run(parser(funcdict, **parserkw), argv)
  File "/Users/npaul/openquake/lib/python3.6/site-packages/openquake/baselib/sap.py", line 216, in _run
    return func(**dic)
  File "/Users/npaul/openquake/lib/python3.6/site-packages/openquake/commands/engine.py", line 164, in main
    run_jobs(job_inis, log_level, log_file, exports, **pars)
  File "/Users/npaul/openquake/lib/python3.6/site-packages/openquake/engine/engine.py", line 407, in run_jobs
    run_calc(*args)
  File "/Users/npaul/openquake/lib/python3.6/site-packages/openquake/engine/engine.py", line 263, in run_calc
    calc.run(exports=exports, **kw)
  File "/Users/npaul/openquake/lib/python3.6/site-packages/openquake/calculators/base.py", line 250, in run
    self.post_execute(self.result)
  File "/Users/npaul/openquake/lib/python3.6/site-packages/openquake/calculators/scenario_risk.py", line 188, in post_execute
    prc.run(exports='')
  File "/Users/npaul/openquake/lib/python3.6/site-packages/openquake/calculators/base.py", line 248, in run
    self.result = self.execute()
  File "/Users/npaul/openquake/lib/python3.6/site-packages/openquake/calculators/post_risk.py", line 181, in execute
    alt_df['agg_id'] = idxs[alt_df['agg_id'].to_numpy()]
IndexError: index 4 is out of bounds for axis 0 with size 2
```

This occurs after the `event_based_risk` has successfully completed and at the start of `post_risk`. There is no error if the `aggregate_by` parameter is not specified.

I'm attaching a simple example model that replicates the issue.
[...]

PR #6613, opened 2021-03-15
Fixed bug with aggregate_by and --hc

micheles:
Closes https://github.com/gem/oq-engine/issues/6602


nicolepaul (commented):
Successfully ran on 3.11.2, thanks!

PR #6640, opened 2021-03-22
remove CMD from Dockerfile.dev

vot4anto:


vot4anto:
The build using new Dockerfile is successfull within the github action

PR #6646, opened 2021-03-24
Refactoring/ground motion maps

schmidni:
I will be expanding on the existing code, therefore I would like to have the logic separated:

read&parse: parsers.py
data check & site assoc: maps.py
calculations & sampling: gmfs.py

also a little fix, where get_sitecol_shakemap has a return case which only returns 2 arguments instead of three. Its only used for tests for now, but of course I tried someting and fell right into that trap.

apart from that there is NO change in functions or logic, only copy-paste and adjusting references


schmidni:
I don't understand it, I cant reproduce that test locally?

image

micheles:
That damn not reproducible and intermittent error on validate_zip has been bugging me for a week. It has nothing to do with your changes, so do not worry.

micheles (approved):

PR #6649, opened 2021-03-25
first deployment for kubernetes master and worker

vot4anto:


vot4anto:
@micheles set the original timeout for zeromq.py at 15000

PR #6654, opened 2021-03-26
Refactor to_gmfs and faster calculation method

schmidni:
Refactoring with the purpose of making the code extendable.
- some simple refactoring of the method to make it more readable
- added docstrings
- created CallableDict "gmfs_calculation_methods" which returns a function. This function contains the specific implementation how the gmfs are calculated, when given the matrix of mu (mean) and Z (random variables).
I could, instead of returning a function, also just return the finished result, but this way the "calculation method" and the actual "calculation" are clearly separated.

Improvement:
I also added new calculation method, which directly calculates the sampling in case no correlation is needed. Until now Cholesky was computed every time, which is pointless, if the correlation matrixes are identity matrices.

### DIFF
--- a/openquake/hazardlib/shakemap/gmfs.py
+++ b/openquake/hazardlib/shakemap/gmfs.py
@@ -152,11 +152,38 @@ def cholesky(spatial_cov, cross_corr):
     return numpy.linalg.cholesky(numpy.array(LLT))
 
 
-sample_gmfs = CallableDict()
+gmfs_calculation_methods = CallableDict()
+
+MAXSITES = 1000
+CORRELATION_MATRIX_TOO_LARGE = '''\
+You have a correlation matrix which is too large: %s > %d.
+To avoid that, set a proper `region_grid_spacing` so that your exposure
+involves less sites.'''
 
 
-@sample_gmfs.add('Silva&Horspool')
-def sample_gmfs_SH(kind, shakemap, imts, spatialcorr, crosscorr):
+@gmfs_calculation_methods.add('Silva&Horspool')
+def gmfs_calculation_methods_SH(kind, shakemap, imts, spatialcorr,
+                                crosscorr, cholesky_limit):
+    """
+    Implementation of paper by Silva and Horspool 2019
+    https://onlinelibrary.wiley.com/doi/abs/10.1002/eqe.3154?af=R
+
+    :param shakemap: site coordinates with shakemap values
+    :param imts: list of required imts
+    :param spatialcorr: 'no', 'yes' or 'full'
+    :param crosscorr: 'no', 'yes' or 'full'
+    :returns: F(Z, mu) to calculate gmfs
+    """
+    # checks
+    N = len(shakemap)
+    M = len(imts)
+    if spatialcorr != 'no' and N > MAXSITES:
+        # hard-coded, heuristic
+        raise ValueError(CORRELATION_MATRIX_TOO_LARGE %
+                         (N, MAXSITES))
+    if N * M > cholesky_limit:
+        raise ValueError(CORRELATION_MATRIX_TOO_LARGE % (
+            '%d x %d' % (M, N), cholesky_limit))
 
     # Cross Correlation
     cross_corr = cross_correlation_matrix(imts, crosscorr)
@@ -175,6 +202,7 @@ def sample_gmfs_SH(kind, shakemap, imts, spatialcorr, crosscorr):
     # Cholesky Decomposition
     L = cholesky(spatial_cov, cross_corr)  # shape (M * N, M * N)
 
+    # mu has unit (pctg), L has unit ln(pctg)
     def generate_gmfs(Z, mu):
         mu = numpy.log(mu)
         return numpy.exp(L @ Z + mu) / PCTG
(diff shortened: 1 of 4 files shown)


openquake/hazardlib/shakemap/gmfs.py
```diff
@@ -151,41 +152,131 @@ def cholesky(spatial_cov, cross_corr):
...
+    # checks
+    N = len(shakemap)
+    M = len(imts)
+    if spatialcorr != 'no' and N > MAXSITES:
+        # hard-coded, heuristic
+        raise ValueError(CORRELATION_MATRIX_TOO_LARGE %
+                         (N, MAXSITES))
+    if N * M > cholesky_limit:
+        raise ValueError(CORRELATION_MATRIX_TOO_LARGE % (
+            '%d x %d' % (M, N), cholesky_limit))
```
schmidni, line 186: I understand the cholesky_limit check. 

But why did you add the check for the "MAXSITES" here, seeing that it is hardcoded. If it is for performance reasons when computing the spatial correlation, then the improvements from #6624 should have fixed this.

micheles: It should not be hard-coded but configurable like the cholesky_limit. Consider it at technical debt.
It is not there for performance, but to avoid small negative eigenvalues due to numerical errors. It would be nicer to have a real solution rather than the ugly workaround :-(

schmidni: But why is only checking N*M not enough?

micheles (commented):
Because in the formula `numpy.linalg.cholesky(LLT)` the matrix LLT has shape (N*M, N*M)

openquake/hazardlib/shakemap/gmfs.py
micheles: Because in the formula numpy.linalg.cholesky(LLT) the matrix LLT has shape `(N*M, N*M)`

schmidni: I am sorry, I think I am a bit slow.
You are checking `if N*M > cholesky_limit:` - so why do you also check `if N > MAXSITES`?

If cholesky_limit = 10000 (default) and MAXSITES = 1000 (hardcoded) then:

N = 4000 and M = 1 => ValueError
N = 999 and M = 5 => no Value Error
even though N*M is bigger in the second example.

openquake/hazardlib/shakemap/gmfs.py
```diff
@@ -151,41 +152,131 @@ def cholesky(spatial_cov, cross_corr):
...
     spatial_cov = spatial_covariance_array(stddev, spatial_corr)
+
+    # Cholesky Decomposition
     L = cholesky(spatial_cov, cross_corr)  # shape (M * N, M * N)
+
+    # mu has unit (pctg), L has unit ln(pctg)
+    def calculate_gmfs(Z, mu):
+        mu = numpy.log(mu)
+        return numpy.exp(L @ Z + mu) / PCTG
+
```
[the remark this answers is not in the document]
schmidni: every call to `gmfs_calculation_methods` returns a function (`calculate_gmfs`). In these two cases the function looks the same (except that one has L@Z+mu and the other one sig@Z+mu). But I will add for example something along those lines:

`@gmfs_calculation_methods.add('MMI')
def gmfs_calculation_methods_basic(kind, shakemap, imts):
    # create diag matrix with std values
    std = numpy.array(shakemap['std']['MMI'])
    sig = numpy.diag(std.flatten())  # shape (M*N, M*N)

    # mu has unit (pctg), sig has unit ln(pctg)
    def calculate_gmfs(Z, mu):
        return sig @ Z + mu

    return calculate_gmfs`

openquake/hazardlib/shakemap/gmfs.py
micheles: The MAXSITES check was there before, introduced years ago, I just added the additional check. Perhaps we can fully remove the check on MAXSITES and instead add a check that there are no negative eigenvalues. Thanks to the cholesky_limit we know that the decomposition will still work without running out of memory.

openquake/hazardlib/shakemap/gmfs.py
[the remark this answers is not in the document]
micheles: I do not like that `gmfs_calculation_methods_basic` returns a function. It should just return the result. If you need the function you can just call `gmfs_calculation_methods_basic[kind]`. BTW, I do not like the name `gmfs_calculation_methods_basic`. It should be something like `shakemap_to_gmfs`.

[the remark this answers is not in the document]
schmidni: I don't think this helps readability and gmf_dict is huge by now since I also have to put in Z and mu. 

But you're the boss, I made the changes as requested.

openquake/hazardlib/shakemap/gmfs.py
schmidni: sorry I went a bit off-topic here. I resolve this conversation and leave you to accept / ask for additional changes in the pull request.

micheles (approved):

PR #6743, opened 2021-05-03
Requirements and GDAL

vot4anto:
remove packages from requirements-extra and use new GDAL wheel


micheles (approved):

vot4anto:
hazardlib + engine tests / engine (ubuntu-latest, 3.7) Fails because on master the requirements files for python 3.7 have the wrong version of wheel for h5py

PR #6900, opened 2021-06-27
Corrected Campbell and Bozorgnia (2014) GMPE

stansays:
## Few corrections to Campbell and Bozorgnia (2014) GMPE were made:

1. Cutoff spectral period for setting Sa (T) mean to PGA should be exclusive of 0.25 s, as stated in the Campbell & Bozorgnia (2013) NGA West 2 Report Equation 1, and noted in the comment in line 108:
> \# If Sa (T) < PGA for T < 0.25 then set mean Sa(T) to mean PGA

2. The c<sub>13</sub> term in _f_<sub>site,J</sub> is used regardless, even if V<sub>S30</sub> is less than 200 m/s (i.e. All V<sub>S30</sub>), as shown in Equation 19 of the paper:
CB14_fsiteJ
This is also how it is implemented in the Excel file provided as supplementary material in the paper for the V/H model by the same authors, Campbell and Bozorgnia:
[https://journals.sagepub.com/doi/suppl/10.1193/100614eqs151m](url)

3. The spectral period of 0.075 s was set to the same precision as used by the paper and Excel file, instead of 0.08 s, to reduce discrepancy.

### The tests were updated in the following aspects: 

1. Follows the Excel file implementation by Campbell and Bozorgnia with a correction in Basin Response (_f_<sub>sed</sub>) term in computing the mean and A<sub>1100</sub>, found in Cell A62 and column L. _f_<sub>sed</sub> should refer to the S<sub>j</sub> flag instead of the S<sub>ji</sub> flag.
A<sub>1100</sub> should also refer to Z<sub>2.5</sub> (VS<sub>30</sub>=1100) in Cell B48, not constant in Cell A48.
2. Tests now include all specific spectral periods considered in the paper. Note the previous tests did not consider the edge case of spectral period 0.25 s.
3. Removed 675 duplicate test rows (e.g. eighth test row was a duplicate of the first test row).


mmpagani (approved):
Thanks for this PR. I am fine with the proposed changes. We will address the changes needed to make this model compatible with the new GMM implementation.

PR #7021, opened 2021-07-22
update DEFINED_FOR_REFERENCE_VELOCITY

rcgee:
This PR updates the DEFINED_FOR_REFERENCE_VELOCITY values of 4 GMPEs since some of the current values are not justifiable ([this commit](https://github.com/gem/oq-engine/commit/0b713d9bab020879780453703a52ab42c6005882#diff-03fa5f1413f39120f86e8dd4404c19551a75ad3c00daa7b42397b16983f3265d)) . In doing so, it also allows calculations using amplification functions to be run, which require the DEFINED_FOR_REFERENCE_VELOCITY parameter.

NB: AB03 and LL08 are applicable to NEHRP site class B

Many GMPEs don't have a clear reference velocity, so this can be a challenging parameter to define. Going forward we will need to think about the best way to handle this for all GMPEs.


mmpagani (approved):
LGTM. Please open an issue on the `DEFINED_FOR_REFERENCE_VELOCITY` parameter

PR #7049, opened 2021-07-29
20201025 planar

rcgee:
This PR adds a method for creating a PlanarSurface given a hypocenter location and other rupture properties. This can be useful for computing ground motion for a given rupture scenario (for example).


mmpagani (approved):
LGTM

PR #7071, opened 2021-08-10
Abrahamson & Gulerce (2020) - backwards compatibility

SnowNooDLe:
# Abrahamson & Gulerce (2020) - backwards compatibility

This PR is related to this issue - #7067 

We noticed this error by passing either a float or the length of 1 numpy.ndarry for `ztor` argument.
(But as the docstring said, within `get_rupture_depth_scaling_term()`, we should always pass a numpy.ndarray instead of a float.)

As a comment was left in the specified issue, implemented backwards compatibility.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,6 @@
+  [Jason Motha/Tom Son]
+  * Implemented backwards compatibility to Abrahamson & Gulerce (2020)
+
   [Michele Simionato]
   * `oq info gsim_logic_tree.xml` now displays the logic tree
   * Fixed a bug in the adjustment term in NSHMP2014 breaking the USA model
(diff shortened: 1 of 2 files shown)


g-weatherill:
This seems like a sensible fix to me. So LGTM from my side.

micheles:
You need to show a case where the previous implementation was broken and the fix works

SnowNooDLe:
> You need to show a case where the previous implementation was broken and the fix works

@micheles Ah, definitely! I will update what we used to get(as an error) and what we get now tomorrow.

SnowNooDLe:
@micheles I should've checked with the updated version first before I opened this PR but.
I was trying to reproduce the error we used to get but I noticed there was a change, like 3 days ago. We got this error on Monday morning and noticed there was another person who had the same issue.
However, with recent  PR, #7068 already solved this problem by converting float `ztor` to numpy.ndarray type.

Hence, I will just close this PR. Thanks for letting me check tho!

## Based on the older version, passing float
Screenshot from 2021-08-12 16-39-48

## After the implementation (This PR)
Screenshot from 2021-08-12 16-57-31

PR #7402, opened 2021-11-23
Fixed the upgrade action

micheles:
Fixes the error in https://github.com/gem/oq-engine/runs/4299246097?check_suite_focus=true


vot4anto (commented):
LGTM

PR #7461, opened 2021-12-20
Added GSIM - Jaimes et al. (2020)

mleonardos:
Implementation of the GMPE derived by Jaimes et al., (2020) for Mexican intermediate-depth intraslab earthquakes.


mmpagani (approved):
LGTM. Thank you and congrats for the good work.

PR #7503, opened 2022-01-21
Restored aggrisk and aggcurves statistics

micheles:
As requested by the risk team.


raoanirudh (commented):
The aggrisk-stats and aggcurves-stats outputs contain the fields 'agg_id' and 'loss_id' with both fields containing integers. It would be preferable to display the tags and loss categories as strings instead.

PR #7511, opened 2022-01-26
Moved the venv in /opt/openquake/venv

micheles:
The upgrade tests are green: https://github.com/gem/oq-engine/runs/4947299511?check_suite_focus=true


vot4anto (approved):
LGTM

PR #7626, opened 2022-03-04
Fixing the definition of the horizontal component in various GMMs + replacing an error with a warning

mmpagani:


openquake/hazardlib/gsim/mgmpe/modifiable_gmpe.py
```diff
@@ -122,9 +121,14 @@ def horiz_comp_to_geom_mean(self, ctx, imt):
                     conv_sigma = (C[2] + (C[3]-C[2]) *
                                   np.log10(T/0.15)/np.log10(0.8/0.15))
                 rstd = C[4]
+    elif comp in ['GEOMETRIC_MEAN']:
+        conv_median = 1
+        conv_sigma = 1
```
kejohnso, line 126: Is this correct? (also line 129). It means that the total_std is changed even if the median is not, and leads to invalid value in sqrt errors

openquake/hazardlib/gsim/mgmpe/modifiable_gmpe.py
```diff
@@ -84,7 +83,7 @@ def horiz_comp_to_geom_mean(self, ctx, imt):
 
     # List of the horizontal component definitions that can be converted into
     # geometric mean
-    tmp = ['AVERAGE_HORIZONTAL', 'GMRotI50', 'RANDOM_HORIZONTAL',
+    tmp = ['GEOMETRIC_MEAN', 'GMRotI50', 'RANDOM_HORIZONTAL',
            'GREATER_OF_TWO_HORIZONTAL', 'RotD50']
 
     # Apply the conversion
```
kejohnso, line 89: lines 78 and 79 should be moved inside the if statement to prevent failures when the horcom is not in COEFF and COEFF_PGA_PGV

openquake/hazardlib/gsim/mgmpe/modifiable_gmpe.py
mmpagani: You're right. convert_sigma must be = 0

openquake/hazardlib/gsim/mgmpe/modifiable_gmpe.py
```diff
@@ -122,9 +121,14 @@ def horiz_comp_to_geom_mean(self, ctx, imt):
...
+        conv_sigma = 1
     else:
+        conv_median = 1
+        conv_sigma = 1
         msg = f'Conversion not applicable for {comp}'
-        raise ValueError(msg)
+        warnings.warn(msg, UserWarning)
 
     # Original total STD
     total_stddev = getattr(self, const.StdDev.TOTAL)
```
kejohnso, line 134: line 137 fails when the `elif` or `else` is used, since rstd has not been defined. I think it should be 1 in these cases (but also see my comment on line 124)

mmpagani: yes rstd should be = 1

kejohnso:
LGTM!

micheles (approved):

PR #7744, opened 2022-04-25
Campbell and Bozorgnia (2014) - Added the Ztor value estimation

SnowNooDLe:
# Campbell and Bozorgnia (2014) - Added the `Ztor` value estimation

As the [Campbell and Bozorgnia (2014)](https://journals.sagepub.com/doi/pdf/10.1193/062913EQS175M) suggested(page 1109, 23/29), implemented the `Ztor` estimation.
Closes https://github.com/gem/oq-engine/issues/7697

image

Which can be found in [Chiou and Youngs (2014)](https://journals.sagepub.com/doi/pdf/10.1193/072813EQS219M), page 1125 (9/37), the equation 4 and 5.

image


micheles:
I like this. The breaking test is not your fault, just refresh from the current master.

SnowNooDLe:
Glad you like it.
I was waiting for your feedback on the failed tests with 3.9 and good to know that it could be fixed by refreshing it.
Let's see how this goes 👍

openquake/hazardlib/gsim/campbell_bozorgnia_2014.py
```diff
@@ -346,16 +346,33 @@ class CampbellBozorgnia2014(GMPE):
...
     SJ = 0  # 1 for Japan
 
+    def set_parameters(self, rup):
+        """
+        Estimate the rupture parameters if not provided.
+        """
+        frv = np.zeros_like(rup.rake)
+        frv[(rup.rake > 30.) & (rup.rake < 150.)] = 1.
+
+        if not hasattr(rup, "ztor"):
```
mmpagani, line 368: This tests is not required if you remove the ztor parameter from the REQUIRES_RUPTURE_PARAMETERS. Is this your preferred option? I think the cleanest solution would be to have two models, one that computes ztor automatically and one where ztor is computed by OQ for each rupture (i.e. the current behaviour).

SnowNooDLe: Hey,

The `REQUIRES_RUPTURE_PARAMETERS` no longer holds a parameter called `ztor`.
As the paper suggested, if the `ztor` is not provided, the model should be able to estimate the `ztor` from the given value, `mag.` but if it's suggested, we do not have to estimate them.
(`mag` is a required parameter, so we do not have to worry about not having `mag`.)

Hence, the `ztor` is now in a set called `REQUIRES_COMPUTED_PARAMETERS`.

[This](https://github.com/gem/oq-engine/pull/7554) was the idea that @micheles suggested.

SnowNooDLe: The only difference with the `in_cshm` from the link I provided above is that `in_cshm` always gets computed whereas `ztor` within this model, may or may not depends on whether a user has provided a `ztor` or not.

mmpagani: Thanks for this note. Let me explain my point. In a hazard analysis 'ztor' is never provided as an input to the OQ Engine. 'ztor' is a parameter that the engine computes automatically for every rupture together with all the other parameters in REQUIRES_RUPTURE_PARAMETERS. If you remove 'ztor' from the REQUIRES_RUPTURE_PARAMETERS list, the Context object will never contain 'ztor' and your proposed change will always be used. 

As I briefly mentioned in my previous note, the cleanest way to add this possibility (i.e. of using the 'ztor' suggested by CB14) is to create a separate model e.g. CampbellBozorgnia2014NoZTor. The use can then decide to use the original version with 'ztor' computed from the rupture or this new version with 'ztor' inferred.

SnowNooDLe: So we use our own wrapper that runs the method called `get_mean_stds()` from `contexts` module
```python
# ctx here is a RuptureContext
results = contexts.get_mean_stds(model, ctx, [im])
```

and if the `RuptureContext` object has an attribute `ztor`, this statement will not get fired because `ctx` has one
test-1

but if the `RuptureContext` object does not have one, then this statement gets triggered to estimate the `ztor` with the given `mag`.
test-2
test-3


I can see where the cleanest way of doing this comes from, and I agree. However, by splitting those, now we need to change the class name every time we want to use/test with/without `ztor` value. 
Currently, this PR is only about `ztor`, but my future PRs will include estimating the following parameters:
- width
- hypocentral depth

This code may not be the cleanest way of doing it, but from the user's point of view, they only need to care about the inputs they want to put.

E.g., We want to compute something with this model with `Ztor`, which we call `CampbellBozorgnia2014`, create a `RuptureContext`, and include `ztor`.
Then we now want to compute it without passing `Ztor`. All we have to do is, comment out the attribute called `ztor` instead of creating another class called `CampbellBozorgnia2014NoZtoR` or renaming the current class.

SnowNooDLe:
Hey, @micheles @mmpagani 

Just a follow-up comment regarding the style we want to have:
1. Keep the current behaviour - One class with an extra method called `set_parameters`
2. Create another class that will inherit from the `CampbellBozorgnia2014` class and add a method called `set_parameters` to do any required estimations.

I personally prefer to keep the current behaviour so that way, it's aligned with what the paper says, `Ztor` can be estimated if not provided by a user. (Instead of asking users to specifically select the right class, which will only happen if we decide to create another class.)

openquake/hazardlib/gsim/campbell_bozorgnia_2014.py
mmpagani: ```I can see where the cleanest way of doing this comes from, and I agree. However, by splitting those, now we need to change the class name every time we want to use/test with/without ztor value.``` This is not true if you set the way in which the model should operate at the instantiation level.

mmpagani:
The alternative is to provide flags at the instantiation level that 
control the way in which these parameters should be computed. An example 
is available here: 
https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/gsim/sgobba_2020.py#L225

PR #7766, opened 2022-04-29
Minor modifications to GSIM YoudEtAl2002 and Adding new GSIM Zhang and Zhao. (2005)

Prajakta-Jadhav-25:
Changed the sigma values as reported in "Franke, K.W., and Kramer, S.L. 2014. Procedure for the Empirical Evaluation of Lateral Spread Displacement Hazard Curves. Journal of Geotechnical and Geoenvironmental Engineering, 140(1): 110–120. doi:10.1061/(asce)gt.1943-5606.0000969"

Added a comment about adjustment suggested to handle deep earthquake sources


raoanirudh:
Hi @Prajakta-Jadhav-25, just a note that the IMT called PGD has been previously defined as Peak Ground Displacement measured in cm (https://github.com/gem/oq-engine/blob/engine-3.14/openquake/hazardlib/imt.py#L114-L118), which is used in the GSIM [Campbell and Bozorgnia (2008)](https://github.com/gem/oq-engine/blob/engine-3.14/openquake/hazardlib/gsim/campbell_bozorgnia_2008.py), whereas YoudEtAl2002 and Zhang_Zhao2005 estimate Permanent Ground Deformation from lateral spread measured in m.

A few geotech-related IMTs (see https://github.com/gem/oq-engine/blob/engine-3.14/openquake/hazardlib/imt.py#L239-L271) were introduced in #3861 and later modified in #6417 — would any of these be more appropriate for YoudEtAl2002 and Zhang_Zhao2005? @mmpagani and @cossatot, please chime in if you have any suggestions to avoid potential confusion between two different entities that are both abbreviated as PGD.

Prajakta-Jadhav-25:
Hi @raoanirudh, Thanks. I consider this quite an important comment/ suggestion. As you rightly pointed out, the 'PGD' term mentioned presently in the codes, Youd et al. (2002) & Zhang and Zhao (2005), is Peak ground displacement (cm), which is different than the permanent ground deformation due to lateral spreading (m). 
In the current list of imts for secondary perils, I wonder if the most applicable imt for now is Disp(), rather than PGDMax() and PGDGeomMean() as these consider vertical settlement also along with lateral spread. While developing the codes, I tried using PGDfLatSpread as the IMT, which was in one of the previous versions, but discontinued using it due to some errors. I am not sure if we need a separate IMT dedicated just for lateral spread displacements.  
Any suggestion from @mmpagani and @cossatot in this regard would be helpful to fix this issue.

mmpagani:
Hi, ideally it would be better having one acronym for each IMT with an 
associated (standard) unit of measure. Use of the same acronym in 
different contexts is unideal.

micheles:
LGTM as it is.

PR #7787, opened 2022-05-05
Simple performance improvement of Kuehn et al. 2020 model

SnowNooDLe:
# Simple performance improvement of Kuehn et al. 2020 model

### Kuehn 2020 proposed workflow on splitting recarrays

As proposed(#7785), we only split `recarrays` into sub-arrays if a user specifies a `sigma_mu_epsilon`

### Minor updates on the way of using `split_array`
In the current workflow, the given magnitudes must be in 2dp. 
To be slightly more robust, we modified a little bit and now, magnitudes can have more than 2dp.

### DIFF
--- a/openquake/hazardlib/contexts.py
+++ b/openquake/hazardlib/contexts.py
@@ -837,9 +837,7 @@ class ContextMaker(object):
             recarrays = [self.recarray(ctxs)]
         if any(hasattr(gsim, 'gmpe_table') for gsim in self.gsims):
             assert len(recarrays) == 1, len(recarrays)
-            recarrays = split_array(
-                recarrays[0], U32(numpy.round(recarrays[0].mag * 100).astype(int))
-            )
+            recarrays = split_array(recarrays[0], U32(recarrays[0].mag*100))
         self.adj = {gsim: [] for gsim in self.gsims}  # NSHM2014 adjustments
         for g, gsim in enumerate(self.gsims):
             compute = gsim.__class__.compute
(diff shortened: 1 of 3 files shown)


openquake/hazardlib/contexts.py
```diff
@@ -837,7 +837,9 @@ def get_mean_stds(self, ctxs):
             recarrays = [self.recarray(ctxs)]
         if any(hasattr(gsim, 'gmpe_table') for gsim in self.gsims):
             assert len(recarrays) == 1, len(recarrays)
-            recarrays = split_array(recarrays[0], U32(recarrays[0].mag*100))
+            recarrays = split_array(
+                recarrays[0], numpy.round(recarrays[0].mag * 100).astype(int)
+            )
```
micheles, line 842: Here it really must be an uint32 otherwise numba will complain. Please do a `pip install numba` and check.

SnowNooDLe: Apparently, the `venv` we set, already had `numba` and it still worked.

test

Should I still cast `U32()` on the second argument?

micheles: I would keep the U32 as it was before, without round, to make sure that the behavior is the same as before.

micheles: You are keeping the round here! I mean going back to `U32(recarrays[0].mag*100)` ;-)

SnowNooDLe: Ah! My bad haha
But in that case, users must use magnitudes in max 2dp, will these be okay?

micheles: It is already using 2 decimal points and was approved by @mmpagani

SnowNooDLe: Gotcha, they are back online.

micheles (approved):

PR #7837, opened 2022-05-19
Implemented infrastructure risk

micheles:
And added a test for scenario_damage.


raoanirudh (commented):
The module is affected by a serious bug in pandas v1.1: https://github.com/pandas-dev/pandas/issues/34603, which has been fixed in pandas ≥ 1.2. The pandas version used by the engine should be upgraded if possible. Other than that, LGTM.

Documentation and formal specs to be added.

PR #7883, opened 2022-06-03
RST Format Manual

pslh:
OpenQuake manual in RST format.
Converted from LaTeX sources via, Paddoc, custom automated scripts and then hand edited.
See how it looks here: http://oqmanual.gem.lan/


pslh:
I am not super-happy with the "<no-title>" HTML title for index.html, but I have not found a way to set this without changing the document structure.

micheles (approved):

PR #7899, opened 2022-06-09
Bahrampouri Arias Intensity GMPE

Shreyasvi91:
The existing GMPE in the master is for a generic case wherein the region-specific coefficients such as Cm and Ck have been set to 0. In this new version of the GMPE, these region-specific coefficients, and distance to volcanic arc are calculated based on the location of the source and the site.


micheles:
You forgot the documentation:
```python

____________________ DocumentationTestCase.test_documented _____________________
hazardlib/tests/gsim/utils_test.py:85: in test_documented
    raise InvalidFile('%s: %s is not documented' %
E   openquake.baselib.InvalidFile: /home/runner/work/oq-engine/oq-engine/doc/sphinx/openquake.hazardlib.gsim.rst: boraetal2019_FAS.py is not documented
```

openquake/hazardlib/gsim/bahrampouri_2021.py
```diff
@@ -74,26 +91,79 @@ def _get_site_term(ctx, C):
...
-
 def _get_stddevs(C):
+
     sig = C['sig']
     tau = C['tau']
     phi = np.sqrt(C['phi_ss']**2 + C['phi_s2s']**2)
+    
     return sig, tau, phi
 
+def _check_cm_ck(kyushu, honshu, ctx):
```
mmpagani, line 102: Is the correction for the volcanic front mandatory? Otherwise, we should have an option at the instantiation level to disable it.

mmpagani (commented):
Thanks for this. Some notes:
- You have a number of binary files that should be removed from the PR.
- You should check that your files comply with pep8 standard. You can find online pep8 checkers

openquake/hazardlib/gsim/bahrampouri_2021.py
Shreyasvi91: The correction for the volcanic front is not mandatory. In the absence of the functions `_check_cm_ck` and `_compute_volcanic_distances`, the path term  Screenshot 2022-06-10 at 3 39 40 PM 
reduces to 
Screenshot 2022-06-10 at 3 30 23 PM

There are two applicability scenarios:
1. sites outside of Japan - The current implementation calculates cm=ck=rrup_b=rrup_f to be equal to 0 and the path term collapses to the same form as above. 
2. sites within Japan - the present implementation holds good. 
Hence, instantiation may not be necessary. Please let me know if I was not clear or didn't understand your input.

mmpagani: Thanks. Wrt to point 1, is this case covered by the tests you created?

Shreyasvi91: yes. The tests covers case 1.

mmpagani: thank you. LGTM

micheles (approved):

micheles:
This GMPE will be exceedingly slow since there is a geospatial operation for each site effectively defeating the vectorization by site. But I cannot do anything about that :-(

Issue #8169, opened 2022-10-19
Abrahamson et al 2014 Japan module Vs30


cbworden:
Hi Michele,
The implementation of ASK14 here:
https://github.com/gem/oq-engine/blob/7a2bf79209988a47a19b0f0e2aac219952fc5691/openquake/hazardlib/gsim/abrahamson_2014.py#L210
fixes the minimum Vs30 at 150. There are two problems with this:
1. The official J-SHIS Vs30 model of Japan has values that go down to 125 m/s. (Please note that this is the official Japanese Vs30 model of the country, and we *must* respect it.) Thus, using the official Japanese Vs30 model of Japan causes errors when applying ASK14 Japan.
2. The Matlab code that Abrahamson et. al developed for their model (as supplied by Ronnie Kamai) includes a function that **extrapolates** the coefficients beyond the fixed range of the table given in the paper.

The extrapolate functionality is relatively new in the Scipy interp1d() function, so I imagine that the original developers of the OpenQuake ASK14 module did not have that available to them. Here we can fix that gap.

I would strongly suggest changing:
```
        f3 = interpolate.interp1d(
            [150, 250, 350, 450, 600, 850, 1150, 2000],
            [C['a36'], C['a37'], C['a38'], C['a39'], C['a40'], C['a41'],
             C['a42'], C['a42']],
            kind='linear')
```
to
```
        f3 = interpolate.interp1d(
            [150, 250, 350, 450, 600, 850, 1150, 2000],
            [C['a36'], C['a37'], C['a38'], C['a39'], C['a40'], C['a41'],
             C['a42'], C['a42']],
            kind='linear', fill_value='extrapolate')
```
where the only change is the addition of `fill_value='extrapolate'` to the argument list. The version of SciPy required is 0.17.0.

mmpagani:
Hi Bruce, thanks for this note. I am fine with your suggestion.

PR #8173, opened 2022-10-20
Adding extrapolate

mmpagani:
See https://github.com/gem/oq-engine/issues/8169


micheles:
Already done in https://github.com/gem/oq-engine/pull/8171.

PR #8248, opened 2022-11-17
Obtain a xml with the original namespaces after converting nrml -> gpkg -> nrml

ptormene:
Also add to the Node class the possibility to print the Node object with or without stripping the namespaces from tags, or displaying the namespace aliases instead of the full strings.


ptormene:
I suspect some of the tags found in the original `mixed.xml` file should have the `gml` namespace, like other similar tags. For instance, `Edge` and `Polygon`

openquake/hazardlib/nrml.py
```diff
@@ -379,6 +379,10 @@ def read(source, stop=None):
     xmlns = nrml.tag.split('}')[0][1:]
     nrml['xmlns'] = xmlns
     nrml['xmlns:gml'] = GML_NAMESPACE
+    nrml.nsmap = {
+        xmlns: '',
+        GML_NAMESPACE: 'gml'
+    }
```
ptormene, line 385: We could find a way to avoid this duplication

openquake/hazardlib/sourceconverter.py
```diff
@@ -1337,7 +1337,7 @@ def convert_sliplist(self, node):
         except AttributeError:
             lst = []
         else:
-            lst = [node_to_dict(n)['slip'] for n in slip_list.nodes]
+            lst = [list(node_to_dict(n).values())[0] for n in slip_list.nodes]
```
ptormene, line 1340: Perhaps it was clearer without this change. I made this in order to be more robust with respect to the retrieval of the tag by name, because I was trying to consider the fully qualified tag instead of the one without the namespace.

openquake/hazardlib/nrml_to.py
```diff
@@ -158,15 +158,15 @@ def convert_to(fmt, fnames, chatty=False, *, outdir='.', geometry=''):
...
+                appendrow(row, srcs, chatty, sections, s2i)
         else:  # nrml/0.5
             for srcgroup in root.sourceModel:
                 trt = srcgroup['tectonicRegion']
                 for srcnode in srcgroup:
                     srcnode['tectonicRegion'] = trt
-                    appendrow(converter.convert_node(srcnode),
-                              srcs, chatty, sections, s2i)
+                    row = converter.convert_node(srcnode)
+                    appendrow(row, srcs, chatty, sections, s2i)
```
ptormene, line 169: This change and the one above were just for debugging purposes, because this way it becomes a little easier to step into the desired functions via pdb

openquake/hazardlib/sourceconverter.py
ptormene: I've reverted this change

micheles (approved):

PR #8256, opened 2022-11-22
Added experimental command `oq submit`

micheles:
For internal usage.


micheles:
Let's close this now that we have SLURM experience.

PR #8276, opened 2022-11-30
Jenkins timeout

nastasi-oq:
Recover jenkins job.
Test are passed except missing xml logs that are produced in the devtest step that here is missing: http://jenkins.gem.lan/job/zdevel_oq-engine/44/console


openquake/server/tests/views_test.py
```diff
@@ -91,7 +91,8 @@ def get_text(cls, path, **data):
     @classmethod
     def wait(cls):
         # wait until all calculations stop
-        for i in range(30):  # 30 seconds of timeout
+        # for i in range(30):  # 30 seconds of timeout
+        for i in range(300):  # 300 seconds of timeout (FIXME: just for test)
```
ptormene, line 95: Perhaps we could try with a lower timeout?

packager.sh
```diff
@@ -927,7 +933,8 @@ sudo systemctl status openquake-dbserver
             echo \"The 'oq webui' command is broken: it reports\n\t\$webui_fail\ninstead of\n\t\$webui_fail_msg\"
             exit 1
         fi
-        sudo -u openquake oq webui migrate"
+        # FIXME: this directory must be discussed
```
ptormene, line 936: Should this be discussed with Michele?

ptormene (approved):

PR #8296, opened 2022-12-14
Optimized `general.group_array`

micheles:
Using numpy is faster.


micheles:
Abandoning, it is too much of an effort :-(

PR #8335, opened 2023-01-14
Stochastic Event Sets with mutex sources

mmpagani:
Part of https://github.com/gem/oq-engine/issues/8631.


micheles:
Superseded by https://github.com/gem/oq-engine/pull/8671.

PR #8533, opened 2023-03-16
Long names update

Shreyasvi91:
This PR addresses test case 76 in classical tests. The GMPEs have been revised as per the oq engine version 3.17. The tests are passing with these new updates.


openquake/hazardlib/gsim/can20/can_shm6_active_crust.py
```diff
@@ -87,6 +87,7 @@ def _get_site_scaling_ba14(kind, region, C, pga_rock, sites, period, rjb):
 
     fnl = BA14._get_nonlinear_site_term(C, sites.vs30, pga_rock)
     fbd = BA14._get_basin_depth_term(region, C, sites, period)  # returns 0
+    fbd = 0.0
```
kejohnso, line 90: if line 89 returns 0.0, can we simply remove these two lines and not add the term in line 92? 
(or maybe we want to keep evidence of it as documentation for these gmpes?)

mmpagani: I did it this way for the reason you were mentioning

micheles (approved):

PR #8605, opened 2023-04-04
Log only when upgrading the DB

micheles:
Before it was logging a message even when the database was already updated.


pslh (approved):
LGTM

PR #8734, opened 2023-05-11
Removing the area type concept

micheles:
Part of https://github.com/gem/oq-engine/issues/8724


micheles:
This is actually used, just not tested.

PR #8756, opened 2023-05-16
Test action rpm with logs of webui

vot4anto:


vot4anto:
Update action with test of rpm created packages

Issue #8702, opened 2023-05-04
Give proper support when pressing the `?` button on top of the WebUI


ptormene:
Currently it links to the oq-engine repository.
It should probably open a page with links to:

- the engine user manual
- the WebUI user manual
- the API documentation

PR #8765, opened 2023-05-17
Replace url for menu help button

ptormene:
Fixes https://github.com/gem/oq-engine/issues/8702

Now it points to https://www.globalquakemodel.org/openquake instead of https://github.com/gem/oq-engine


hascar (approved):
LGTM

PR #8772, opened 2023-05-18
Improve advanced manual about aggregation by multiple keys for reinsurance

ptormene:
Fixes https://github.com/gem/oq-engine/issues/8750

I managed to build both the html and the pdf of the advanced manual and I checked that the cross-reference works as expected in both cases.
Screenshot from 2023-05-18 10-15-34


CatalinaYepes (approved):
LGTM

PR #8778, opened 2023-05-19
Fixes Errors in ESHM20 GMM Aliases and Minor Bug in Lanzano et al. (2019)

g-weatherill:
This PR fixes two minor bugs in European GMMs

1) For the ESHM20 alias functions one of the alias coefficients was incorrect, while the call to the `theta6_adjustment` had been incorrectly entered as `theta_6_adjustment`. Both are now fixed with changes propagated to tests

2) A bug is present in the site scaling term of the Lanzano et al. (2019) GMMs here: https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/gsim/lanzano_2019.py#L71. When Vs30 contains values greater than 1500 then this line gives a mismatched array error. This case is not present in the model's tests. This is now fixed with the use of the cleaner numpy.clip function.


micheles (approved):

openquake/hazardlib/gsim/lanzano_2019.py
```diff
@@ -67,9 +67,7 @@ def _site_amplification(ctx, C):
     it is given by FS = klog10(V0/800) , where V0 = Vs30 when Vs30 <= 1500
     and V0=1500 otherwise
     """
-    v0 = np.ones_like(ctx.vs30) * 1500.
-    v0[ctx.vs30 < 1500] = ctx.vs30
-    return C['k'] * np.log10(v0/800)
+    return C['k'] * np.log10(np.clip(ctx.vs30, -np.inf, 1500.0) / 800.0)
```
mmpagani, line 70: Thanks. Indeed we also noticed this problem. If possible, it would be worth having a test for this (even just one line in a verification table).

mmpagani (approved):
Thanks, Graeme for this PR. Apart from a minor note, it LGTM.

Issue #8788, opened 2023-05-23
Handling multiple human loss types (fatalities, injuries, homeless)


nicolepaul:
We need to cleanly separate between two different human loss types: casualties (i.e., injuries and fatalities) and the population rendered homeless. Both use exposure fields related to the building occupants, but casualties depend on the time of the earthquake event whereas homeless does not.

We also need to add tests to confirm these calculations are happening as expected.

# Casualties: Fatalities, injuries, and time-based calculations

As of https://github.com/gem/oq-engine/pull/8786, we now have an OQ field `occupants_avg`, which represents the time-averaged number of occupants with the assets. This is in addition to previous occupants-related OQ fields: `night`, `day`, and `transit`, which refer to the number of occupants within a given `time_period`. 

I am assuming that we will continue to use the `occupants` loss_type for casualty calculations. Thus, in risk calculations, they use the `occupants_vulnerability_file`.

## Scenario calculations

In scenario calculations, we may specify a `time_period` in the **job.ini**. This tells us which OQ field to use for the injury and fatality calculations. For example:

**job.ini**
```time_period = day```

**exposure.xml**
```
  <exposureFields>
    <field oq="day" input="OCCUPANTS_PER_ASSET_DAY" />
    <field oq="night" input="OCCUPANTS_PER_ASSET_NIGHT" />
    <field oq="transit" input="OCCUPANTS_PER_ASSET_TRANSIT" />
  </exposureFields>
```

In this case, the `day` occupants (OCCUPANTS_PER_ASSET_DAY) should be used for the injury and fatality calculations. 

## Event-based calculations

In event-based calculations, we don't use the `time_period` parameter. Therefore, we want to use the time-averaged occupants. Using the new `occupants_avg` OQ field, we can specify:

**exposure.xml**
```
  <exposureFields>
    <field oq="day" input="OCCUPANTS_PER_ASSET_DAY" />
    <field oq="night" input="OCCUPANTS_PER_ASSET_NIGHT" />
    <field oq="transit" input="OCCUPANTS_PER_ASSET_TRANSIT" />
    <field oq="occupants_avg" input="OCCUPANTS_PER_ASSET_AVERAGE" />
  </exposureFields>
```

In the injury and fatality calculations, only `occupants_avg` (OCCUPANTS_PER_ASSET_AVERAGE) should be used.

# Population rendered homeless

[...]

PR #8832, opened 2023-06-05
Introduced new loss type `residents`

micheles:
Part of https://github.com/gem/oq-engine/issues/8788.


nicolepaul (commented):
LGTM; tested with some regions in the GRM and the results seem reasonable.

PR #8862, opened 2023-06-21
Modification to Kuehn sigma_mu_adjustment calculation

claudio525:
PR contains:
1) Bug fix for https://github.com/gem/oq-engine/issues/8856, which is already addressed by https://github.com/gem/oq-engine/pull/8861
2) Removes requirement to only have a single unique magnitude when using `sigma_mu_epsilon` parameter (line 716)
3) Replace multi-step interpolation with single 2D/3D space interpolation to support fast calculation of adjustment factor

To ensure correct adjustment values outside of the model domain, the model domain is manually extended to prevent having to manually deal with out of domain values, as RegularGridInterpolator (https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.RegularGridInterpolator.html#scipy.interpolate.RegularGridInterpolator) does not support specifying fill values for each dimension.

### DIFF
--- a/.github/workflows/macos_intel_install.yml
+++ b/.github/workflows/macos_intel_install.yml
@@ -60,7 +60,9 @@ jobs:
           pip3 install pytest
           oq --version
           oq dbserver start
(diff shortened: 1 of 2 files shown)


mmpagani (approved):
LGTM

PR #8885, opened 2023-07-08
Adding disappeared test for mrd calculation

mmpagani:


micheles:
Thanks @mmpagani , LGTM.

PR #8904, opened 2023-07-14
Backported critical fix https://github.com/gem/oq-engine/pull/8902

micheles:


micheles:
@raoanirudh checked this and it works :-)

PR #8914, opened 2023-07-21
Use --trusted-host while installing the engine with install.py

ptormene:
Fixes https://github.com/gem/oq-engine/issues/8908


vot4anto (approved):
LGTM

PR #8978, opened 2023-09-02
Fixed check on reaggregate

micheles:
This is not solving the issue of @martinacaruso yet; it will be followed by another PR.


raoanirudh (commented):
Still not working I believe. Just checked with EventBasedRisk/job_eb.ini and `oq reaggregate -1 NAME_1`. The aggcurves-stats-NAME_1 and aggrisk-stats-NAME_1 outputs are incomplete.

PR #9014, opened 2023-09-15
Upgrade Pandas to v2.0.3

ptormene:
Starting from https://github.com/gem/oq-engine/pull/9013

Addressing part of https://github.com/gem/oq-engine/issues/8810

NOTE: we are migrating to pandas 2.0.3 and not to pandas 2.1.0 for now, because otherwise  it would imply dropping support for python 3.8, that we want to keep.


vot4anto:
The tests are not green: 
https://github.com/gem/oq-engine/actions/runs/6233279084/job/16918152259
@ptormene it is better to check this error

PR #9163, opened 2023-10-30
Simplified aelo outputs page

ptormene:
As discussed in the last project meeting, I've differentiated a standard ("advanced") outputs page and a simplified one that is specific for AELO. From each of these pages there is a button pointing to the alternative page.
The simplified page shows a limited set of outputs, with the two ASCE outputs highlighted in bold and displaying the key-value pairs inline. To further simplify the page, I removed the "Type" column, that provided internal information not important to the final user.
When a AELO calculation is completed, the email sent to the user points to the new simplified page.


ptormene:
Currently the simplified page looks like this:
image

PR #9175, opened 2023-11-02
plots for low hazard

ManuelaVillani:
adding the governing plot also for low hazard sites


ptormene:
Even if it is not specifically related to this PR, we should fix the following:

- [ ] This warning: `/home/ptormene/GIT/oq-engine/openquake/calculators/postproc/compute_rtgm.py:424: RuntimeWarning: divide by zero encountered in log
  f = interpolate.interp1d(np.log(imls), np.log(afe))`
- [ ] The docstrings marked with FIXME

Besides these aspects, I've verified that the proposed changes work as expected, therefore LGTM

PR #9177, opened 2023-11-02
Give a clearer error message when attempting to create a new normal user assigning a username that is already taken

ptormene:


pslh (approved):
LGTM

PR #9180, opened 2023-11-03
Doc23 switcher capability to the new documentation

nastasi-oq:
Outcome is already visible here:
https://docs.openquake.org/.dev/.oq-engine-new/doc23_switcher/manual/


CatalinaYepes (approved):
LGTM

PR #9186, opened 2023-11-07
Customize messages in password reset workflows

ptormene:
The `createnormaluser` Django command that we created automatically sends an email to the new user asking them to reset their password. This PR differentiates what the user receives in that case with respect to what happens if the user actively requests a password reset.
I created different templates to be rendered when the engine is set to be running in aelo mode or in restricted mode.
Currently those templates are very similar to the default ones provided by Django, but there's space for further customization.
I'd ask @pslh to check if the current text looks good or if we may improve it a bit.

NOTE: I have checked what happens if a user that has more than one account associated to the same email address requests a password reset: one email for each username associated to that address is sent, and each email indicates the corresponding username inside, avoiding confusion. Therefore if you click on the link in one of the emails you received you activate the password reset process for the corresponding username and not for the others (that is in fact the correct thing that is expected to happen).


openquake/server/management/commands/createnormaluser.py
```diff
@@ -45,8 +45,19 @@ def handle(self, *args, **kwargs):
...
+        if settings.APPLICATION_MODE.upper() == 'AELO':
+            password_reset_subject = (
+                'registration/normal_user_creation_email_subject_aelo.txt')
+            email_template_name = (
+                'registration/normal_user_creation_email_aelo.txt')
+        else:
+            password_reset_subject = (
+                'registration/normal_user_creation_email_subject.txt')
+            email_template_name = (
+                'registration/normal_user_creation_email.txt')
```
ptormene, line 57: I fixed the extensions to txt

openquake/server/templates/registration/password_reset_email_aelo.txt
```diff
@@ -0,0 +1,16 @@
+{% load i18n %}{% autoescape off %}
+{% blocktranslate %}
+You are receiving this email because you requested a password reset for your user account at {{ site_name }}.
+(If you did not request this password reset, you can safely ignore this message and your password will not be changed)
```
ptormene, line 4: I added this note

pslh (approved):
LGTM

PR #9190, opened 2023-11-07
Use Fiona wheel that include also GDAL

vot4anto:
PyPI wheels of Fiona  include GDAL, PROJ  and GEOS .


micheles (approved):

openquake/sep/utils.py
```diff
@@ -6,10 +6,6 @@
 import numpy as np
 import pandas as pd
 import pyproj as pj
-try:
```
vot4anto, line 9: @micheles also on this files we need to remove import of gdal ?

micheles: NO!

antonioettorre:
Use new wheel from oq-wheel repos and not the original fiona wheel

PR #9199, opened 2023-11-08
Add USE_REVERSE_PROXY variable to local settings and set request server port depending from it

ptormene:
TODO:

- [x] add description of USE_REVERSE_PROXY into the documentation for production installation
- [x] add USE_REVERSE_PROXY to settings.py with a default value False if not available in the local settings


openquake/server/local_settings.py.aelo
```diff
@@ -8,8 +8,10 @@ APPLICATION_MODE = 'aelo'
 # WEBUI config uncomment and set properly if needed
 # WEBUIURL = 'http://localhost:8800/'
 
+# TODO: ask Antonio
```
ptormene, line 11: @vot4anto may you check these USE_X_FORWARDED_HOST and USE_X_FORWARDED_PORT, please?

pslh: We think that these variables might/should help when used in combination with NGINX, but when tested things did not behave as expected.

pslh (approved):
LGTM

openquake/server/local_settings.py.aelo
vot4anto: This variable is not needed since we already setup the log in right format: https://github.com/gem/oq-engine/pull/8684.

ptormene:
We may further investigate USE_X_FORWARDED_HOST and USE_X_FORWARDED_PORT in a separate PR

PR #9240, opened 2023-11-22
NZ NSHM 2022 modification of Parker2020

chrisdicaprio:


mmpagani (approved):
If you want to make this module more compact some of the functions could be imported from the original module. At this stage, I am happy with this version. If necessary we can cleanup the code afterwards.

PR #9272, opened 2023-12-06
AELO: better handle cases with low or zero hazard

ptormene:
1. Added a warning both in the email and in the simplified outputs page when using "preliminary" models ('CEA', 'CHN', 'NEA')
2. Improved logging and added warning to the simplified outputs page in case of zero or low hazard
3. If ASCE tables can not be computed, they are anyway stored and displayed with zeros, n.a. and 'Low' values
4. The warning was moved below the outputs title and it was prefixed with "WARNING: "
5. Numbers are formatted to always display 2 decimal places also when the last digit is zero


kejohnso:
Ciao Paolo, all LGTM!!

kejohnso:
@ptormene on second thought, I'm wondering: would it be possible to include the warning messages also on the "advanced outputs" page? @ManuelaVillani

PR #9277, opened 2023-12-12
AELO: when presenting ASCE tables, format numbers to display 2 significant figures

ptormene:
As suggested by Zach


kejohnso (approved):
LGTM

PR #9365, opened 2024-01-18
Add pinax.announcements package to manage temporary broadcast messages via the admin interface

ptormene:
This is a proposal to obtain what we discussed during last scrum.
It requires a migration.
The configuration via the admin interface is intuitive:
Screenshot from 2024-01-18 17-31-03
Then the result is something like:
Screenshot from 2024-01-18 17-31-21
Obviously this is just a draft and the style can be further improved.

NOTE: the most painful aspect of this addition is that it would require the dependency from `pinax-announcements==4.0.0`

### DIFF
--- a/openquake/server/settings.py
+++ b/openquake/server/settings.py
@@ -33,7 +33,8 @@ except ImportError:
 
 TEST = 'test' in sys.argv
 
-INSTALLED_APPS = ('openquake.server.db',)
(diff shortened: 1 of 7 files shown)


ptormene:
We chose to implement this feature without introducing new external dependencies: https://github.com/gem/oq-engine/pull/9372

PR #9372, opened 2024-01-22
Add Django model `Announcement` and read it from context processor to display announcements in all templates

ptormene:
Alternative solution with respect to https://github.com/gem/oq-engine/pull/9365/files
This way we don't need to depend from any external package. For our purpose, the functionality is very similar: you can add announcements through the admin interface and all the existing announcements are displayed on top of every template. If you delete any instance of Announcement and reload any template, the corresponding item will not be visible anymore. The announcements box completely disappears if there are no announcements to display.

`pinax.announcements` offers some other functionalities that we probably don't need, e.g.: you can specify a starting and ending time for displaying the announcement; you can decide if you want to show an announcement to everybody or to members only, etc.

NOTE: I added the model `Announcement` in the only registered app, i.e. `openquake.server.db`. I guess this is reasonable. Otherwise we can move it to a separate app.

The following is an example of announcement:
image

### DIFF
--- a/openquake/server/static/css/engine.css
+++ b/openquake/server/static/css/engine.css
@@ -277,6 +277,8 @@ tr.asce_output {
 .outputs-general-btns-right {
     width: calc(100% / 3);
     float: left;
+    min-height: 100px;
+    display: block;
 }
 
 .outputs-general-btns-left {


pslh (approved):
I would like to see how the announcement looks "in real life" following the changes but in general LGTM.

ptormene:
With @hascar we also improved the styling of buttons and fixed the background of GEM icon. The result is as in the attached screenshot:

image

PR #9483, opened 2024-02-23
Fix warnings aelo

kejohnso:
- Fixes aelo warnings shown in banner above main outputs on the webui
- adds tests for each type of low hazard warning


kejohnso:
- added a new warning for when ASCE values are fully computed but the results are super low
- added tests on correct figure production

micheles (approved):

PR #9501, opened 2024-03-01
Add ARISTOTLE mode

ptormene:
image

When APPLICATION_MODE is set to ARISTOTLE, the webui displays a form to retrieve rupture information given a shakemap id. That form triggers an ajax call that populates a second form with pre-populated (editable) variables, plus dip, strike and a selector for trt that is populated based on the mosaic model corresponding to the lon and lat of the rupture that was retrieved.
The submit button currently only collects the user-edited values and the selected trt, printing those values in the console.
What comes after that is still to be done.

NOTE: site_model.hdf5 needs to be put into mosaic_dir in order to make it possible to retrieve the set of trts.

### DIFF
--- a/openquake/server/views.py
+++ b/openquake/server/views.py
@@ -44,7 +44,7 @@ import numpy
 
 from openquake.baselib import hdf5, config
 from openquake.baselib.general import groupby, gettemp, zipfiles, mp
-from openquake.hazardlib import nrml, gsim, valid
+from openquake.hazardlib import nrml, gsim, valid, geo
 from openquake.hazardlib.shakemap.parsers import get_rupture_dict
 from openquake.commonlib import readinput, oqvalidation, logs, datastore, dbapi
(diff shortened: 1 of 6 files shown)


micheles:
This is good. The defaults for strike and dip are strike=0., dip=90.

openquake/server/views.py
```diff
@@ -598,6 +611,141 @@ def aelo_callback(
...
+@csrf_exempt
+@cross_domain_ajax
+@require_http_methods(['POST'])
+def aristotle_run(request):
+    """
+    Run an ARISTOTLE calculation.
+
+    :param request:
+        a `django.http.HttpRequest` object containing lon, lat, dep, mag, rake
+    """
```
micheles, line 685: Probably you want to put the validation part in a helper function, but it can be done later on

ptormene:
image
I added the default values for dip and strike

PR #9538, opened 2024-03-20
Add to the webui outputs page buttons to display avg_gmf plots for all the available imts

ptormene:
TODO:

- [x] check if the retrieval of calc_id is done in the correct way
- [x] check if the creation of the avg_gmf plots is done in the right place or, for instance, it should be moved into the postprocessing stage
- [x] better refactor the make_figure


openquake/calculators/event_based.py
```diff
@@ -50,6 +51,11 @@
 from openquake.calculators.getters import get_rupture_getters, sig_eps_dt
 from openquake.calculators.classical import ClassicalCalculator
 from openquake.engine import engine
+from openquake.commands.plot import plot_single_avg_gmf
+try:
+    from PIL import Image
```
ptormene, line 56: We can import it normally (here and wherever it was imported this way)

openquake/calculators/event_based.py
```diff
@@ -739,6 +745,15 @@ def save_avg_gmf(self):
                 gmf_df, self.weights, self.oqparam.min_iml).items():
             avg_gmf[:, sid] = avgstd
         self.datastore['avg_gmf'] = avg_gmf
+        imts = list(self.oqparam.hazard_imtls)
+        calc_id = -1  # FIXME: is this the right way to get it?
```
ptormene, line 749: self.datastore.calc_id

openquake/calculators/event_based.py
ptormene: Done

openquake/calculators/event_based.py
ptormene: Done

micheles (approved):

PR #9565, opened 2024-04-04
Tools cookie consent

hascar:
Added code in urls, settings and local_setting.tools.


openquake/server/settings.py
```diff
@@ -226,6 +226,17 @@
 if not os.environ.get('OQ_APPLICATION_MODE'):
     os.environ['OQ_APPLICATION_MODE'] = APPLICATION_MODE
 
+if APPLICATION_MODE.upper() in ('TOOLS_ONLY',):
```
ptormene, line 229: Please be aware that yesterday we made the APPLICATION_MODE case-sensitive (always capitalized) and we added a check to make sure that the specified application mode is included in the list of available ones.

ptormene (approved):
If tests are green and everything works as expected, LGTM

openquake/server/settings.py
hascar: Fixed with this commit https://github.com/gem/oq-engine/pull/9565/commits/c62fb9812ac7fcfb7735361ec173b191b0efce85

openquake/server/fixtures/0001_cookie_consent.json
```diff
@@ -0,0 +1,64 @@
+[
```
nastasi-oq, line 1: Should be better if this file will be moved below oq-platfrom-standalone repository because is there that is managed.

nastasi-oq (approved):
LGTM

Issue #9579, opened 2024-04-08
AAL Statistics output for classical_risk/case_2 not exportable


raoanirudh:
Trying to export the Average Asset Losses Statistics output for [qa_tests_data/classical_risk/case_2](https://github.com/gem/oq-engine/tree/master/openquake/qa_tests_data/classical_risk/case_2) results in the following error:

```sh
Traceback (most recent call last):
  File "oq-engine/openquake/engine/export/core.py", line 64, in export_from_db
    exported = export(output_key, dstore)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "oq-engine/openquake/baselib/general.py", line 609, in __call__
    return self[key](obj, *args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "oq-engine/openquake/calculators/export/risk.py", line 206, in export_avg_losses
    name, value, rlzs_or_stats = _get_data(
                                 ^^^^^^^^^^
  File "oq-engine/openquake/calculators/export/risk.py", line 177, in _get_data
    weights = dstore['weights'][()]
              ~~~~~~^^^^^^^^^^^
  File "oq-engine/openquake/commonlib/datastore.py", line 517, in __getitem__
    raise KeyError('No %r found in %s' % (key, self))
KeyError: "No 'weights' found in <DataStore oqdata/calc_1.hdf5 open>"
```

Reported on the mailing list: https://groups.google.com/g/openquake-users/c/3xVJGxqfZlI

PR #9584, opened 2024-04-10
Fixed `avg_losses-stats` exporter in the case of a single realization

micheles:
Closes https://github.com/gem/oq-engine/issues/9579.


raoanirudh (commented):
LGTM

PR #9617, opened 2024-04-23
Aristotle: improve layout of the outputs page

ptormene:
- add 'Type' column (as in the standard outputs page)
- add horizontal scroller to the losses table to avoid exceeding the right limit of the container space
- fix positioning and width of buttons
image


hascar (approved):
LGTM

PR #9662, opened 2024-05-13
Changed the UHS exporter in AELO mode

micheles:
Following a suggestion by @ManuelaVillani which is more Excel-friendly.


pslh (approved):
LGTM

PR #9690, opened 2024-05-22
Installation instruction for python3.11and oq-engine on macos

vot4anto:


micheles:
There is nobody linking the new page, so how will the user discover it? We need a link from the universal installer page to the macos.md page.

vot4anto:
The page is linked in the index of getting started

Il gio 23 mag 2024, 08:23 Michele Simionato ***@***.***> ha
scritto:

micheles (approved):

PR #9773, opened 2024-06-11
Use Django commands instead of webui alias for server installation

hascar:


doc/getting-started/installation-instructions/server.md
```diff
@@ -46,22 +46,22 @@ Upgrade the database to host users and sessions:
...
 
 ```console
 cd /opt/openquake/src/oq-engine/openquake/server
-sudo -u openquake oq webui collectstatic
+sudo -u openquake /opt/openquake/venv/bin/python3 manage.py collectstatic
 ```
-The `oq` commands must be run as the `openquake` user and the installation must be of kind `server` or `devel_server`.
-if, for any reason, the `oq` command isn't available in the path, you can use the following syntax:
+The commands must be run as the `openquake` user and the installation must be of kind `server` or `devel_server`.
+if, for any reason, the commands are not available in the path, you can use the following syntax:
```
ptormene, line 64: Please capitalize "If"

ptormene (approved):

PR #9837, opened 2024-07-11
Fix dev-tools installation

ptormene:
* Fix commenting out a jinja variable correctly (it was inside a commented section of javascript, but jinja interpreted it anyway, looking for an url that is missing when authentication is disabled
* Fix local_settings.py.tools template, that had a redundant redefinition of LOCKDOWN


hascar (approved):
LGTM

PR #9863, opened 2024-07-26
ARISTOTLE: improve format of losses table, plus other improvements

ptormene:
Losses table:

- better text alignment
- multi-line gsims
- human-readable values

Other:

- handle exceptions in case retrieving rupture data fails for any reason
- if a rupture file is selected, disable dip and strike and get their values from the parsed rupture (we still need to check the management of dip and strike when the usgs shakemap is used, but it can be done in a separate PR)

image

image

### DIFF
--- a/openquake/calculators/views.py
+++ b/openquake/calculators/views.py
@@ -1740,7 +1740,7 @@ def view_aggrisk(token, dstore):
     AVG = len(rlzs)
     arr = numpy.zeros(AVG + 1, dt)
     for r, rlz in enumerate(rlzs):
-        arr[r]['gsim'] = repr(repr(rlz.value[0]))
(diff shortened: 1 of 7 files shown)


openquake/calculators/views.py
```diff
@@ -1740,7 +1740,7 @@ def view_aggrisk(token, dstore):
     AVG = len(rlzs)
     arr = numpy.zeros(AVG + 1, dt)
     for r, rlz in enumerate(rlzs):
-        arr[r]['gsim'] = repr(repr(rlz.value[0]))
```
ptormene, line 1743: @micheles I hope I did not break anything by replacing the double `repr` with a single one. If that's a problem, I can modify the string afterwards while presenting the data.

raoanirudh (approved):

PR #9874, opened 2024-07-30
Improve documentation of server installation and fix a corner case in using temporary directory

ptormene:
For some reason, in the aristotle-beta machine, uploaded input files were saved into `/tmp` instead of using the directory specified in the configuration file, even though the TMPDIR environment variable was properly set (and visible via os.environ). With this change, the specified directory is used. Django still saves the original uploaded files to `/tmp`, but it automatically deletes them as soon as the request is completely processed.


hascar (approved):
LGTM

PR #9876, opened 2024-07-30
In server/views.py always use config.directory.custom_tmp if present, when creating temp files or directories

ptormene:
Continuation of https://github.com/gem/oq-engine/pull/9874


hascar (approved):
LGTM

PR #9890, opened 2024-08-06
fix: convert expectation of lognormal distribution to mean of normal

schmidni:
In the Shakemap XML, the values for the SA intensity measures are given in %g and the uncertainties in ln(%g).

In the oq code, the sampling of the shakemaps is done on a normal distribution (`truncnorm.rvs`) - therefore the intensity values need to be transformed to `mu` and `sigma` of a normal distribution.

Here I made an error when extending the sampling scripts, namely I dropped the difference between the location parameter and the expectation of a lognormal distribution
`E_lognormal[X] = e^(mu + 1/2 * sig^2)`

This was done before in the old version of the code [here](https://github.com/gem/oq-engine/blob/e335c2e82e27bd97a2dc97d0b0ecd32c383ab523/openquake/hazardlib/shakemap.py#L292) but has been forgotten in the new version.


micheles (approved):

raoanirudh:
Hi @schmidni, 

The [USGS `grid.xml` exporter](https://code.usgs.gov/ghsc/esi/shakemap-modules/-/blob/v1.1.19/src/shakemap_modules/coremods/gridxml.py?ref_type=tags#L125-145) takes the mean in natural-log space and writes `exp(mean)` (multiplied by 100 for acceleration) to `grid.xml`. Thus, the `grid.xml` values appear to be the median ground motion, not its arithmetic expectation.

Would you agree that, in this case, we should remove the `−σ²/2` correction introduced in #9890 and restore sampling as:

```python
numpy.exp(numpy.log(mu) + residual)
```

instead of the change introduced in this PR as:
```python
numpy.exp(numpy.log(mu) + residual - sigma**2 / 2)
```

In my understanding, the `−σ²/2` term would be needed only if the value supplied in the ShakeMap grid.xml were the arithmetic mean. If the supplied value in grid.xml is actually the median, then adding this correction term shifts the sampled median down from the actual median.

schmidni:
Me (and apparently also the original authors) assumed the value exported to grid.xml from shakemap was the mean, not the median....

raoanirudh:
Right, the correction factor seems to have been present (mistakenly) also in the original version of the code. I'll open a PR to remove this correction factor.

raoanirudh:
See https://github.com/gem/oq-engine/pull/11701.

Issue #9868, opened 2024-07-26
Add a warning to the WebUI login page about the possibility to be banned


ptormene:
Whenever the user fails a login attempt, the error message should state something like:
"WARNING: after some consecutive failed login attempts you may be banned and become unable to access the website for the next couple of hours"

PR #9892, opened 2024-08-07
WebUI: customize login form to include a warning about the possibility to be banned after failing multiple login attempts

ptormene:
Fixes https://github.com/gem/oq-engine/issues/9868

We don't know in advance the precise rules about the number of allowed consecutive login attempts and about the amount of time the user will be banned after failing too many attempts, because such rules are defined via Fail2Ban in each specific installation. Therefore, the warning is intentionally generic.

### DIFF
--- a/openquake/server/forms.py
+++ b/openquake/server/forms.py
@@ -22,7 +22,7 @@ from django.contrib.auth.forms import AuthenticationForm
 
 class CustomAuthenticationForm(AuthenticationForm):
     """
-    Customizing the invalid_login message to include a warning about the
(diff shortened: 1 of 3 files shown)


hascar (approved):
LGTM

ptormene:
It looked like a trivial change, but actually customizing the LoginView under the condition that LOCKDOWN == True seems to cause some malfunctioning in Django and I couldn't easily manage to find a workaround. Considering that anyway the warning could not be very specific and it would not be very helpful (it would not be visible after actually being banned), I am giving up for now.

PR #9906, opened 2024-08-17
Sep landslide

LanaTodorovic93:
Adding Nowicki Jessee et al. 2018 landslide model which is part of the USGS Ground Failure Product. It calculates the areal coverage accounting for modifications to the original model introduced by Allstadt et al. (2022). The model uses globally available layers as the inputs that correlate with landslide occurrence. The inputs are peak ground velocity, compound topographic index, topographic slope, lithology and landcover.

This PR adds:

- [x] add nowicki_jessee_2018 model including the modifications incorporated within USGS GF Product.
- [x] test
- [x] documentation (only the brief theoretical background on the model, which inputs are required and the corresponding reference)

Check whether the following is needed:
- [x] qa tests / demos
- [x] more work on the documentation


doc/underlying-science/secondary-perils.rst
```diff
@@ -495,9 +495,9 @@ Nowicki Jessee et al. (2018)
 A geospatial model used to predict probability of landsliding using globally available geospatial variables was proposed by 
 `Nowicki Jessee et al. (2018) <https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2017JF004494>`_. The level of shaking is 
 characterised by Peak Ground Velocity , :math:`PGV`. Slope steepness affects slope stability, and here, the topographic 
-slope, :math:`slope`, has been dervide from the median elevation value from the Global Multi-resolution Terrain Elevation Data 
-(`Danielson and Gesch, 2011 <https://pubs.usgs.gov/of/2011/1073/>`_). The model uses lithology, as a proxy for the strength of 
-the shaken material. The global lithology map is available in `Hartman and Moosdort, 2012 <https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2012GC004370>`_. 
+slope, :math:`slope`, has been dervide from the median elevation value from the 7.5 arc sec Global Multi-resolution Terrain
```
raoanirudh, line 498: *derived

doc/underlying-science/secondary-perils.rst
```diff
@@ -488,6 +488,50 @@ Finally, the coseismic displacements are estimated using the sliding block displ
...
+
+These probabilities are converted to areal percentages to unbias the precitions.
+.. math::
+
+	L_{P}(P) = e^(-7.592 + 5.237 \cdot P - 3.042 \cdot P^2 + 4.035 \cdot P^3) \quad (24)
+
+Furthermore, we introduced modifications by the USGS, capping the peak ground velocity at :math:`PGV = 211 \, \text{cm/s}`, 
+and compound topographic index at :math:`CTI = 19`. To exclude high probabilities of landsliding in nearly flat areas due to 
+the combination of other predictor variables, areas with slopes less than :math:`2^\circ` are excluded.  Zero probability is 
+heuristically assigned if :math:`PGA = 0.02 \, \text{g}`. Finally, we addopted the USGS recommendation for modifying the
```
raoanirudh, line 533: *adopted

openquake/sep/landslide/nowicki_jessee.py
```diff
@@ -0,0 +1,164 @@
...
+    b"70": 1.71,
+    b"90": -1.26,
+    b"100": 1.50,
+    b"110": 0.68,
+    b"120": 1.13,
+    b"130": 0.79,
+    b"140": 1.03,
+    b"150": 0.54,
+    b"160": 2.34,
+    b"170": 1.19,
```
raoanirudh, line 21: The paper doesn't provide a coefficient for class 170. How do we get the value for this class?

openquake/sep/landslide/nowicki_jessee.py
```diff
@@ -0,0 +1,164 @@
...
+    b"180": 1.19,
+    b"190": 0.30,
+    b"200": -0.06,
+    b"220": -0.18,
+    b"230": -1.08,
+    "14": 0.91,
```
raoanirudh, line 27: It would be good to check if we can find a way to avoid the duplication of coefficients for class codes represented as bytes or strings, because we're introducing more surface for errors to creep in.

openquake/sep/landslide/nowicki_jessee.py
```diff
@@ -0,0 +1,164 @@
...
+    "sm": -1.36,
+    "ss": -1.92,
+    "su": -1.36,
+    "va": -1.54,
+    "vb": -1.50,
+    "pi": -0.81
+}
+
+
+def sigmoid(x):
```
raoanirudh, line 81: Given this function is also used in the liquefaction classes, we should move it outside and import and reuse the same function everywhere instead of redefining.

raoanirudh (changes requested):

openquake/sep/landslide/nowicki_jessee.py
LanaTodorovic93: I was checking the description of 170 class and took the coefficient of the 'closest by description' class from table 3, but I will do one check and return back to this comment

openquake/sep/landslide/nowicki_jessee.py
LanaTodorovic93: will be addressed later.

openquake/sep/classes.py
```diff
@@ -208,6 +213,67 @@ def compute(self, mag, imt_gmf, sites):
...
+                pgv = gmf
+            elif im.string == "PGA":
+                pga = gmf
+            else:
+                continue
+        # Raise error if either PGA or PGV is missing
+        if pga is None or pgv is None:
+            raise ValueError(
+                "Both PGA and PGV are required to compute liquefaction "
+                "probability using the AllstadtEtAl2022Liquefaction model"
```
raoanirudh, line 259: Liquefaction warning in landslide model

openquake/sep/landslide/nowicki_jessee.py
```diff
@@ -0,0 +1,143 @@
...
+        Land cover, procxy for vegetation cover
+    :param cti:
+        Compound Topographic Index, a proxy for soil wetness.
+
+    :returns:
+        prob_ls: Probability of landslide.
+        coverage: Landslide areal coverage.
+    """
+
+    if isinstance(lithology, (str)):
```
raoanirudh, line 115: Do we need the tuple `(str)` or can it be simply `str`? Same for `(int)` on line 120 below.

openquake/sep/landslide/nowicki_jessee.py
```diff
@@ -0,0 +1,143 @@
...
+
+LANDCOVER_TABLE = {**landcover_values, **{bytes(k, 'utf-8'): v for k, v in landcover_values.items()}}
+LITHOLOGY_TABLE = {**lithology_values, **{bytes(k, 'utf-8'): v for k, v in lithology_values.items()}}
+
+
+def sigmoid(x):
+    return 1.0 / (1.0 + np.exp(-x))
+
+
+def _landslide_spatial_extent(a: float, b: float, c: float, d: float, p: float):
```
raoanirudh, line 54: If the coefficients `a`, `b`, `c`, `d` in this equation are constants, then it would be better to define them inside the function and take only `p` as an input to the function.

raoanirudh (changes requested):

openquake/sep/landslide/nowicki_jessee.py
LanaTodorovic93: we don't. fixed

openquake/sep/landslide/nowicki_jessee.py
LanaTodorovic93: fixed

openquake/sep/landslide/nowicki_jessee.py
```diff
@@ -0,0 +1,143 @@
...
+        lithology_coeff +
+        landcover_coeff +
+        cti_coeff * cti +
+        interaction_term * np.log(pgv) * slope +
+        intercept
+    )
+
+    prob_ls = sigmoid(Xg)
+    LSE = _landslide_spatial_extent(-7.592, 5.237, -3.042, 4.035, prob_ls)
+    LSE = np.where((slope < 2) | (pga < 0.02), 0, LSE)
```
raoanirudh, line 145: Are these the modifications proposed by Allstadt et al. (2022)? Would be helpful to add a comment line above to mention this, if so.

raoanirudh (approved):

openquake/hazardlib/imt.py
```diff
@@ -300,7 +300,7 @@ def LiqOccur():
 
 def LSE():
     """
-    Liquefaction spatial extent as percentage of a pixel area.
+    Liquefaction or Landslide spatial extent.
```
raoanirudh, line 303: On second thought, if the user is running a liquefaction analysis and a landslide analysis in the same job, the two spatial extent IMTs will need to be separated otherwise the second will overwrite the first. We might not be able to use the same IMT for both after all.

demos/risk/ScenarioRisk/job.zip
raoanirudh, line 1: Stray file, probably needs to be deleted?

LanaTodorovic93: removed

openquake/hazardlib/imt.py
LanaTodorovic93: Part of future enhancement

micheles (approved):

PR #10078, opened 2024-10-22
ARISTOTLE: Download and display intensity map from the USGS website if available

ptormene:
I am going to work on producing our own version of the intensity map and the pga map, but for now I am displaying those downloadable from the USGS website.

The `get_shakemap_version` function is useful in general to retrieve objects from urls formatted such as `https://earthquake.usgs.gov/product/shakemap/us7000n7n8/us/1726699735514/download/intensity.jpg`, where `us7000n7n8` is the USGS id and `1726699735514` is the shakemap version. With that function we can obtain the shakemap version given the USGS id.

image


ptormene:
I am closing this PR and I plan to open an alternative one soon

PR #10098, opened 2024-10-28
Add set -x and remove unused installation of sphinx theme

ptormene:
The zip file with the demos for engine-3.21 were never successfully produced (see e.g. https://github.com/gem/oq-engine/actions/runs/11174022834/job/31063030868), therefore corresponding QGIS plugin tests have been red for a while.
I slightly cleaned up the yml and launched the workflow manually and it produced the desired outcome.
We fixed the workflow to properly handle the case triggered by a push.


antonioettorre (approved):
LGTM

PR #10142, opened 2024-11-13
Update engine_pr_test.yml

vot4anto:
Add test for webui start

### DIFF
--- a/.github/workflows/engine_pr_test.yml
+++ b/.github/workflows/engine_pr_test.yml
@@ -101,6 +101,11 @@ jobs:
           source ~/openquake/bin/activate
           pip install pytest https://wheelhouse.openquake.org/v3/py/rtgmpy-1.0.0-py3-none-any.whl
           oq engine --upgrade-db
+          oq webui start 127.0.0.1:8800 -s &
+          echo "Waiting WEBUI up on port 8800...."
+          while ! nc -z localhost 8800; do
+           sleep 5 # wait for 1/10 of the second before check again
+          done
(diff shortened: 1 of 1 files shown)


.github/workflows/engine_pr_test.yml
```diff
@@ -101,6 +101,11 @@ jobs:
           source ~/openquake/bin/activate
           pip install pytest https://wheelhouse.openquake.org/v3/py/rtgmpy-1.0.0-py3-none-any.whl
           oq engine --upgrade-db
+          oq webui start 127.0.0.1:8800 -s &
+          echo "Waiting WEBUI up on port 8800...."
+          while ! nc -z localhost 8800; do
+           sleep 5 # wait for 1/10 of the second before check again
```
ptormene, line 107: Wouldn't it sleep for 5 seconds?
We might replace the line with
`sleep 5  # seconds`

.github/workflows/engine_pr_test.yml
```diff
@@ -101,6 +101,11 @@ jobs:
           source ~/openquake/bin/activate
           pip install pytest https://wheelhouse.openquake.org/v3/py/rtgmpy-1.0.0-py3-none-any.whl
           oq engine --upgrade-db
+          oq webui start 127.0.0.1:8800 -s &
```
ptormene, line 104: We might add a comment like:
`Starting the webui here is not needed by test_public_mode, but here we also want to check that the webui starts without errors`

ptormene:
Please refresh from master, in order to avoid the error while checking dependencies

ptormene (approved):

PR #10160, opened 2024-11-22
IMPACT: Plot shakemaps, ruptures and assets on top of basemaps obtained from external tiles services

ptormene:
Based on https://github.com/gem/oq-engine/pull/10157

It introduces a dependency from the package `contextily` (if we decide to merge this PR, we have to add that package to our weelhouse)

Some examples using CartoDB.Positron (we may choose among lots of others):
assets
avg_gmf-PGA
Screenshot From 2024-11-22 14-34-37


ptormene:
This PR is not urgent and it has been there for more than one year. Furthermore, now that we have the additional dependency from geopandas, some of these things can likely be made differently. I am closing it for now.

PR #10240, opened 2024-12-20
Move `reset_is_running` to be executed whenever the webui is ready

ptormene:
We need to check if this fixes the behavior after a stop / start of the webui service in production under gunicorn, in particular when the following conditions occur:
1) a job fails because the machine goes OOM, while there are some other pending jobs
2) we stop and restart the webui service
The expected outcome is that all the unfinished calculations will have the status "failed" and they do not prevent new calculations from running.

### DIFF
--- a/openquake/commands/webui.py
+++ b/openquake/commands/webui.py
@@ -22,7 +22,7 @@ import subprocess
 import webbrowser
 
 from openquake.baselib import config, general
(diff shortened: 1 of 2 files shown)


micheles (approved):

vot4anto (approved):
LGTM

PR #10248, opened 2025-01-08
Add a SUPPRESS_PERMISSION_DENIED_WARNINGS flag to the webui settings

ptormene:
...and add setting variable WEBUI_USER to override the default 'openquake'.

An open browser could keep polling to retrieve the list of calculations even after the session expires, producing in the logs a lot of warnings like `WARNING Forbidden: /v1/calc/list`.
Setting `SUPPRESS_PERMISSION_DENIED_WARNINGS = True` in the local settings it is possible to avoid such logging, still returning 403 responses.


vot4anto (approved):
LGTM

PR #10292, opened 2025-01-27
Revert the addition of pytest-django

ptormene:
It was causing troubles running tests in other repositories, due to a cleanup forced by pytest-django on the mail outbox that was attempted even if the outbox was undefined.


ptormene:
We did that instead: https://github.com/gem/oq-engine/pull/10293

PR #10337, opened 2025-02-12
When the PAM authentication backend is used, do not force the user to actualize email notification templates

ptormene:
For instance, in oqdb.gem.lan we have authentication administered via PAM and users do not even have associated email addresses to be used for email notifications. In cases like that, it would be annoying to force the actualization of email notification templates after installing the engine.


vot4anto (approved):
LGTM

PR #10356, opened 2025-02-17
In engine_weekly_test.yml add a step to actualize default templates for email notificatios before starting the webui

ptormene:
See https://github.com/gem/oq-engine/actions/runs/13366980697


vot4anto (approved):
LGTM

PR #10401, opened 2025-03-03
Import `timezonefinder` at runtime as it was before

ptormene:
This reverts commit 8a1863fa4e52b065c191aa399ed34eecee0e5652, that broke the MBTK on windows (because `timezonefinder` was recently added to the requirements for Linux, where it's needed for oq-impact, but not to the requirements of the other operating systems).


micheles (approved):
LGTM

PR #10417, opened 2025-03-06
Fixing bug in the std of Thingbaijam et al. (2017)

mmpagani:
The std to be returned is the one of the log of area. Bug discovered by V. Cascone


micheles (approved):

micheles:
Please add a line to the changelog

PR #10487, opened 2025-03-31
Reverted  https://github.com/gem/oq-engine/pull/10314

micheles:
Now the field name in the consequence files must be `risk_id` and not `taxonomy` to prevent confusions.


raoanirudh (commented):
LGTM

PR #10493, opened 2025-04-01
Fixing check in the Modifiable GMPE

mmpagani:


micheles:
It would be nice to have a test sensitive to this change.

PR #10528, opened 2025-04-15
Make completeness analysis robust to a certain common failure

cossatot:
This PR adds an option to return empty arrays upon failure of the `get_completeness_counts` algorithm. The default is to return the error, which was the previous behavior.


mmpagani (approved):
lgtm!

PR #10535, opened 2025-04-16
Allow free mapping between standalone apps and their url path prefix.

nastasi-oq:
Allow free mapping between standalone apps and their url path prefix (regardless of their name)


ptormene (approved):

openquake/server/settings.py
```diff
@@ -26,12 +26,20 @@
 from openquake.baselib import config
 from openquake.commonlib import datastore
 
+# optionally overrided in local_settings.py
```
ptormene, line 29: overridden

nastasi-oq: done

PR #10620, opened 2025-05-27
In the API endpoint listing the outputs for a job, specify the export type in the url

ptormene:
E.g. `http://localhost:8800/v1/calc/result/4489?export_type=csv` instead of just `http://localhost:8800/v1/calc/result/4489`.
For the sake of simplicity, the export_type will be the first available one for that output. In the future we may prefer to specify a "preferred" export type for each kind of output or to provide multiple urls, one for each available export format.


pslh (approved):
LGTM

micheles (approved):

PR #10624, opened 2025-05-29
Added a warning for missing site_labels

micheles:
As requested by @CB-quakemodel


CB-quakemodel:
> As requested by @CB-quakemodel

Grazie Michele!

PR #10634, opened 2025-06-04
AELO: improve the management of warnings

ptormene:
Instead of checking the presence of warnings by their text, we collect a mapping `warning_id: warning_text`, so we can modify the warning text without breaking any of the checks.

@ManuelaVillani and @kejohnso, in https://github.com/gem/oq-engine/pull/10626/files#diff-673eea543f153a9935e2890ee1a9baf39e14974776ab85a8192b081202afbb8bR514 you added some notes about code to be changed. Please, may you check about it?


openquake/calculators/postproc/compute_rtgm.py
```diff
@@ -750,17 +762,19 @@ def main(dstore, csm):
         dstore.create_df('rtgm', pd.concat(rtgm_dfs))
 
     plot_sites(dstore, update_dstore=True)
-    if rtgm_dfs and N == 1:  # and not warnings[sid]:
+    if rtgm_dfs and N == 1:
         sid = 0
-        if not warnings[sid].startswith(('Zero hazard', 'Very low hazard')):
+        if not warnings:
```
ptormene, line 767: Please @kejohnso and @ManuelaVillani check also if the creation of these plots is still made under the correct conditions

kejohnso: LGTM - the plots generated per case are tested in the qa tests for aelo.

PR #10668, opened 2025-06-19
IMPACT: improve rupture conversion in the workflow using the ShakeMap fault rupture

ptormene:
Using Anirudh's code to convert a wider set of ruptures into the OQ format.

I also added a check to identify cases in which the input data indicate that the hypocenter coordinates are outside the surface of the rupture. In those cases a warning is logged and presented to the user in the WebUI, and the hypocenter is relocated to the center of the surface of the rupture.


CB-quakemodel:
This will break GEESE which uses the current `convert_to_oq_rupture` which returns a `BaseRupture.` Would it be possible to split the new `convert_to_oq_xml` to return in a sub function the rupture object as is currently the case in the engine?

PR #10761, opened 2025-08-19
Exclude `'region'` and `'xvf'` from the columns of the `site_model` exported csv

ptormene:


openquake/calculators/export/hazard.py
```diff
@@ -470,13 +469,18 @@ def export_gmf_data_csv(ekey, dstore):
 @export.add(('site_model', 'csv'))
 def export_site_model_csv(ekey, dstore):
     sitecol = dstore['sitecol']
+    if os.environ.get('OQ_APPLICATION_MODE') == 'AELO':
```
ptormene, line 472: @CB-quakemodel does this look more reasonable? This way, columns are filtered only if the engine is running in AELO mode. @ManuelaVillani and @kejohnso, please let me know if the specified columns to keep are ok for AELO or if I should modify the list.

CB-quakemodel: Thanks @ptormene, this is what I personally had in mind. One side note - I don't know too much about AELO's inner-workings, but I was not aware sites in the site model could have a depth assigned as I currently see in the list of params to be retained. Perhaps this is representative of site elevation, or something else of use to the AELO applications rather than a true depth per say?

kejohnso (approved):

PR #10785, opened 2025-09-04
Make it possible to specify the mosaic_model_boundaries_file and mosaic_model_code_field_name in openquake.cfg

ptormene:
Fiona is smart enough to handle either a shp or a gpkg. I tried with the AELO Year 4 shapefile and with a gpkg obtained from it, without problems.
@kejohnso I noticed that the field containing the model codes in the AELO Year 4 file is `code` whereas in the past it was `name`, so I made also that configurable via openquake.cfg.
@micheles I see we have multiple versions of the mosaic boundaries in https://github.com/gem/oq-engine/tree/master/openquake/qa_tests_data/mosaic. We may add also the year 4 version, but perhaps we should keep only one version there, unless we need to run tests for multiple versions.
Another question for @micheles: I am putting these 2 new openquake.cfg variables in the `[directory]` section because they are related to the `mosaic_dir` and because I don't see a `[file]` section. I suppose it makes sense, but perhaps we could add a separate section for files.


micheles:
If `ModelBoundaries_Year3-4_v2.shp` is to be removed, it means that we don't need `mosaic_model_code_field_name`?

ptormene:
The mosaic model code field name has always been `name`, and it is `code` only in the new Year4 file that is still not in the engine repository. Therefore we might just rename that column in the Year 4 file, and keep using `name` as a convention.

micheles (approved):

PR #10814, opened 2025-09-24
added separate get_middle_point() method for kite surfaces

cossatot:
The method is robust to NaNs (as long as any of the vertices are finite). It finds the centroid of the non-NaN vertices, then finds the closest finite point to that vertex and returns it.


mmpagani (commented):
The method is acceptable, although it does not cover the case where the rupture crosses the IDL. You should add a test for this case and adjust the function accordingly.

cossatot:
Nonetheless I can add a test case for this.

PR #10822, opened 2025-09-25
Add bssa14 site term to mgmpe

CB-quakemodel:
This PR adds the BSSA14 site term to ModifiableGMPE + a separate GMPE class (BSSA14SiteTerm) + a set of unit tests.

The impact here is shown for some interface events (focal depth = 40 km) with vs30 of 500 m/s when the site term is applied to the AtkinsonMacias2009 GMM, which is a GMM which does not have a native site term. An identical result (as expected) is obtained when using either the ModifiableGMPE class OR the BSSA14SiteTerm class with the same underlying GMM - hence the visiblilty of two curves despite plotting 3 GMMs)

I also reduce the US23 GMM unit tests to just the modified versions of each GMM - the "default" versions are removed because they are really just artefacts of the testing phase of the implementation of the USGS adjustments rather than being useful now for QA.

TrellisPlots


openquake/hazardlib/gsim/mgmpe/bssa14_site_term.py
```diff
@@ -0,0 +1,123 @@
...
+        See :meth:`superclass method
+        <.base.GroundShakingIntensityModel.compute>`
+        for spec of input and result values.
+        """
+        # Make sites with ref bedrock vs30
+        rup_rock = copy.copy(ctx)
+        rup_rock.vs30 = np.full_like(ctx.vs30, 760.)
+
+        # Compute mean on bedrock
+        self.gmpe.compute(rup_rock, imts, mean, sig, tau, phi)
```
mmpagani, line 119: Isn't this already computing hazard on bedrock? Line 43 could be a repeated calculation.

openquake/hazardlib/gsim/mgmpe/modifiable_gmpe.py
```diff
@@ -378,21 +396,19 @@ def compute(self, ctx: np.recarray, imts, mean, sig, tau, phi):
...
-                'cy14_site_term' in self.params or
-                'ba08_site_term' in self.params):
+        if any(sm in self.params for sm in SITE_TERMS):
             ctx_copy = ctx.copy()
             if 'nrcan15_site_term' in self.params:
                 rock_vs30 = 760.
             elif 'cy14_site_term' in self.params:
                 rock_vs30 = 1130.
-            elif 'ba08_site_term' in self.params:
+            elif 'ba08_site_term' or 'bssa14_site_term' in self.params:
```
mmpagani: This could be combined with line 401

mmpagani (approved):
Apart from the two minor comments, I am fine with the PR. The stuff for the US model would have been better putting it into a separate PR.

openquake/hazardlib/gsim/mgmpe/bssa14_site_term.py
CB-quakemodel: I don't think so, if I understand the comment correctly. Here we compute the mean for bedrock vs30 for the first time when using the BSSA14SiteTerm gsim class. The other time the mean for bedrock vs30 is computed is in the ModifiableGMPE version with a kwarg indicating we wish to use the BSSA14 site term, which is not directly related, they both just call the same `get_bssa14_site_term` function, but in different ways for flexibility.

PR #10854, opened 2025-10-06
Update douglas_et_al_2024 model

aldabus:
Units incompatibility solved (m/s2 to g). Also, coefficients for the 3- and 5-branch models were updated to fix an error in the previous version


mmpagani (approved):
LGTM. Thank you

PR #10868, opened 2025-10-09
Various improvements in Django views for impact and aelo

ptormene:
- adding display names for `asce41_sa_final` and `spectra_asce41`
- displaying notifications (notes and warnings, if present) also on top of the standard (i.e. the "advanced") outputs page
- refactoring the view for the standard outputs page, to pass to the page only variables that are relevant with respect to the current application mode, and grouping all images into a `pngs` dictionary
- adding cross-validation in case both the site class and the Vs30 are specified (needed when the API is used programmatically)


openquake/calculators/export/__init__.py
```diff
@@ -24,6 +24,8 @@
 DISPLAY_NAME = {
     'asce07': 'ASCE 7 Parameters',
     'asce41': 'ASCE 41 Parameters',
+    'asce41_sa_final': 'ASCE41 BSE Spectra',
+    'spectra_asce41': 'ASCE41 Calculations',
```
ptormene, line 28: I'm using the display names suggested by @ManuelaVillani. Please @kejohnso double-check if they look fine to you as well.

PR #10946, opened 2025-11-07
change deprecated df.applymap => df.map

vot4anto:


ptormene (approved):
If tests are green, LGTM

PR #11048, opened 2025-12-11
Added output `avg_losses_by`

micheles:
The average losses aggregated by taxonomy (and possibly MACRO_TAXONOMY) are automatically computed if `collect_rlzs` is true. This is useful for the global risk model and also in general. Part of https://github.com/gem/oq-engine/issues/11022.


raoanirudh:
Thank you @micheles ! It would be helpful if we could add a similar output for exposure aggregated by TAXONOMY and MACRO_TAXONOMY, `aggexp_by`

Issue #10999, opened 2025-11-25
install.py installs always the first version of a release from PyPI


vot4anto:
https://github.com/gem/oq-engine/blob/d4ac52940b3102d03baae680c5cf783f47ec09fb/install.py#L472

For instance,` python install.py --version=3.23`

```
Collecting openquake.engine==3.23
  Downloading openquake.engine-3.23.0-py3-none-any.whl.metadata (13 kB)
```

But engine 3.23.4 exists:  https://pypi.org/project/openquake.engine/3.23.4/

PR #11057, opened 2025-12-15
When installing the engine like `--version=3.23` (without the patch number), get the latest available patch from PyPI

ptormene:
Fixes https://github.com/gem/oq-engine/issues/10999

### DIFF
--- a/.github/workflows/engine_nightly_user.yml
+++ b/.github/workflows/engine_nightly_user.yml
@@ -23,7 +23,7 @@ jobs:
       matrix:
         os: [ubuntu-24.04]
         python-version: ["3.11", "3.12"]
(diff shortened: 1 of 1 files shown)


install.py
```diff
@@ -386,6 +386,19 @@ def fix_version(commit, venv):
...
 
 
+def normalize_version(version):
+    """
+    Convert a user-supplied --version argument into a pip-compatible
+    version specifier.
+    """
+    if version.count('.') == 1:
+        # e.g. "3.23" -> latest patch in that series, expands to >=3.23.0,<3.24
+        return f"~={version}.0"
```
vot4anto, line 396: https://packaging.python.org/en/latest/tutorials/installing-packages/#installing-from-pypi

vot4anto (approved):
LGTM

PR #11084, opened 2025-12-22
Add GMMs Scala et al 2025 (BEEE, https://doi.org/10.1007/s10518-025-02315-6)

antonio-scala:
After pushing to my fork, some workflow actions fail.
I assume this is related to running the workflows in a forked repository and should not affect the pull request.
If not, I would appreciate any suggestions on how to address the issue.

### DIFF
--- a/openquake/hazardlib/gsim/scala_2025.py
+++ b/openquake/hazardlib/gsim/scala_2025.py
@@ -33,7 +33,6 @@ def _compute_distance(rval2, C, ctx, h):
     """
     Compute the distance function
     """
-
     rval = np.sqrt(rval2 ** 2 + h ** 2)
     return (C['c1'] + C['c2']*ctx.mag) * np.log10(rval)
 
@@ -96,8 +95,6 @@ class Scala2025CampiFlegreiRepiMW(GMPE):
     and Moment magnitude (Mw) as magnitude type.!!!
 
     """
-
-
     #: Supported tectonic region type is 'volcanic'
     DEFINED_FOR_TECTONIC_REGION_TYPE = const.TRT.VOLCANIC
 
@@ -178,10 +175,7 @@ class Scala2025CampiFlegreiRhypoMW(Scala2025CampiFlegreiRepiMW):
     """
     #: Required distance measure is Rhypo.
     REQUIRES_DISTANCES = {'rhypo'}
-
     H=1.0
-    #DIST_TYPE = 'rhypo'
-
     # Sigma values in log10
     COEFFS = CoeffsTable(sa_damping=5, table="""
     IMT		a		b		c1		 c2		 sB		sC		tau	 	phi     sigma0	sigma


openquake/hazardlib/gsim/scala_2025.py
```diff
@@ -0,0 +1,241 @@
...
+from openquake.hazardlib.gsim.base import GMPE, CoeffsTable
+from openquake.hazardlib import const
+from openquake.hazardlib.imt import PGA, PGV, SA
+
+
+def _compute_distance(rval2, C, ctx, h):   
+    """
+    Compute the distance function
+    """
+
```
CB-quakemodel: Spurious space

openquake/hazardlib/gsim/scala_2025.py
```diff
@@ -0,0 +1,241 @@
...
+    Bulletin of Earthquake Engineering.
+
+    GMPE derives from earthquakes in the volcanic area of Campi Flegrei in Italy 
+    in the magnitude range 1.5<Mw<4.0 (corresponding to a range 2.5<Md<4.4) 
+    for epicentral distances <40 km, and for stiff soil (EC8-B) and soft soil (EC8-C).
+
+    Test tables are generated from a spreadsheet provided by the authors, and
+    modified according to OQ format (e.g. conversion from cm/s2 to m/s2).
+    """
+
```
CB-quakemodel: Spurious space

openquake/hazardlib/gsim/scala_2025.py
```diff
@@ -0,0 +1,241 @@
...
+    REQUIRES_DISTANCES = {'repi'}
+    H=1.4
+    DIST_TYPE = 'repi'  # check kind
+    def compute(self, ctx: np.recarray, imts, mean, sig, tau, phi):
+        """
+        See :meth:`superclass method
+        <.base.GroundShakingIntensityModel.compute>`
+        for spec of input and result values.
+        """
+
```
CB-quakemodel: Spurious spaces here

openquake/hazardlib/gsim/scala_2025.py
```diff
@@ -0,0 +1,241 @@
...
+        """
+        See :meth:`superclass method
+        <.base.GroundShakingIntensityModel.compute>`
+        for spec of input and result values.
+        """
+        
+        
+        for m, imt in enumerate(imts):
+            C = self.COEFFS[imt]
+            rval_distance = getattr(ctx, self.DIST_TYPE)
```
CB-quakemodel: It would be cleaner I think to retrieve the dist type from `self.REQUIRES_DISTANCES` given the information is the same as that you are storing in `self.DIST_TYPE`

antonio-scala:
> Dear Antonio,
> 
> Many thanks for submitting your GMM in an OQ implementation.
> 
> I have made some small comments throughout the new files provided in your PR.
> 
> I think it would be advisable to incorporate all of the subclasses into a single python module (i.e. place the moment magnitude and duration magnitude subclasses into a single .py file, rather than having separate .py files for each - correspondingly the unit test files could be reduced by following this approach). This would make sense from a gsim library management perspective given it will keep the number of python modules down, and these GMMs belong to the same publication.
> 
> Many thanks,
> 
> Christopher

Dear Christopher,
many thanks for your feedback about my PR.

Before proceeding with the editing, regarding the suggestion to merge the subclasses into a single Python module, I would like to briefly explain why we initially decided to keep them separate and ask for your guidance on whether the merging is strictly required. 

The decision to calibrate two different magnitude–based formulations was motivated by a specific application context. Differently from common seismological applications, for the considered volcanic area, the official bulletins are released reporting the duration magnitude (Md). Having a Md-based GMM is therefore particularly useful for near–real-time applications, like the computation the ShakeMaps automatically computed and released by INGV (Italian National Institute of Geophysics and Volcanology).

In this view, keeping the two formulations in separate modules was aimed to facilitate their alternative use within OpenQuake-based ShakeMap workflows, where the input magnitude type is single and fixed by the upstream monitoring products.

That said, although I fully understand the management considerations you mention, I would like to ask if merging the subclasses into a single module is a strict requirement for inclusion, or whether the current separation could be acceptable given the peculiar use case.
Many thanks again for your feedback and guidance.
Best regards,
Antonio

mmpagani:
For example the ASK14 has four models in the same file https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/gsim/abrahamson_2014.py

antonio-scala:
> > > Dear Antonio,
> > > Many thanks for submitting your GMM in an OQ implementation.
> > > I have made some small comments throughout the new files provided in your PR.
> > > I think it would be advisable to incorporate all of the subclasses into a single python module (i.e. place the moment magnitude and duration magnitude subclasses into a single .py file, rather than having separate .py files for each - correspondingly the unit test files could be reduced by following this approach). This would make sense from a gsim library management perspective given it will keep the number of python modules down, and these GMMs belong to the same publication.
> > > Many thanks,
> > > Christopher
> > 
> > 
> > Dear Christopher, many thanks for your feedback about my PR.
> > Before proceeding with the editing, regarding the suggestion to merge the subclasses into a single Python module, I would like to briefly explain why we initially decided to keep them separate and ask for your guidance on whether the merging is strictly required.
> > The decision to calibrate two different magnitude–based formulations was motivated by a specific application context. Differently from common seismological applications, for the considered volcanic area, the official bulletins are released reporting the duration magnitude (Md). Having a Md-based GMM is therefore particularly useful for near–real-time applications, like the computation the ShakeMaps automatically computed and released by INGV (Italian National Institute of Geophysics and Volcanology).
> > In this view, keeping the two formulations in separate modules was aimed to facilitate their alternative use within OpenQuake-based ShakeMap workflows, where the input magnitude type is single and fixed by the upstream monitoring products.
> > That said, although I fully understand the management considerations you mention, I would like to ask if merging the subclasses into a single module is a strict requirement for inclusion, or whether the current separation could be acceptable given the peculiar use case. Many thanks again for your feedback and guidance. Best regards, Antonio
> 
> Dear Antonio,
> 
> Many thanks for your explanation. If I understand correctly, you need to be able to compute ShakeMaps using both Md and ML. With the suggested change, this would still be possible, given we are not suggesting removing either the Md or ML based GMMs, and instead are suggesting having all of these GMMs (each a subclass of the "parent" class) within a single GSIM module, rather than 2 as currently is the case.
> 
> You would still be able within the OQ-based ShakeMap workflows to use whichever GMM is contextually required (e.g. if you need the ML version, you could specify the use of `Scala2025CampiFlegrei_repi_ML` or `Scala2025CampiFlegrei_rhypo_ML`, and if you needed Md, you could specify the use of a `Scala2025CampiFlegrei_repi_Md` or `Scala2025CampiFlegrei_rhypo_Md`.
> 
> Hopefully this suggestion is agreeable to you, but of course please let me know if I'm misunderstanding your concerns.
> 
> Thanks,
> 
> Christopher

Thanks @CB-quakemodel, @mmpagani, @micheles for your feedbacks  

Your explanations were very clear and straight to the point. Following @mmpagani’s suggestion, I will try to implement the option to select the magnitude type, similarly to what is done in the Abrahamson et al. script for region selection. While doing so, I will make sure that file names no longer contain capital letters. If needed, I may reach out again for your support.

Thanks again,
Best,
Antonio

openquake/hazardlib/gsim/scala_2025.py
```diff
@@ -0,0 +1,321 @@
...
+    Bulletin of Earthquake Engineering. DOI: https://doi.org/10.1007/s10518-025-02315-6
+
+    GMPE derives from earthquakes in the volcanic area of Campi Flegrei in Italy
+    in the magnitude range 1.5<Mw<4.0 (corresponding to a range 2.5<Md<4.4)
+    for epicentral distances <40 km, and for stiff soil (EC8-B) and soft soil (EC8-C).
+
+    !!!This class considers Epicentral distance (Repi) as distance measure
+    and Moment magnitude (Mw) as magnitude type.!!!
+
+    Test tables are generated from a spreadsheet provided by the authors, and
```
CB-quakemodel: You can remove lines 97 and 98 from the docstring here (it's best to just place this info in the `scala_2025_test.py`

openquake/hazardlib/gsim/scala_2025.py
```diff
@@ -0,0 +1,321 @@
...
+
+def _get_site_amplification(ctx, C):
+    """
+    Compute the site amplification function given by FS = eiSi, for
+    i = 1,2 where Si are the coefficients determined through regression
+    analysis, and ei are dummy variables (0 or 1) used to denote the
+    different EC8 site classes. Due to the regression dataset only sites
+    B and C are considered. Hence in the current version site class A (rock)
+    is considered as equivalent to site class B, and site class D (soft soil)
+    is considered as equivalent to site class C."""
```
CB-quakemodel: Please place this `"""` on the next line (very trival but looks neater)

openquake/hazardlib/gsim/scala_2025.py
```diff
@@ -0,0 +1,321 @@
...
+    #: Required site parameter is Vs30
+    REQUIRES_SITES_PARAMETERS = {'vs30'}
+
+    #: Required rupture parameter is magnitude.
+    REQUIRES_RUPTURE_PARAMETERS = {'mag'}
+
+    #: Required distance measure is Repi.
+    REQUIRES_DISTANCES = {'repi'}
+    H=1.4
+    #DIST_TYPE = 'repi'  # check kind
```
CB-quakemodel: You can remove this hashed out line here and from all of the subclasses too

openquake/hazardlib/gsim/scala_2025.py
```diff
@@ -0,0 +1,321 @@
...
+    def compute(self, ctx: np.recarray, imts, mean, sig, tau, phi):
+        """
+        See :meth:`superclass method
+        <.base.GroundShakingIntensityModel.compute>`
+        for spec of input and result values.
+        """
+
```
CB-quakemodel: Please remove the two empty lines here

openquake/hazardlib/gsim/scala_2025.py
```diff
@@ -0,0 +1,321 @@
...
+    1.500   -2.3896	0.8644	-2.6398	 0.3780	 0.     0.2010	0.2425	0.2092	0.1781	0.3664
+    2.000   -2.6768	0.8664	-2.7034	 0.3972	 0.     0.2272	0.2509	0.2030	0.1844	0.3717
+    3.000   -2.7556	0.8177	-3.0000	 0.4472	 0.     0.1923	0.2360	0.1757	0.1917	0.3512
+    4.000   -2.7906	0.7817	-3.1544	 0.4670	 0.     0.1771	0.2207	0.1724	0.1965	0.3421
+    5.000   -2.7500	0.7332	-3.3747	 0.5232	 0.     0.1573	0.2095	0.1735	0.1991	0.3371
+    """)
+
+class Scala2025CampiFlegreiRhypoMW(Scala2025CampiFlegreiRepiMW):
+    """
+    Implements GMPE developed by Antonio Scala and co-authors (2025) and
```
CB-quakemodel: You do not need the entire docstring for the subclasses, you can just keep the part (at most) which states `This class considers HYPOCENTRAL DISTANCE (Rhypo) as distance measure
    and MOMENT MAGNITUDE (Mw) as magnitude type.`

openquake/hazardlib/gsim/scala_2025.py
```diff
@@ -0,0 +1,321 @@
...
+    for epicentral distances <40 km (but considering Hypocentral distance as a covariate),
+    and for stiff soil (EC8-B) and soft soil (EC8-C).
+
+    !!!This class considers HYPOCENTRAL DISTANCE (Rhypo) as distance measure
+    and MOMENT MAGNITUDE (Mw) as magnitude type.!!!
+
+    Test tables are generated from a spreadsheet provided by the authors, and
+    modified according to OQ format (e.g. conversion from cm/s2 to m/s2).
+    """
+
```
CB-quakemodel: Please remove the 2 empty lines

openquake/hazardlib/gsim/scala_2025.py
```diff
@@ -0,0 +1,321 @@
...
+        const.StdDev.TOTAL, const.StdDev.INTER_EVENT, const.StdDev.INTRA_EVENT}
+
+    #: Required site parameter is Vs30
+    REQUIRES_SITES_PARAMETERS = {'vs30'}
+
+    #: Required rupture parameter is magnitude.
+    REQUIRES_RUPTURE_PARAMETERS = {'mag'}
+
+    #: Required distance measure is Rhypo.
+    REQUIRES_DISTANCES = {'rhypo'}
```
CB-quakemodel, line 177: You can remove all of the other `DEFINED_FOR_...` and `REQUIRES_...` from this class, it is the same as for the parent class and therefore is inhereited, you just need to keep the `REQUIRES_DISTANCES = {'rhypo'}` line to ensure this info is overwritten in the parent class

openquake/hazardlib/gsim/scala_2025.py
```diff
@@ -0,0 +1,321 @@
...
+    Implements GMPE developed by Antonio Scala and co-authors (2025) and
+    submitted as "Ground Motion Models for Campi Flegrei (Italy)"
+    Bulletin of Earthquake Engineering. DOI: https://doi.org/10.1007/s10518-025-02315-6
+
+    GMPE derives from earthquakes in the volcanic area of Campi Flegrei in Italy
+    in the magnitude range 1.5<Mw<4.0 (corresponding to a range 2.5<Md<4.4)
+    for epicentral distances <40 km (but considering Hypocentral distance as a covariate),
+    and for stiff soil (EC8-B) and soft soil (EC8-C).
+
+    !!!This class considers EPICENTRAL DISTANCE (Repi) as distance measure
```
CB-quakemodel, line 206: Again you do not need such a long docstring for a subclass. You can here just keep at most `This class considers EPICENTRAL DISTANCE (Repi) as distance measure and DURATION MAGNITUDE (Md) as magnitude type.`

openquake/hazardlib/gsim/scala_2025.py
```diff
@@ -0,0 +1,321 @@
...
+    and DURATION MAGNITUDE (Md) as magnitude type.  The implementation of this model became
+    necessary for practical purposes, as the official bulletin for the area uses Md
+    as the magnitude measure, as detailed in the paper.!!!
+
+    Test tables are generated from a spreadsheet provided by the authors, and
+    modified according to OQ format (e.g. conversion from cm/s2 to m/s2).
+    """
+
```
CB-quakemodel: You can remove here lines 268 to 295 because the only info we need to overwrite is the coefficients used, all of the other info can be safely inherited from the parent class

PR #11087, opened 2025-12-29
Normalize the existing job_tag schema

ptormene:
- Introducing a separate tag table to hold distinct tag values
- Reworking job_tag into a proper many-to-many join table between job and tag, with its own primary key
- Preserving existing tag data and preferred tag associations during migration
- Improving consistency, constraints, and future extensibility

I had to re-implement all queries and all the corresponding admin interface, also adding a few new API endpoints.

I also fixed the TagAdminSite specifying the site_url taking into account settings.WEBUI_PATHPREFIX like it is done for the AdminSite. This way, if the WEBUI_PATHPREFIX is specified, the `VIEW SITE` button from the Job Tags Management links to the correct url.

### DIFF
--- a/.github/workflows/engine_pr_test.yml
+++ b/.github/workflows/engine_pr_test.yml
@@ -231,7 +231,7 @@ jobs:
     env:
       GITHUB_HEAD_REF: ${{ github.head_ref }}
       FROM_FORK: ${{ github.event.pull_request.head.repo.fork }}
-      OQ_APPLICATION_MODE: ARISTOTLE
+      OQ_APPLICATION_MODE: IMPACT
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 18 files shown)


openquake/server/db/actions.py
```diff
@@ -669,10 +674,35 @@ def share_job(db, job_id, share):
...
+    return rows[0].id
+
+
 def add_tag_to_job(db, job_id, tag_name):
     try:
-        db("INSERT INTO job_tag (job_id, tag, is_preferred) VALUES (?x, ?x, 0)",
-           job_id, tag_name)
+        tag_id = _get_or_create_tag_id(db, tag_name)
+        db(
+            """
```
micheles: Perhaps there is an excessive newline?

Issue #9897, opened 2024-08-08
New CAV models


emthompson-usgs:
Hi @mmpagani and @micheles, 

@kslytherin and I are going to start working on a few new CAV models to contribute, specifically: 
- [Bullock et al. 2019](https://doi.org/10.5459/bnzsee.52.4.193-207)
- [Bullock et al. 2021](https://doi.org/10.1177/8755293020957342)
- [Liu and Macedo, 2022](https://doi.org/10.1177/87552930211043897)

This issue's main purpose is to inform you about our plans and try to avoid duplicating effort, which happened for the [Campbell and Bozorgnia model](https://github.com/gem/oq-engine/pull/9535). 

Also, the Liu and Macedo paper includes a conditional model, so I wanted to ask what the strategy for these models in OQ will be. If @g-weatherill's implementation in [this PR](https://github.com/gem/oq-engine/pull/9197) has issues that prevent it from being merged, could we discuss how it needs to be modified to move things forward? 

Thanks.

emthompson-usgs:
Also, I'll note that @kslytherin has an implementation of the [Macedo et al. 2020](https://doi.org/10.1785/0120190321) model that uses the `ConditionalGMPE` class by @g-weatherill. I'm pointing this out mostly to indicate that this is an important topic for us, and hopefully, OQ will also benefit from our contributions. If there are changes that are needed to the `ConditionalGMPE` class, we'll be happy to update the Macedo et al. 2020 implementation to reflect those changes.

mmpagani:
> Hi @mmpagani and @micheles,
> 
> @kslytherin and I are going to start working on a few new CAV models to contribute, specifically:
> 
> * [Bullock et al. 2019](https://doi.org/10.5459/bnzsee.52.4.193-207)
> * [Bullock et al. 2021](https://doi.org/10.1177/8755293020957342)
> * [Liu and Macedo, 2022](https://doi.org/10.1177/87552930211043897)
> 
> This issue's main purpose is to inform you about our plans and try to avoid duplicating effort, which happened for the [Campbell and Bozorgnia model](https://github.com/gem/oq-engine/pull/9535).
> 
> Also, the Liu and Macedo paper includes a conditional model, so I wanted to ask what the strategy for these models in OQ will be. If @g-weatherill's implementation in [this PR](https://github.com/gem/oq-engine/pull/9197) has issues that prevent it from being merged, could we discuss how it needs to be modified to move things forward?
> 
> Thanks.

[...]

PR #11090, opened 2026-01-04
Completely refactor conditional GMPEs to be specified within ModifiableGMPE for better/safer handling

CB-quakemodel:
**CONTEXT**

The way conditional GMPEs are currently handled is quite messy and there are several problems with their current implementations which we need to correct.

These issues include:

1. Currently, we cannot specify a different ConditionalGMPE for each IMT. This new approach allows the specification of a different conditional GMPE for each required IMT as follows (here for IA and PGV) using ModifiableGMPE syntax:

image

2. In the current implementations, if the conditional GMPE does not support a requested IMT, then the underlying GSIM is used instead **without warning**. This behaviour is silent and therefore rather dangerous (the user might not realise that the predictions for an IMT are actually from the underlying GMM rather than the conditional GMPE they mistakenly thought could be used for the given IMT). The silent provision of the underlying GMPE's predictions is something that should always be forbidden.

3. There are currently some hackish approaches for ensuring that all of the IMTs required for a calculation are available (which are causing the silent passing of the underlying GSIMs predictions if not supported by the conditional GMPE). The new approach safely manages which IMT's predictions are from the underlying GSIM and which are from the conditional GMPEs (see the compute methods of ModifiableGMPE and the currently available conditional GMPEs in OQ).

**PR FEATURES**

This PR addresses these issues by using ModifiableGMPE and a dictionary style syntax to permit explicit use of a different conditional GMPE (when required) for each IMT. The base GMM is then used for all other IMTs specified within the job file. The currently available conditional GMPEs (Macedo et al. 2019 and Abrahamson and Bhasin 2020) are refactored accordingly to work with this approach (now they are much simpler - additional conditional GMPEs can use these as implementation guides).

All of the existing GSIM unit tests and QA tests using conditional GMPEs pass using this ModifiableGMPE approach/syntax.

For the use of ModifiableGMPE to specify the conditional GMPE for each period, a modified (and fully viable) version of the each conditional GMPE's `compute` method is used (which is much cleaner/simpler too given it removes the hackish approaches mentioned above).

The documentation has been updated to provide guidance on implementing conditional GMPEs. One of the key differences is that conditional GMPEs require `REQUIRES_IMTS` (which is different to `DEFINED_FOR_INTENSITY_MEASURE_TYPES`) and uses a modified `compute` method. A flag of `conditional = True` must be set in the parent GSIM class too which is checked when the conditional GMPE is being instantiated within ModifiableGMPE (prevent non-conditional GMPEs being used in this ModifiableGMPE capability).

This PR is of interest to (hopefully) definitively resolving the conditional GMPE issues discussed previously in https://github.com/gem/oq-engine/issues/9897

### DIFF
--- a/openquake/hazardlib/gsim/mgmpe/modifiable_gmpe.py
+++ b/openquake/hazardlib/gsim/mgmpe/modifiable_gmpe.py
@@ -103,10 +103,10 @@ def conditional_gmpe_setup(self, imts, ctx_copy, mean, sig, tau, phi):
         imt.string: {} for imt in imts_req}
     
     for i, imt in enumerate(imts_req):
-        self.params["conditional_gmpe"]["base_preds"][str(imt)]["mean"] = mean_b[i]
-        self.params["conditional_gmpe"]["base_preds"][str(imt)]["sig"] = sig_b[i]
-        self.params["conditional_gmpe"]["base_preds"][str(imt)]["tau"] = tau_b[i]
-        self.params["conditional_gmpe"]["base_preds"][str(imt)]["phi"] = phi_b[i]
+        self.params["conditional_gmpe"]["base_preds"][imt.string]["mean"] = mean_b[i]
+        self.params["conditional_gmpe"]["base_preds"][imt.string]["sig"] = sig_b[i]
+        self.params["conditional_gmpe"]["base_preds"][imt.string]["tau"] = tau_b[i]
+        self.params["conditional_gmpe"]["base_preds"][imt.string]["phi"] = phi_b[i]
         
     # Sometimes need underlying GSIM within conditional GMPEs compute method if has
     # ctx-dependent conditioning periods like possible within AbrahamsonBhasin2020
@@ -135,7 +135,7 @@ def conditional_gmpe_setup(self, imts, ctx_copy, mean, sig, tau, phi):
         arrays = [mean.copy(), sig.copy(), tau.copy(), phi.copy()]
         reordered = [np.zeros_like(arr) for arr in arrays]
 
-        # for instance in the test case_90 one has
+        # For instance in the test case_90 one has
         # imts_map = {PGA: 0, PGV: 1, IA: 2, SA(0.2): 3, SA(1.0): 4}
         # and imts_base = {SA(1.0), SA(0.2), PGA}
         for idx, imt in enumerate(imts_base):
@@ -157,7 +157,7 @@ def conditional_gmpe(ctx, imt, me, si, ta, ph, **kwargs):
     base_preds = kwargs.get('base_preds')
     
     # If the imt is requested from a conditional gmpe... 
-    if str(imt) in conditional_gmpes: 
+    if imt.string in conditional_gmpes: 
 
         # Get the conditional GMM specified for the given IMT
         cond = conditional_gmpes[imt.string]["gsim"]


openquake/hazardlib/gsim/base.py
```diff
@@ -284,6 +293,12 @@ def __init__(self, **kwargs):
                    'may not be as expected - '
                    'the user is liable for their application') % cls.__name__
             warnings.warn(msg, AdaptedWarning)
+        if cls.conditional:
+            msg = ('%s is a conditional GMPE - the user must specify such GMPEs '
+                   'within ModifiableGMPE in combination with a base GMPE to '
+                   'condition the ground-motions upon (i.e., this GMPE cannot be '
+                   'instantiated as a "regular" standalone GMPE)') % cls.__name__
+            warnings.warn(msg, ConditionalWarning)
```
micheles: Why a warning and not an error?

openquake/hazardlib/tests/gsim/utils.py
```diff
@@ -223,7 +223,11 @@ def check(self, *filenames, max_discrep_percentage,
             normalize(fnames)
             return
 
-        gsim = self.GSIM_CLASS(**kwargs)
+        if self.GSIM_CLASS.__class__.__name__ == "ModifiableGMPE":
+            gsim = self.GSIM_CLASS # GSIM from valid.modified_gsim
```
micheles: It is really ugly to use the name `GSIM_CLASS` to mean a gsim instance :-/ Can't we just use the name `GSIM`?

openquake/hazardlib/tests/gsim/abrahamson_bhasin_2020_test.py
```diff
@@ -15,38 +15,46 @@
...
+BASE_GMM = gsim("Lin2009")
+
+
 class AbrahamsonBhasin2020TestCase(BaseGSIMTestCase):
-    GSIM_CLASS = AbrahamsonBhasin2020
+    GSIM_CLASS = modified_gsim(
+        BASE_GMM,
+        conditional_gmpe={
+            "conditional_gmpes": 
+            {"PGV": {"gmpe": {"AbrahamsonBhasin2020": {}}}}})
```
micheles: Can't we remove a level of nesting? I.e.remove `conditional_gmpe` and have have the subdictionary `conditional_gmpes` ?

openquake/hazardlib/gsim/base.py
CB-quakemodel: I've improved this so that an error is strictly raised if a conditional GMPE is not instantiated from within ModifiableGMPE.

openquake/hazardlib/tests/gsim/abrahamson_bhasin_2020_test.py
CB-quakemodel: I have now removed one level of nesting which simplifies things a lot, especially in the XMLs

openquake/hazardlib/tests/gsim/utils.py
CB-quakemodel: I'm not too sure I understand this comment? Here `self` is an instance of `BaseGSIMTestCase` which has no `GSIM` attribute, and the same was done here previously for 'regular' GSIMs?

micheles: `modified_gsim` returns a gsim instance, not a gsim class, so something like
```python
 GSIM_CLASS = modified_gsim(
        BASE_GMM,
        conditional_gmpe={"PGV": {"gmpe": {"AbrahamsonBhasin2020": {}}}})
```
is misleading; also, in normal subclasses of `BaseGSIMTestCase` `GSIM_CLASS` is indeed a class. So I am saying that for the tests using modifiable gmpes the name of the attribute should be `GSIM` and not `GSIM_CLASS` and `utils.py` should be changed as follows:
```python
if hasattr(self, 'GSIM'):  # for ModifiableGMPE tests
    gsim = self.GSIM
else:  # assume a GSIM_CLASS attribute for regular tests
    gsim = self.GSIM_CLASS(**kwargs)
```

CB-quakemodel: Thanks for the clarification, I have just made the required modifications

doc/contributing/implementing-new-gsim.md
```diff
@@ -38,3 +38,36 @@ Therefore assignment statements such as REQUIRES_RUPTURE_PARAMETERS = {'mag'} an
 
 - Update the changelog file 
 https://github.com/gem/oq-engine/blob/master/debian/changelog following the [Developers notes](updating-the-changelog.md)
+
+# Implementing a new conditional GSIM in the OpenQuake hazard library
+
+Conditional GMPEs within OpenQuake are specified within `ModifiableGMPE`. We use `ModifiableGMPE` because it allows us to easily specify each conditional GMPE
+that we wish to use to compute each IMT (probably) not supported by the underlying GSIM which we are conditioning the predictions of the conditional GMPE upon.
```
mmpagani: "that we wish to use to compute each IMT (probably) not supported by the underlying GSIM" I would write it as "that we wish to use to compute scalar intensity measure types (IMTs) not supported by the underlying GSIM"

mmpagani (approved):
Thank you Chris. Nice work!

doc/contributing/implementing-new-gsim.md
CB-quakemodel: Thanks Marco, I've rephrased as suggested

openquake/hazardlib/gsim/mgmpe/modifiable_gmpe.py
```diff
@@ -47,33 +50,152 @@
...
+    self.params["conditional_gmpe"]["base_preds"]["base"] = self.gmpe
+
+    # Now get the IMTs we wish to compute using the underlying GSIM
+    imts_gmm = [imt.__qualname__ for imt in
+                self.gmpe.DEFINED_FOR_INTENSITY_MEASURE_TYPES]
+    imts_bse = []
+    for imt in imts:
+        if str(imt).split("(")[0] in imts_gmm and imt not in imts_bse:
+            imts_bse.append(imt)
+
```
micheles: This code is ugly, let us discuss it later today

openquake/hazardlib/gsim/mgmpe/modifiable_gmpe.py
```diff
@@ -408,8 +550,15 @@ def compute(self, ctx: np.recarray, imts, mean, sig, tau, phi):
...
         g = globals()
         
-        # Compute the original mean and standard deviations
-        self.gmpe.compute(ctx_copy, imts, mean, sig, tau, phi)
+        # If necessary, compute the means and std devs for the required
+        # IMTs that are not going to be calculated using conditional GMPEs 
+        if "conditional_gmpe" in self.params:
+            conditional_gmpe_setup(self, imts, ctx_copy, mean, sig, tau, phi)
+            
+        # Otherwise, compute the original mean and std devs for all
```
micheles: The comment should go below the `else`

Issue #10438, opened 2025-03-12
Import from engine.js separate blocks corresponding to different application modes


ptormene:
This would also avoid annoying warnings such as the impossibility to get default impact form values when using other application modes.

PR #11110, opened 2026-01-14
Refactor the WebUI code separating its logic into multiple files and add Playwright tests to test the frontend

ptormene:
This PR supersedes (and includes) the refactoring done in https://github.com/gem/oq-engine/pull/11113.
In order to be more confident about the fact that the refactoring would not cause regressions, I decided to implement tests for the frontend application, that still were missing. While checking the refactoring with @hascar, we also made some minor improvements to the UI usability, e.g. we added spinners during processing time and we better handled the synchronization of some UI elements.
The aim of playwright tests is to test the UI only, not the logic behind the API. For instance, I have added an example using the login form in RESTRICTED mode, and another one filling the AELO form, running an AELO calculation and aborting it afterwards.

I added UI tests for the RESTRICTED and AELO application modes to the CI running at each pull request. For the IMPACT mode, that is much slower and that interacts with third-party services, I created a new CI routine running nightly. Here are examples of a successful run (https://github.com/gem/oq-engine/actions/runs/21244268803) and of a failure (https://github.com/gem/oq-engine/actions/runs/21243945416).

With the current setup, if a playwright test fails we have the error message in the log, plus a link to download artifacts (created only on failure). Artifacts contain a screnshot of the UI at the moment of failure, a video of the UI steps of the test and a trace that contains information that currently is not very clear to me.

Tests for the impact workflows include one example for the level 1 interface and one example for each of the workflows available in the level 2 interface. Those tests are still incomplete. They do not cover some functionalities (e.g. reading stations from a file uploaded by a user), and they do not check the visibility/invisibility of all form elements depending from the chosen workflow. However they are a reasonable starting point and they test a full end-to-end run for each main use case.

Fixes https://github.com/gem/oq-engine/issues/10438


hascar (approved):
LGTM

openquake/server/tests/pages/engine_page.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def latest_job_row(self):
+        """
+        Return the most recently created job row.
+        Assumes newest job appears first.
+        """
+        return self.calculation_table.locator("tbody tr").first
+
+    def abort_latest_job(self):
+        job_row = self.latest_job_row()
```
micheles, line 17: Why the timeouts are so long?

ptormene: When we retrieve data from the USGS, I don't want tests to fail just because the response is slow. In this specific context, the calculation may take some time from the instant you press the button to launch it, to the moment the job is displayed as "executing". Anyway, if things go quicker, the test continues without waiting the timeout. For these tests that run nightly, in general, I am not so much worried about speed, as long as all the steps of the workflow are completed correctly. On the other hand, I am worried about possible failures that might occur when the workflow happens to be a bit slower than expected.

micheles: I just thought the timeout was in seconds, not in milliseconds.

PR #11113, opened 2026-01-14
Refactor the WebUI javascript and css code separating its logic into multiple files

ptormene:
This is a first step towards a proper separation between the different components of the WebUI. We had a single CSS and a single JS, that included general utilities, the code managing the list of calculations, the AELO and the IMPACT specific elements.
With this refactoring, I am redistributing those elements into separate files, so we can include in each specific context only the elements that are required. I was careful moving code without applying any unnecessary changes.


ptormene:
Superseded by https://github.com/gem/oq-engine/pull/11110

PR #11128, opened 2026-01-20
Fix edge case where no AvgSA IMTs specified but using an indirect AvgSA GSIM

CB-quakemodel:
There is an edge case where the user might only have "regular" IMTs in the job file, but is still using an indirect approach GSIM, with the assumption being their intention is to use the underlying GMPE to compute the non-AvgSA IMTs (and have not yet added AvgSA IMTs to the job file but will do in a subsequent calculation - this is a niche edge case but one worth considering).

In my initial implementation (https://github.com/gem/oq-engine/pull/11122), such scenarios would lead to `_get_periods` failing because it would try to concatenate on a empty list (no AvgSA IMTs). Now, it returns an empty list instead for the averaging periods, which means the "AvgSA part" of the compute method is being skipped as is correct in such an edge case (no AvgSA IMTs and therefore no averaging periods, so no AvgSA to compute either).

A unit test has been added checking this edge case is now resolved.

### DIFF
--- a/openquake/engine/aelo.py
+++ b/openquake/engine/aelo.py
@@ -136,6 +136,7 @@ def main(lon: valid.longitude,
          lat: valid.latitude,
          vs30: valid.positivefloat,
          siteid: str,
+         description: str,
          asce_version: str,
          site_class: str,
          jobctx,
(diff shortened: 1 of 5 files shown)


openquake/calculators/tests/classical_test.py
```diff
@@ -41,7 +41,7 @@
     case_60, case_61, case_62, case_63, case_64, case_65, case_66,
     case_67, case_69, case_70, case_72, case_74, case_75, case_76, case_77,
     case_78, case_80, case_81, case_82, case_83, case_84, case_85,
-    case_86, case_87, case_88, case_89, case_90, case_91)
+    case_86, case_87, case_88, case_89, case_90, case_91, case_92)
```
micheles, line 45: I would rename this to case_68, so to fill a long standing hole ;-)

CB-quakemodel: Done!

mmpagani (approved):

PR #11130, opened 2026-01-20
Store both the description and the site id when an AELO calculation runs from the WebUI…

ptormene:
…so the original description becomes visible in the outputs page


ptormene:
I made some internal renaming to avoid confusion. In the form we have a `Site name` label (that is the one requested in the specifications), so I replaced the ambiguous internal variable name `siteid` with `site_name` and I added notes in the code when the `site_name` is used to obtain the job `description` and when it is transformed into an actual `siteid` (without spaces, etc.).

PR #11136, opened 2026-01-21
Use of new taxonomy tools and introduction of postinstall django app commands management

nastasi-oq:
This PR add to the stable branch of engine the management of new taxonomy tools.
When it will work we can update master branch too.


install.py
```diff
@@ -263,37 +263,75 @@ def get_requirements_branch(version, inst, from_fork):
         return version
 
 
-def install_standalone(venv):
+def install_or_postinstall_standalone(venv, is_install=True):
     """
```
micheles, line 266: Please rename `is_install` as `installed`

micheles (approved):

install.py
nastasi-oq: why do it? it tell if is `install` or `postinstall`, not if are installed or not the sources.
And in any case if I change it I must invert the logic, no?

nastasi-oq: @micheles maybe you miss the point: this method perform 2 things: **install** or **postinstall** , that argument define which behaviour perform.

PR #11138, opened 2026-01-21
Add helper to include demos folder 

vot4anto:


setup.py
```diff
@@ -24,6 +24,15 @@
 if sys.version_info < (3, 11):
     sys.exit('Sorry, Python < 3.11 is not supported')
 
+# --- 1. Helper Function for Demos ---
+def generate_data_files(source_dir):
+    data_files_list = []
+    for root, dirs, files in os.walk(source_dir):
+        if files:
+            target_dir = root
```
micheles: Seems useless, just call root target_dir in the first place

install.py
```diff
@@ -505,14 +505,6 @@ def install(inst, version, from_fork):
...
                  "--upgrade", GITBRANCH % commit, "--no-clean"],
                 env=custom_env)
-
-            for build in os.listdir(tmp):
-                if build.startswith("pip-req-build-"):
-                    demos= os.path.join(tmp, build, 'demos')
-                    shutil.copytree(demos, os.path.join(inst.VENV, 'demos'))
-                    break
-            else:
-                raise RuntimeError('Could not find where pip installed the zip')
```
micheles, line 515: I am not sure if removing this will break the tests for engine-3.24 (maybe not, did you try?)

vot4anto: I will try now

vot4anto: https://github.com/gem/oq-engine/actions/runs/21213555925

PR #11145, opened 2026-01-23
IMPACT: add `exposure_by_location` extractor

ptormene:
This new extractor is very similar to the existing one for `losses_by_location`.

### DIFF
--- a/openquake/calculators/extract.py
+++ b/openquake/calculators/extract.py
@@ -1101,7 +1101,7 @@ def extract_losses_by_site(dstore, what):
     return pandas.DataFrame(dic)
 
 
-def _group_assets_by_location(assetcol):
+def group_assets_by_location(assetcol):
     """
     Group assets by rounded (lon, lat).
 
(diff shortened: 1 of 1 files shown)


openquake/calculators/extract.py
```diff
@@ -1132,6 +1132,38 @@ def extract_losses_by_location(dstore, what):
...
+        name for name in assetcol[:].dtype.names
+        if name.startswith('value-') or name.startswith('occupants_')
+    ]
+    for field in fields:
+        values = assetcol[field][lonlats['ordinal']]
+        dic[field] = F32(general.fast_agg(indices, values))
+    logging.info('There are {:_d} assets on {:_d} locations'.format(
+            len(lonlats), len(lons)))
+    return pandas.DataFrame(dic)
+
```
micheles, line 1169: Maybe add a helper function to avoid duplicating too much logic?

ptormene: Done

openquake/calculators/extract.py
```diff
@@ -1101,29 +1101,42 @@ def extract_losses_by_site(dstore, what):
     return pandas.DataFrame(dic)
 
 
+def _group_assets_by_location(assetcol):
+    """
```
micheles, line 1105: The convention is: docstring for public functions, no docstring for private functions. So I would remove the underscore here.

PR #11154, opened 2026-01-27
Support GMPE Tables in Conditional GMPEs

CB-quakemodel:
Bit tricky because of how the tables are set within the context maker but we need this to ensure NGAEast models (and other GMMS that use tables) are supported as underlying models for conditional GMPEs.

Also fix a typo where `from_mgmpe` in `init_underlying_gmpes` is being set as `from_mgpme` in ModifiableGMPE.

A unit test (classical/case_71) using conditional GMPEs with the underlying GMPE being an NGAEast model is added.

### DIFF
--- a/openquake/hazardlib/gsim/gmpe_table.py
+++ b/openquake/hazardlib/gsim/gmpe_table.py
@@ -86,10 +86,9 @@ def _return_tables(self, mag, imt, which):
     """
     assert which in "IMLs Total", which
     if imt.string not in self.imls and imt.name != "SA":
-        # Scalar IMT is not supported (conditional GMPEs -
-        # we in effect skip setting a table which ensures
-        # an error is still raised when an unsupported IMT
-        # is specified outside of conditional GMPE usage)
+        # Scalar IMT is not supported (conditional GMPEs - we in skip
+        # setting a table which ensures an error is still raised when
+        # unsupported IMT is specified outside of conditional GMPE use)
         return
     elif imt.string in ("PGA", "PGV"): 
         # Get supported scalar imt


openquake/hazardlib/gsim/gmpe_table.py
```diff
@@ -85,16 +85,21 @@ def _return_tables(self, mag, imt, which):
        the string "IMLs" or "Total"
     """
     assert which in "IMLs Total", which
-    if imt.string in 'PGA PGV':
-        # Get scalar imt
+    if imt.string not in self.imls:
+        # IMT is not supported (case of conditional GMPEs where
+        # we need to still set an arbitrary table in ctx maker)
+        ec = self.imls[next(iter(self.imls))].shape
```
micheles: Can you be more explicit about what `ec[0]` and `ec[2]` are?

CB-quakemodel: I've just clarified this a bit (I think)

openquake/hazardlib/gsim/gmpe_table.py
```diff
@@ -85,16 +85,21 @@ def _return_tables(self, mag, imt, which):
        the string "IMLs" or "Total"
     """
     assert which in "IMLs Total", which
-    if imt.string in 'PGA PGV':
-        # Get scalar imt
+    if imt.string not in self.imls and imt.name != "SA":
```
CB-quakemodel, line 88: @micheles ideally it would be nice to only permit this empty array being returned in the case of a modifiable gmpe being used (e.g. `from_mgmpe = True`) but I cannot see a way to pass this info given the way the context maker works with setting the tables very early and only the underlying gsim being contained in `self` here.

Note that using an unsupported IMT in gmpe tables outside of conditional GMPEs still causes a failure, but like before this PR it's still not a clear one, so I was wondering if you had an idea for this? It would be nice to avoid the GSIM only failing because of something ambiguous.

Ideally it would use something like this:

Image

mmpagani (approved):

openquake/hazardlib/gsim/gmpe_table.py
CB-quakemodel: It turns out safest way is to just return nothing - therefore if an unsupported IMT by the "underlying" GMPE table GSIM is ever actually requested, a key error is always raised like before because no arbitrary table is ever set at all.

PR #11167, opened 2026-01-30
Refactor: Make avg_periods optional in BaseAvgSACorrelationModel and Remove static methods

nc-hsu:
Motivation: Currently, The correlation models in this module are used in either GenericGmpeAvgSA or GmpeIndirectAvgSA classes. In GenericGmpeAvgSA the correlation models are instantiated and require avg_periods, however in GmpeIndirectAvgSA they are not instantiated correlation models are only static methods are used to obtain the correlation coefficients.

This inconsistency makes it more cumbersome to add new empirical correlation models (which require importing or loading external data) to the database.

By making avg_periods optional this refactor allows for "stateless" instantiation. GmpeIndirectAvgSA can be refactored to instantiate the correlation models without avg_periods. In this way the correlation models are used by GenericGmpeAvgSA and GmpeIndirectAvgSA in a similar manner

Changes:

    Modified __init__ to allow avg_periods=None.

    Added a RuntimeError guard in __call__ if the matrix is accessed before being built.

    This change is required to support empirical correlation models currently being implemented (e.g., Clemett et al.) and indirect GMPEs.

Testing:

    Verified that existing models (BakerJayaram, Akkar) still pass hazardlib tests.


nc-hsu:
I have just tried to run some hazard calculations using the new correlation models implemented here and in PR #11168 but the calculations raise a `TypeError ` related to the pickling of a `dict_keys` object. This error can be replicated by running the job in \openquake\qa_tests_data\classical\case_34\job.ini with `corr_func="clemett_asc" `.

How would you like me to fix this? Should i fix it on my forked branch and create a new pull request or should I open an issue because PR #11168 has already been merged?

PR #11197, opened 2026-02-08
Correct Hashash et al. (2020) site term in original NGA East

CB-quakemodel:
@g-weatherill informed us of a small error in the Hashash et al. (2020) site term in the original NGA East GSIM module (`gsim\nga_east.py`) (NB: the implementation within `gsim\mgmpe\hashash2020.py` does not require correction).

In this PR the required correction is made and the associated test tables and QA expected results are regenerated as required.


mmpagani:
Grazie @CB-quakemodel and sorry for missing your note on the other implementation. How big are the differences? the results for EUR for this spectral acceleration are not correct

PR #11200, opened 2026-02-09
Add EMME24 backbone site amp model

CB-quakemodel:
Implement the EMME24 site amplification model described in the files kindly provided by Abdullah Sandıkkaya:

image

Amplifications of ground-motion relative to the reference velocity of 800 m/s were checked against those in the provided files. The test files were regenerated to consider sites with Vs30s of 300 m/s and 900 m/s too.

**Spectra over a range of Vs30 values for some suitable scenarios using `EMME24BB_GMM3SGM2` (central mean and central sigma):**

image

**Comparison of spectra on vs30 of 300 m/s against the ASCR GMMs of which the site models were considered when developing the EMME24 site model:**

ResponseSpectra


openquake/hazardlib/gsim/chiou_youngs_2014.py
```diff
@@ -579,29 +612,39 @@ def get_tau(C, mag):
...
+    else:
+        # Compute EMME24 site term instead (no basin effects considered here)
+        mean = ln_y_ref + get_emme_site_term(C, ctx, emme_coeffs[imt], conf)
 
-    # Get standard deviations
-    sig, tau, phi = get_stddevs(
-        conf['peer'], C, ctx, ctx.mag, y_ref, f_nl_scaling)
+        # Sigma components are determined within EMME backbone's compute method
+        sig, tau, phi =\
+              np.empty_like(mean), np.empty_like(mean), np.empty_like(mean)
```
micheles, line 648: `numpy.empty_like` is potentially dangerous (i.e. if due to a future refactoring the array is not initialized properly it will contain bogus data), it is much better to use `numpy.zeros`

Issue #284, opened 2011-06-14
Users should be able to find/view/reuse existing calculation parameters and results


al-maisan:
Presently there is now way for the web GUI user to find and view the results of OpenQuake calculations performed in the past.

PR #11201, opened 2026-02-09
Add get_x_l_ratio() method to BaseSurface for PFDHA applications

vup1120:
This PR adds support for computing normalized along-strike position (x/L) for Probabilistic Fault Displacement Hazard Assessment (PFDHA) applications.

It includes: 

1.  `_get_surface_trace()`: 
Private helper method that extracts the surface trace (top edge) as a 2D array of geographic coordinates with NaN 
  handling for KiteSurface compatibility.
2. `get_x_l_ratio(mesh)`: It computes the normalized along-strike position (x/L) for each point in the mesh relative to the surface trace
3. Testing suite added to `base_surface_test.py`

Notes:

1. Point-to-polyline projection algorithm finds the nearest trace segment 
  for each site using vectorized distance calculations
2. Returns tuple `(x_over_L, L_km)` where `x_over_L` is a numpy array with 
  values in [0, 1] and `L_km` is the total trace length in kilometers
3. All x/L values are clamped to [0, 1] before return

### DIFF
--- a/openquake/hazardlib/geo/surface/base.py
+++ b/openquake/hazardlib/geo/surface/base.py
@@ -251,7 +251,7 @@ class BaseSurface(metaclass=abc.ABCMeta):
             Tuple (x_over_l, l_km) where x_over_l is numpy array in [0,1]
             and l_km is trace length in km.
         """
-        trace = self._get_top_rupture_trace()
+        trace = self._get_tor()
         site_lons = mesh.lons.flatten()
         site_lats = mesh.lats.flatten()
         n_sites = len(site_lons)
@@ -400,7 +400,7 @@ class BaseSurface(metaclass=abc.ABCMeta):
             dep = numpy.min(top_edge.depths)
             return dep
 
-    def _get_top_rupture_trace(self):
+    def _get_tor(self):
         """
         Extract the top rupture trace (top edge) of the rupture surface as a
         2D array of geographic coordinates.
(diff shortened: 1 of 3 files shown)


openquake/hazardlib/geo/surface/base.py
```diff
@@ -240,6 +240,84 @@ def get_ry0_distance(self, mesh):
...
+        if trace.shape[0] < 2:
+            return numpy.zeros(n_sites), 0.0
+
+        # Compute geodetic trace length
+        seg_lengths_km = geodetic.geodetic_distance(
+            trace[:-1, 0], trace[:-1, 1],
+            trace[1:, 0], trace[1:, 1]
+        )
+        L_km = float(numpy.sum(seg_lengths_km))
+        if L_km <= 0.0:
```
micheles: How can this be negative?

vup1120: I removed this and only keep the if L_proj == 0.0 check ( at #284). It was if L_proj <= 0.0 before, but it actually won't be negative, so I modified it now to "== 0.0"

openquake/hazardlib/geo/surface/base.py
```diff
@@ -322,6 +398,26 @@ def get_top_edge_depth(self):
             dep = numpy.min(top_edge.depths)
             return dep
 
+    def _get_surface_trace(self):
```
mmpagani: I would call this _get_tor(self)

vup1120: All Fixed 🙂

micheles (approved):

PR #11203, opened 2026-02-09
Improved management of geospatial queries and plotting boundaries

ptormene:
I created an infrastructure to build and reuse spatial indices for the mosaic and for country borders (of different administrative levels). If you specify the geodata sources in the openquake config, it reads from there, otherwise it uses the existing geodata in qa_tests_data. The spatial index can be initialized starting from shp, gpkg or parquet (an efficient binary format). Once initialized, an instance of SpatialIndex holds the geodataframe, its geometries and the corresponding STRTree. It offers methods `locate` and `nearby`, to identify the geometry covering a point of given coordinates or those close to those coordinates within a given threshold in degrees. Each spatial index is cached once on initialization and then it can be reused efficiently.

I also added a `geodata_to_parquet` command to convert shp or gpkg into parquet. It is convenient for speedup, reducing disk occupation and avoiding some annoying gdal logs. NOTE: it requires `pip install pyarrow`, that would need to be added to our dependencies. If we prefer to avoid adding a new requirement, we can keep using the existing shp or gpkg files, and the minor consequences happen only at initialization time.

I modified many existing functions to get geometries from the spatial indices created in advance. I also improved the `add_borders` functionality making it more efficient, so now it plots borders only within the current viewport and it redraws them if the x and/or y limits change due to panning/zooming (otherwise, for instance, you would not see much if you zoom out).

I am opening this PR as a draft as a way to share what I've done so far and to check if these changes are causing any regressions. I think most of the work is done, but I added many `FIXME`, for instance where there are functions that should still be replaced using the proposed paradigm. Probably I also need to add some tests, though part of what I've done is already covered by existing ones.

NOTE: we could further optimize things by pickling the spatial indices, therefore adding a speedup at initialization time. However, the creation of the spatial index is already very fast (especially when using the parquet format, but also with shp or gpkg), and I am using as examples the existing mosaic borders and the full geometries provided by geoBoundaries for admin levels 0, 1 and 2.


CB-quakemodel:
Hi @ptormene the original version of  `geolocate` is very useful for me personally but I see you want to remove it eventually. I think keeping this function would be a good idea as it's used currently in GEESE and is useful in general.

ptormene:
> Hi @ptormene the original version of `geolocate` is very useful for me personally but I see you want to remove it completely eventually. I think keeping a version of this function (as I see currently is retained here as `geolocate_with_geom_df`) would be a good idea as it's used currently in GEESE and is useful in general.

I agree the geolocate function is very useful. My aim was to replace it with a more efficient version. Perhaps a better approach would be to keep the original function as it was and to add a new one with a different name, so the existing code using the original function will not break. I will change this PR following this more conservative approach.

Issue #11190, opened 2026-02-05
Remove all wrong occurrences of `registry[gmpe_name](**kw)` in hazardlib


micheles:
There must be only one doing the right thing and setting the `._toml` attribute.

PS: alternatively, one should set `._toml` in MetaGSIM, so that it is always there and no replacements is needed.

PR #11206, opened 2026-02-10
Automagic `gsim._toml` attribute

micheles:
Closes https://github.com/gem/oq-engine/issues/11190 without removing any call to `registry[gmpe_name](**kwargs)`.

Since now every instantiation automatically sets the `._toml` attribute, GSIMs are hashable and can be used as dictionary keys in all circumstances, while before if the user forgot to call `valid.gsim` the `_toml` attribute could have been missing and inherited from the base class as the empty string: all gsims manually instantiated would have had the same hash, resulting in potential disasters.


CB-quakemodel (approved):
Thanks Michele. FYI I just switched to this branch and the SMT tests are all green still with this change

PR #11212, opened 2026-02-12
Replace `geolocate_geometries` with `geolocate_within_buffer`

ptormene:
`geoocate_geometries` was used only in one context, inside `get_close_mosaic_models` used in the impact workflow.
The function `geolocate_geometries` was not as optimized as `geolocate`, though is not a problem in this context querying a single point at a time (geospatial query was already pretty fast). However, it identified all the models intercepting a circular buffer built around the hypocenter, without sorting them by ascending distance from the site.
With this change I am returning the list of models sorted by distance, so by default the UI will pre-select the closest model. It is still not optimized, but it works fast in the context we are using it.

### DIFF
--- a/openquake/commonlib/calc.py
+++ b/openquake/commonlib/calc.py
@@ -492,7 +492,7 @@ def get_close_mosaic_models(lon, lat, buffer_radius):
         centered on the given coordinates having the specified radius
     """
     mosaic_df = readinput.read_mosaic_df()
-    close_mosaic_models = geo.utils.geolocate_within_dist(
+    close_mosaic_models = geo.utils.geolocate_within_buffer(
         lon, lat, buffer_radius, mosaic_df)
     if not close_mosaic_models:
         raise ValueError(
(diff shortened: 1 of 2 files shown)


openquake/commonlib/calc.py
```diff
@@ -493,10 +492,8 @@ def get_close_mosaic_models(lon, lat, buffer_radius):
         centered on the given coordinates having the specified radius
     """
     mosaic_df = readinput.read_mosaic_df()
-    hypocenter = Point(lon, lat)
-    hypo_buffer = hypocenter.buffer(buffer_radius)
-    geoms = numpy.array([hypo_buffer])
-    [close_mosaic_models] = geo.utils.geolocate_geometries(geoms, mosaic_df)
+    close_mosaic_models = geo.utils.geolocate_within_dist(
```
micheles: Maybe a name like `geolocate_within_buffer` would be better?

PR #11231, opened 2026-02-19
FDHA: Youngs et al. (2003) primary surface displacement (AD/MD) and tests

vup1120:
Adds the primary surface displacement package under openquake.fdha.primary_surf_displ, implementing the Youngs et al. (2003) model for conditional exceedance probability of primary fault displacement, plus tests and changelog.

### DIFF
--- a/openquake/calculators/classical.py
+++ b/openquake/calculators/classical.py
@@ -216,22 +216,23 @@ def classical(grp_keys, tilegetter, cmaker, dstore, monitor):
     elif len(grps) == 1 and len(grps[0]) >= 2:
         # tested in case_25
         b0, *blks = _split_src(list(grps[0]), 4)
-        rest = sum(blks, [])
         t0 = time.time()
         res = baseclassical(b0, sites, cmaker, True)
         dt = time.time() - t0
         yield res
         if dt > 2 * cmaker.oq.split_time:
-            for b in _split_src(rest, 3):
-                yield baseclassical, b, tilegetter, cmaker, True, dstore
+            for blk in blks[1:]:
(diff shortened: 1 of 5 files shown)


openquake/fdha/primary_surf_displ/base.py
```diff
@@ -0,0 +1,71 @@
...
+        exceeds *d* metres, given magnitude and position along the rupture.
+
+        :param d:
+            Target displacement(s) in metres, scalar or array-like.
+        :param x_l:
+            Relative along-strike position x/L (0 = end, 0.5 = centre),
+            scalar or array-like.
+        :param mag:
+            Earthquake moment magnitude (scalar).
+        :param style:
```
mmpagani: I understood we wanted to use rake here. See https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/valid.py#L1219 . For the youngs et al. model you need just a small utility function that takes the rake and defines the values of the `style` parameter accordingly.

mmpagani (approved):
Apart from the comment on the rake, the PR looks good to me.

openquake/fdha/primary_surf_displ/base.py
vup1120: Hello Marco!

1. get_prob(...) now takes rake (instead of style) in the Youngs primary displacement model.
2. Added shared utility openquake.fdha.utils.rake_to_style(rake) to map rake to:
normal for normal-faulting rake range,
otherwise strike_slip / reverse / all as applicable,
and "undefined" -> "all".

So the model now uses rake-driven style selection as requested, rather than requiring style to be passed directly.

PR #11235, opened 2026-02-21
Fixed event_based from SES and stored relevant_ruptures and relevant_events

micheles:
Instead of the requested sites the GMFs were computed on the original sites :-(


micheles:
There is some other bug I am still working on :-/

PR #11263, opened 2026-03-02
Add PGA to AvgSA computation

nc-hsu:
This change allows the PGA to be considered in the calculation of AvgSA.

It is assumed that the same GMPE used to calculate the SA values can also be used to calculate the PGA.

Empirical correlation models, such as `ClemettCorrelationModelAsc` etc. will work with PGA so long as they have been defined for a period of T = 0.0s. Otherwise and error will be raised saying the period is out of range.

### DIFF
--- a/openquake/hazardlib/tests/gsim/mgmpe/correlation_models_test.py
+++ b/openquake/hazardlib/tests/gsim/mgmpe/correlation_models_test.py
@@ -332,7 +332,7 @@ class ClemettCorrelationModelSSlabTestCase(unittest.TestCase):
 class ClemettCorrelationModelVranceaTestCase(unittest.TestCase):
     def setUp(self):
         self.corr_model = ClemettCorrelationModelVrancea(np.array([1, 1.1, 2.0]))
-    # TODO
+    
     def test_total_correlations(self):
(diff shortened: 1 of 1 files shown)


openquake/hazardlib/tests/gsim/mgmpe/correlation_models_test.py
```diff
@@ -314,40 +314,40 @@ def test_total_correlations(self):
...
         actual = self.corr_model.get_within_event_correlation(0.06, 0.025)
-        np.testing.assert_almost_equal(actual, 0.954)
+        np.testing.assert_almost_equal(actual, 0.9525)
 
 
 class ClemettCorrelationModelVranceaTestCase(unittest.TestCase):
     def setUp(self):
         self.corr_model = ClemettCorrelationModelVrancea(np.array([1, 1.1, 2.0]))
-
+    # TODO
```
CB-quakemodel: Please remove this comment or expand what is still to be done here

nc-hsu: This is an erroneous comment that I forgot to delete. I have removed it now

PR #11282, opened 2026-03-08
Support sampling from the truncated multivariate normal distribution

raoanirudh:
Currently all `scenario` / `event_based` jobs employing a cross-IM correlation model completely ignore the `truncation_level` parameter, with warnings like ```The `truncation_level` param is ignored with GodaAtkinson2009```, leading to the possibility of larger (or lower) than desired epsilon values being sampled.

Though scipy.stats does provide a truncated normal distribution sampler ([scipy.stats.truncnorm](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.truncnorm.html)), it does not provide a similar truncated multivariate normal distribution sampler. Thus, we make an attempt to use Paul Brunzema's Python reimplementation[^1] (https://github.com/brunzema/truncated-mvn-sampler) of the Botev (2017)[^2] algorithm for simulation and independent and identically distributed sampling of the truncated multivariate normal distribution.

A second critical change introduced in this PR is allowing the user to specify different truncation levels for between-event residuals and within-event residuals, allowing for a finer-grained control on the desired limits on the two types of epsilons sampled in a scenario or event-based calculation. Two new job parameters are introduced: `truncation_level_between` and `truncation_level_within`. The previous behavior with the user specifying only `truncation_level` is maintained as is. The new parameters are used only in scenario and event-based calculations.

### References
[^1]: Brunzema, Paul (2021). Efficient sampling from the truncated MVN. GitHub repository. https://github.com/brunzema/truncated-mvn-sampler. MIT License.
[^2]: Botev, Zdravko I. (2017). The normal law under linear restrictions: simulation and estimation via minimax tilting. Journal of the Royal Statistical Society Series B: Statistical Methodology, 79(1), 125-148. https://doi.org/10.1111/rssb.12162

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,13 @@
+  [Anirudh Rao]
+  * Support sampling from the truncated multivariate normal distribution
+    using Paul Brunzema's implementation of the Botev (2017) algorithm
+    (https://github.com/brunzema/truncated-mvn-sampler), permitting
+    truncation in calculations employing a cross-IM correlation model
+  * Introduce two new job parameters: `truncation_level_between` and
+    `truncation_level_within` allowing the user to specify different
+    truncation levels for between-event residuals and within-event
+    residuals in a scenario or event-based calculation
+
   [Christopher Brooks]
   * Add ability to specify deltas for tau and phi within ModifiableGMPE.
 


openquake/hazardlib/contexts.py
```diff
@@ -595,6 +595,12 @@ def _init1(self, param):
         self.ses_seed = param.get('ses_seed', 42)
         self.ses_per_logic_tree_path = param.get('ses_per_logic_tree_path', 1)
         self.truncation_level = param.get('truncation_level', 99.)
+        self.truncation_level_between = param.get('truncation_level_between')
```
micheles, line 599: You can rewrite this as a single line
```python
self.truncation_level_between = param.get('truncation_level_between', self.truncation_level)
```
Same for `self.truncation_level_within`

openquake/hazardlib/cross_correlation.py
```diff
@@ -117,6 +116,34 @@ def get_correlation(self, from_imt: IMT, to_imt: IMT) -> float:
...
+            samp = TruncatedMVN(mu, corr, -bounds, bounds, seed=seed).sample(
+                num_events)
+            if np.isfinite(samp).all():
+                return samp
+        except RuntimeError as err:
+            if 'not positive semi-definite' not in str(err):
+                raise
+        # Use the existing PSD regularization available in conditioned_gmfs.py
+        # also in case TruncatedMVN returns non-finite samples.
+        from openquake.hazardlib.calc.conditioned_gmfs import corr_clipped
```
micheles: Is there a reason why the import is not at top level?

raoanirudh: ```openquake.hazardlib.calc.conditioned_gmfs``` itself imports `cross_correlation`, so it triggers a circular import error if I import from `conditioned_gmfs` at the top level. Can this be done in a more elegant manner?

openquake/hazardlib/contexts.py
raoanirudh: Done!

openquake/hazardlib/cross_correlation.py
micheles: In case of circular imports one needs to put the code to be imported in a third, independent, module

raoanirudh: Matrix helper functions refactored to an independent module ```openquake/hazardlib/correlation_utils.py```. Both `conditioned_gmfs` and `conditioned_gmfs` now import the required functions from `correlation_utils`, avoiding circular import situations.

doc/user-guide/configuration-file/event-based-psha-config.rst
```diff
@@ -34,6 +34,24 @@ model to be used for the generation of ground motion fields.::
...
+Truncation levels
+*****************
+
+For event-based PSHA calculations, truncation of residual sampling in GMFs can
+be controlled with:
+
+- ``truncation_level_between``: truncation level for between-event residuals.
+- ``truncation_level_within``: truncation level for within-event residuals.
+
+For backward compatibility, if only ``truncation_level`` is specified, the
```
mmpagani: I would replace `For backward compatibility` with `Otherwise`

doc/user-guide/configuration-file/scenario-hazard-config.rst
```diff
@@ -52,6 +52,24 @@ The parameters unique to the scenario calculator are described below:
...
+Truncation levels
+*****************
+
+For scenario hazard calculations, truncation of the residual sampling can be
+controlled separately for between-event and within-event residuals:
+
+- ``truncation_level_between``: truncation level for between-event residuals.
+- ``truncation_level_within``: truncation level for within-event residuals.
+
+For backward compatibility, if only ``truncation_level`` is specified, the
```
mmpagani: As above

openquake/commonlib/oqvalidation.py
```diff
@@ -1267,11 +1281,71 @@ def fix_legacy_names(self, dic):
             inp['site_model'] = [inp.pop('sites')]
             self.prefer_global_site_params = True
 
+    def _set_truncation_levels(self):
+        # Backward-compatible truncation levels:
+        # - legacy: truncation_level applies to both components
```
mmpagani: As far as I am aware, there is no scientific work suggesting an asymmetric truncation. I would not recommend giving the impression that we now consider this strategy our default.

openquake/commonlib/oqvalidation.py
```diff
@@ -1267,11 +1281,71 @@ def fix_legacy_names(self, dic):
...
+        # - new: truncation_level_between/within can be set independently
+        tl = getattr(self, 'truncation_level', None)
```
mmpagani, line 1289: I am assuming there is already a check somewhere making sure we have at least the `truncation_level` defined

openquake/commonlib/oqvalidation.py
```diff
@@ -1267,11 +1281,71 @@ def fix_legacy_names(self, dic):
...
+                tlb = tlw
+            if tlw is None:
+                tlw = tlb
+            tl = tlw
+        if tl is not None:
+            self.truncation_level = tl
+        self.truncation_level_between = tlb
+        self.truncation_level_within = tlw
+
+    def _set_hazard_imtls(self, names_vals):
```
mmpagani, line 1297: Is this part of the code actually changed?

openquake/hazardlib/calc/gmf.py
```diff
@@ -388,17 +399,27 @@ def _compute(self, mean_stds, m, imt, gsim, intra_eps, idxs, rng=None):
...
-                    check_valid="raise", tol=1e-5, method="cholesky")
-                gmf = exp(arr, imt != "MMI").T
+                std = numpy.sqrt(numpy.maximum(numpy.diag(cov_Y_Y), eps))
+                corr = cov_Y_Y / std[:, None] / std[None, :]
+                corr = (corr + corr.T) / 2.0
+                numpy.fill_diagonal(corr, 1.0)
+                tl = max(tlw, tlb)
+                bounds = numpy.full(len(corr), tl)
+                seed = int(rng.integers(0, numpy.iinfo(numpy.int32).max))
+                eps_y = TruncatedMVN(
```
mmpagani: You are using TruncatedMVN by default, even when there is no asymmetry in the truncation. Out of curiosity, did you check what the increase is in computational cost?

openquake/hazardlib/truncated_mvn.py
```diff
@@ -0,0 +1,446 @@
...
+        c = self.L @ x
+        lt = self.lb - mu - c
+        ut = self.ub - mu - c
+        p = np.sum(lnNormalProb(lt, ut) + 0.5 * mu ** 2 - x * mu)
+        return p
+
+    def get_gradient_function(self):
+        # wrapper to avoid dependancy on self
+
+        def gradpsi(y, L, l, u):
```
mmpagani, line 332: This is probably 'as is' from Brunzema, but I do not understand why this was not simply put in an external function.

openquake/qa_tests_data/event_based/case_34/expected/avg_gmf.csv
```diff
@@ -1,39 +1,39 @@
...
-eydkp7hp,-7.76179E+00,3.72834E+01,6.53830E-01,2.02128E+00,1.31872E-02,1.69646E+00,1.00000E-10,1.00000E+00
-eydkn2g1,-7.80696E+00,3.72700E+01,6.08160E-01,2.09305E+00,1.18161E-02,1.37450E+00,1.00000E-10,1.00000E+00
-eydsf735,-7.63406E+00,3.74158E+01,6.16823E-01,2.43695E+00,1.22959E-02,1.36871E+00,1.00000E-10,1.00000E+00
-eyfctpyd,-7.15397E+00,3.82755E+01,3.88225E-01,1.54950E+00,1.00019E-02,1.00000E+00,1.00000E-10,1.00000E+00
-eycdgtmd,-8.97915E+00,3.84815E+01,3.36188E-01,1.25681E+00,1.00212E-02,1.01072E+00,1.00000E-10,1.00000E+00
-eyf4ftze,-8.31731E+00,3.84844E+01,3.47489E-01,1.36684E+00,1.00824E-02,1.04350E+00,1.00000E-10,1.00000E+00
-eycmvyw2,-9.23082E+00,3.90153E+01,3.08501E-01,1.15626E+00,1.00000E-02,1.00000E+00,1.00000E-10,1.00000E+00
-eyczqm6b,-8.51056E+00,3.92721E+01,3.03671E-01,1.06524E+00,1.00000E-02,1.00000E+00,1.00000E-10,1.00000E+00
-ez402g0r,-8.40400E+00,3.94368E+01,3.07488E-01,1.13667E+00,1.00000E-02,1.00000E+00,1.00000E-10,1.00000E+00
+eyd51dp4,-8.36172E+00,3.71013E+01,7.31678E-01,2.40018E+00,1.17271E-02,1.38746E+00,1.00000E-10,1.00000E+00
```
mmpagani, line 3: The change in some of the values is non-negligible. Is this because we were doing something wrong before, or did you change the test to check the asymmetric truncation case?

doc/user-guide/configuration-file/event-based-psha-config.rst
raoanirudh: Done

openquake/commonlib/oqvalidation.py
raoanirudh: Agreed, and rephrased the function comment

doc/user-guide/configuration-file/scenario-hazard-config.rst
raoanirudh: Changed as above.

openquake/commonlib/oqvalidation.py
raoanirudh: In contexts.py, the truncation_level is set based on the value in the job file, or to 99 if not provided. Replicating the same below.

openquake/commonlib/oqvalidation.py
raoanirudh: No the code is unchanged, but I refactored it out into a separate function because @micheles has a test forbidding functions over 90 lines and the __init__() function was starting to exceed 90 lines. Refactoring includes the following:

- Extract truncation compatibility logic into _set_truncation_levels.
- Extract IMT/hazard level setup into _set_hazard_imtls.
- Extract minimum-intensity normalization into _normalize_minimum_intensity.

To avoid running into the forbid\_long\_funcs error:
```
openquake/baselib/tests/flake8_test.py::test_forbid_long_funcs - RuntimeError: [('openquake.commonlib.oqvalidation.__init__', 93)]
```

openquake/hazardlib/calc/gmf.py
raoanirudh: This code block is handling the gmf sampling for the conditioned gmf calculator, in which cross-IM correlation is always considered for calculations involving more than one IMT. Previously we were simply unable to truncate at all for this case.

openquake/hazardlib/truncated_mvn.py
raoanirudh: Yes, for now the entire truncated_mvn.py script is ported as-is from https://github.com/brunzema/truncated-mvn-sampler/blob/main/minimax_tilting_sampler.py. I've already begun refactoring some of the other helper functions already used by the correlation modules and conditioned gmf module into correlation_utils.py, and these helper functions from Paul Brunzema's script could also be moved there. But since the Brunzema script uses a different license, it might be better to do that in a subsequent pull request with due consideration of the license implications.

openquake/qa_tests_data/event_based/case_34/expected/avg_gmf.csv
raoanirudh: In all of these cases, the job files include a truncation_level parameter, which was being ignored previously.

openquake/hazardlib/calc/gmf.py
```diff
@@ -329,6 +332,18 @@ def strip_zeros(self, data):
...
+    def get_symmetric_bounds(cov_matrix, level):
+        """
+        Calculates the lower and upper bound vectors for symmetric truncation
+        based on the marginal standard deviations of the covariance matrix.
+        """
+        # Extract marginal standard deviations from the diagonal
+        sigmas = numpy.sqrt(numpy.diag(cov_matrix))
+        upper = level * sigmas
+        lower = -level * sigmas
+        return lower, upper
```
micheles: You can just return `-upper, upper` and avoid a multiplication (not that it will make a signicant difference in speed, but it looks cleaner to me)

openquake/hazardlib/calc/gmf.py
```diff
@@ -354,15 +369,23 @@ def compute_all(self, mean_stds=None, max_iml=None,
...
-                                 for _ in range(self.M)]
-                    self.eps[idxs] = self.cross_correl.get_inter_eps(
-                        self.imts, E, rng).T
+                    if tlw <= 1E-9:
+                        intra_eps = [numpy.zeros((self.N, E), F32)
+                                     for _ in range(self.M)]
+                    else:
+                        intra_eps = [self.within_dist.rvs((self.N, E), rng)
+                                     for _ in range(self.M)]
+                    if tlb <= 1E-9:
```
micheles: Instead of sprinkling the code with `1E-9` you could introduce a constant `MINIMUM_TRUNCATION=1E-9`

openquake/hazardlib/cross_correlation.py
```diff
@@ -117,6 +117,32 @@ def get_correlation(self, from_imt: IMT, to_imt: IMT) -> float:
...
+        try:
+            samp = TruncatedMVN(mu, corr, -bounds, bounds, seed=seed).sample(
+                num_events)
+            if np.isfinite(samp).all():
+                return samp
+        except RuntimeError as err:
+            if 'not positive semi-definite' not in str(err):
+                raise
+        # Use the existing PSD regularization available in correlation_utils.py
+        # also in case TruncatedMVN returns non-finite samples.
```
micheles, line 141: What does it mean `non-finite samples`?

openquake/hazardlib/truncated_mvn.py
```diff
@@ -0,0 +1,446 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+import numpy as np
+import math
+from scipy import special, optimize
+
+EPS = 10e-15
+
```
micheles, line 31: Did you mean `1E-15` or `1E-14`?

raoanirudh: `10e-15` or `1e-14`. This part is directly from Paul Brunzema's script.

openquake/hazardlib/calc/gmf.py
raoanirudh: Done

openquake/hazardlib/cross_correlation.py
raoanirudh: Refers to the case where **TruncatedMVN** returns `inf` or `NaN`, as tested by [np.isfinite](https://numpy.org/doc/stable/reference/generated/numpy.isfinite.html)

openquake/hazardlib/calc/gmf.py
raoanirudh: Done, using `TRUNCATION_LEVEL_THRESHOLD = 1E-9`

PR #11307, opened 2026-03-16
Fix SiMidorikawa1999 to use Rhypo instead of Rrup for NE correction term

CB-quakemodel:
Equation 3.5.2-1 of "Technical Reports on National Seismic Hazard Maps for Japan" uses Rhypo for the northeast correction, but the current OQ implementation uses Rrup.

I update the test files to include `dist_rhypo` as is required for the NE correction GSIM classes - the differences in expected values for the test scenarios are very small so still pass with the current 0.1 % diff tolerance.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,8 @@
+  [Christopher Brooks]
+  * Correct the `_apply_subduction_trench_correction` term in 
+    Si and Midorikawa (1999)'s `NorthEastCorrection` subclasses
+    to use `rhypo` instead of `rrup`.
+
   [Anirudh Rao]
   * Support sampling from the truncated multivariate normal distribution
     using Paul Brunzema's implementation of the Botev (2017) algorithm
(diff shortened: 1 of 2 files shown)


openquake/hazardlib/tests/gsim/data/SM99/SM99SInterNECorr_MEAN.csv
```diff
@@ -1,97 +1,97 @@
...
-8.5,50,100,144.027324,37.699782,400,MEAN,47.9550765631
-8.5,50,0,144.936566,37.694545,400,MEAN,231.955224838
-8.5,50,25,144.936566,37.694545,400,MEAN,126.66565449
-8.5,50,50,144.936566,37.694545,400,MEAN,81.3750938469
-8.5,50,100,144.936566,37.694545,400,MEAN,41.4790764881
-8.5,50,0,148.342925,37.6128,400,MEAN,134.629880487
-8.5,50,25,148.342925,37.6128,400,MEAN,73.5184212282
-8.5,50,50,148.342925,37.6128,400,MEAN,47.2311807887
-8.5,50,100,148.342925,37.6128,400,MEAN,24.075004623
+,rup_mag,rup_hypo_depth,dist_rrup,site_lon,site_lat,site_vs30,result_type,pgv,dist_rhypo
```
raoanirudh: Check index column

PR #11310, opened 2026-03-17
If any worker dies, log the event and kill the master and its children

ptormene:
We are also avoiding having deceased workers after a job launched from the WebUI completes. Those deceased workers were not removed until the WebUI was stopped.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,8 @@
+  [Paolo Tormene, Michele Simionato, Paul Henshaw]
+  * If any worker dies before completion, the corresponding master and
+    all its subprocesses are killed and the event causing the
+    interruption is logged
+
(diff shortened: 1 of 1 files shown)


pslh (approved):
LGTM

openquake/server/apps.py
```diff
@@ -17,13 +17,35 @@
...
+import logging
 from django.apps import AppConfig
 from django.conf import settings
 from sqlite3 import OperationalError
 from openquake.baselib import config
 from openquake.server import dbserver, db
 
 oqdir = os.path.dirname(os.path.dirname(__file__))
+logger = logging.getLogger(__name__)
+
```
micheles: Just import reap_children from engine.py, no?

PR #11312, opened 2026-03-17
Missing backarc requirement for GMMs based on AbrahamsonEtAl2015

claudio525:
These models require the backarc flag as AbrahamsonEtAl2015 requires it, but its not specified in the REQUIRES_SITES_PARAMETERS, this PR simply adds the missing backarc requirement.

### DIFF
--- a/openquake/hazardlib/gsim/eshm20_bchydro.py
+++ b/openquake/hazardlib/gsim/eshm20_bchydro.py
@@ -202,7 +202,7 @@ class BCHydroESHM20SInter(AbrahamsonEtAl2015SInter):
     """
 
     # Requires Vs30 and distance to the volcanic front
-    REQUIRES_SITES_PARAMETERS = {'vs30', 'xvf'}
+    REQUIRES_SITES_PARAMETERS = {'vs30', 'xvf', 'backarc'}
 
     COEFFS = CoeffsTable(sa_damping=5, table="""\
     imt          vlin        b   theta1    theta2        theta6    theta7    theta8  theta10  theta11   theta12   theta13   theta14  theta15   theta16      phi     tau   sigma  sigma_ss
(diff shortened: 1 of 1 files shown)


claudio525:
Okay, so I probably should have provided some additional context. 
After upgrading to OQ 3.25 some of our tests were failing with the error:
`AttributeError: recarray has no attribute backarc` in this line (https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/gsim/abrahamson_2015.py#L128).
This left me somewhat confused, given that in our tests we just set backarc to False in the context. 
After some digging, I identified that the provided backarc values don't make it into the recarray (https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/contexts.py#L765), as `self.defaultdict` in the ContextMaker is only populated with the required site parameters from the GMM, not its parents, i.e., `backarc` is not inherited from Abrahamson (2015) (@CB-quakemodel ). 
My conclusion was therefore that `backarc` should be in the REQUIRED_SITE_PARAMETERS for the BCH16 models, hence I created this PR.

However, I did some digging based on @g-weatherill comments, and the real reason why our tests started breaking is the change in this PR (https://github.com/gem/oq-engine/pull/11216) that sets the default value of the `faba_taper_model` from `Step` to `None` in the Abrahamson 2015 model. Given that in our code we didn't explicitly set `faba_taper_model="Step"` we then ran into this line (https://github.com/claudio525/oq-engine/blob/master/openquake/hazardlib/gsim/abrahamson_2015.py#L128), resulting in the above error message.
I have updated our code to explicitly set `faba_taper_model="Step"` for the BCH16 models, and this has fixed the issue.

However, I do think the fact that the BCH16 models default to `faba_taper_model=None` is confusing, as the `backarc` term is required if `faba_taper_model=None`. So either `backarc` should be a required site parameter for the BCH16 models, or they shouldn't default to `faba_taper_model=None`. Based on @g-weatherill 's comment I assume the latter is the correct approach. So my suggestion would be to simply add a `__init__` to the BCH16 models that either requires the `faba_taper_model` argument, or make it an optional argument and set it to `"Step"` by default.

PR #11315, opened 2026-03-18
Tried to remove setup.py

micheles:
Towards a single file for package management

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@
 import os
 import re
 import sys
-from setuptools import setup, find_namespace_packages
+from setuptools import setup
 
 def generate_data_files(source_dir):
     # Helper function for demos


pyproject.toml
```diff
@@ -2,6 +2,65 @@
...
+]
+
+# Core dependencies (simplified based on common engine requirements)
+# Note: oq-engine uses requirements-py3x.txt files for exact pinning
+dependencies = [
+    "numpy>=1.20",
+    "scipy>=1.7",
+    "pandas>=1.3",
+    "h5py>=3.0",
+    "django>=4.2",
```
vot4anto: We need to have upper limit, django 5 is huge change from 4

pyproject.toml
```diff
@@ -1,7 +1,74 @@
 [build-system]
-requires = ["setuptools >= 69.0"]
+requires = ["setuptools >= 69.0", "wheel"]
```
micheles: Should be "setuptools >= 79.0" ?

setup.py
```diff
@@ -21,10 +21,6 @@
 import sys
 from setuptools import setup, find_namespace_packages
```
micheles: `find_namespace_packages` looks useless now

pyproject.toml
```diff
@@ -1,7 +1,74 @@
...
+    "qa_tests", "qa_tests.*",
+    "tools",
+    "*.*.tests", "*.*.tests.*",
+    "openquake.engine.bin",
+    "openquake.engine.bin.*"
+]
+
+# If you use specific dev tools like pytest or black
+[tool.pytest.ini_options]
+testpaths = ["openquake/qa_tests"]
```
micheles: This is confusing (the tests are not in "openquake/qa_tests") and should be removed

pyproject.toml
vot4anto: Yes

pyproject.toml
vot4anto: I do not touch your first template in this section, agree to remove

PR #11320, opened 2026-03-19
Fix check on email notifications for impact and aelo tests

ptormene:
We had a couple of FIXME because the previous implementation incorrectly expected that django overridden settings would be propagated to the subprocesses running aelo or impact jobs.
Now I am explicitly passing that information to the child processes, so we can be sure that tests will actually check the emails corresponding to the jobs and not, for instance, old emails corresponding to previous failed jobs.

### DIFF
--- a/openquake/server/views.py
+++ b/openquake/server/views.py
@@ -16,6 +16,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with OpenQuake. If not, see <http://www.gnu.org/licenses/>.
 
+import sys
 import ast
 import csv
 import shutil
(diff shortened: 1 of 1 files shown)


openquake/server/views.py
```diff
@@ -1463,11 +1482,19 @@ def aelo_run(request):
...
         lon, lat, vs30, params['siteid'], description, asce_version,
         site_class, jobctx, job_owner_email, outputs_uri_web,
-        config.directory.mosaic_dir, aelo_callback)).start()
+        config.directory.mosaic_dir, aelo_callback, email_file_path)
+
+    # hack for debugging the tests
+    # import sys
+    # if 'pytest' in sys.argv[0]:
+    #     aelo.main(*args)
+
```
micheles: I still want the hack working, ideally in conjunction with OQ_DISTRIBUTE="no"

openquake/server/views.py
```diff
@@ -1221,14 +1238,14 @@ def create_impact_job(request, params):
             % (outputs_uri_api, traceback_uri))
 
     args = ([params], [jobctx], job_owner_email, outputs_uri_web,
-            impact_callback)
+            impact_callback, email_file_path)
+
+    # hack for debugging the tests
+    # import sys
     # if 'pytest' in sys.argv[0]:
```
micheles: I still want the hack working, ideally in conjunction with OQ_DISTRIBUTE="no"

PR #11321, opened 2026-03-19
Refactor the BCHydro GSIM to handle ESHM20 FABA models better

CB-quakemodel:
https://github.com/gem/oq-engine/pull/11312 describes an issue where no ESHM20 FABA model is specified within an ESHM20 subclass of the BCHydro GSIM, which results in a missing attribute error when the GSIM incorrectly diverts to the BCHydro backarc term, rather than just skipping the ESHM20 FABA term as was intended in such instances.

This bug was introduced in https://github.com/gem/oq-engine/pull/11216. Prior to https://github.com/gem/oq-engine/pull/11216 the default FABA model was `Step`, which implicitly safeguarded against the reported bug of `xvf` being missing because the FABA model could not be `None` and therefore it was not possible to incorrectly fallback to the BCHydro backarc term (which requires the missing backarc site parameter).

This PR resolves this issue by providing separate functions for the default backarc term of the original BCHydro model (uses the backarc param in the site model), and the ESHM20 FABA term (uses the `xvf` parameter) along with some more explicit safeguards on use of the ESHM20 FABA models.

**Important:** This PR intentionally does not make it possible to use the original BCHydro backarc term in the ESHM20 subclasses as an (implicit) fallback if no FABA model is specified. Permitting a fallback in the case of an ESHM20 subclass with no FABA term to the original BCHydro backarc term was not intended in the original implementation and therefore this safeguard is restored here in a more explicit manner. The option of specifying no FABA model is for 1) turning off the FABA term in the ESHM20 subclasses for some flexibility and 2) ensuring that the default ESHM20 FABA model is None (which makes sense given it's only used in the ESHM20 subclasses - the Step taper being the default for BCHydro would be ambiguous given it's not used in the original GMM).

--------------------------------------------------------

A QA test (classical/case_61) is expanded to consider this scenario (no FABA model explicitly specified).

The changes here for simplification of the BCHydro GSIM in general are propagated accordingly throughout the rest of the GSIM library + some small cleanup.

### DIFF
--- a/openquake/hazardlib/gsim/eshm20_bchydro.py
+++ b/openquake/hazardlib/gsim/eshm20_bchydro.py
@@ -304,14 +304,6 @@ class BCHydroESHM20SSlab(AbrahamsonEtAl2015SSlab):
     """
     ESHM20 adjustment of the BC Hydro GMPE for subduction in-slab events with
     theta6 calibrated to Mediterranean data.
-
-    Introduces two configurable parameters:
-
-    a6_adjustment - the amount to increase or decrease the theta6 (should be
-    +0.0015 (for slower attenuation) and -0.0015 (for faster attenuation)
-
-    sigma_mu_epsilon - number of standard deviations above or below the mean
-    to apply the statistical uncertainty sigma_mu term.
     """
     kind = "eshm20"
 


micheles:
Please update the changelog

PR #11328, opened 2026-03-23
Add a capability to ModifiableGMPE to apply IMT-dependent delta to total, tau or phi

CB-quakemodel:
We already had the ability to specify an IMT-constant scalar to any of these GMM sigma components. This PR adds the ability to specify this on an IMT-dependent basis similar to the existing capabilities for scaling the median or specifying a fixed total sigma per IMT.

Classical/case_94 is expanded to test these capabilities. Also a unit test added.

Also added a test for tabular GMM and a non-tabular GMM both with extra parameters within a mixture model (classical/case_81) for documentation purposes.

### DIFF
--- a/openquake/calculators/export/hazard.py
+++ b/openquake/calculators/export/hazard.py
@@ -486,14 +486,16 @@ def export_gmf_data_csv(ekey, dstore):
 def export_site_model_csv(ekey, dstore):
     sitecol = dstore['sitecol']
     if os.environ.get('OQ_APPLICATION_MODE') == 'AELO':
-        core_params = ('custom_site_id', 'site_id', 'sids', 'lat', 'lon', 'depth',
-                       'vs30', 'vs30measured', 'z1pt0', 'z2pt5')
-        keep = [name for name in sitecol.array.dtype.names if name in core_params]
+        core_params = ('custom_site_id', 'site_id', 'sids', 'lat', 'lon',
+                       'depth', 'vs30', 'vs30measured', 'z1pt0', 'z2pt5')
+        keep = [name for name in sitecol.array.dtype.names
+                if name in core_params]
         arr = sitecol.array[keep]
     else:
         arr = sitecol.array
     fname = dstore.build_fname(ekey[0], '', ekey[1])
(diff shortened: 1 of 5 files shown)


openquake/hazardlib/gsim/mgmpe/modifiable_gmpe.py
```diff
@@ -437,6 +447,69 @@ def add_delta_std_to_phi_std(ctx, imt, me, si, ta, ph, delta):
...
+    ta[:] = new_tau
+
+    # Need to adjust total sig too after adjusting tau for given IMT
+    si[:] = np.sqrt(ta**2 + ph**2)
+
+
+def add_delta_to_phi_std_vector(ctx, imt, me, si, ta, ph, delta):
+    """
+    :param delta:
+        An IMT-dependent vector of deltas to be applied to phi for
```
micheles: Not sure why you say vector when this is a dictionary keyed by the IMT

openquake/qa_tests_data/classical/case_94/gmcLT.xml
```diff
@@ -4,31 +4,64 @@
   <logicTreeBranchSet uncertaintyType="gmpeModel" branchSetID="bs1"
```
micheles, line 4: I don't see why you have an extra dictionary. I mean, instead of passing a dictionary `{'delta': {imt: value}}` you could just pass `{imt: value}`, then  instead of the syntax `add_delta_to_total_std_scalar.delta = ` you would just have  `add_delta_to_total_std_scalar =`. There is no need to repeat the `delta`, right?

CB-quakemodel: I think this dictionary structure is currently necessary to how ModifiableGMPE works given it needs to read `add_delta_to_total_std_vector` as the adjustment/method to apply, and then `delta` and everything within this key is the dictionary of IMT-dependent adjustments to use within `add_delta_to_total_std_vector`.

The approach is consistent with other IMT-dependent adjustments e.g. https://github.com/gem/oq-engine/blob/master/openquake/qa_tests_data/classical/case_54/gsim_logic_tree_modifiable_vector.xml#L13

PR #11359, opened 2026-04-09
Introduce extensions to allow engine to work with oq-geoviewer and qgis-server

nastasi-oq:
REQUIREMENTS:
- access to cloned instance of private PAPERS repository
- look at `local_settings.py.papers` template to configure properly some variable

With this PR we introduce `run_scenario_calc_from_ses_rupture/(?P<rup_id>[0-9]+)` API entry-point to be able to create a calculation starting from an hypothetical rupture defined in `rups_eur_branch_faults.hdf5` file in the `PAPERS` repository.


openquake/server/papers/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with OpenQuake. If not, see <http://www.gnu.org/licenses/>.
+
+from .module import run_scenario_calc_from_ses_rupture
+
+__all__ = ['run_scenario_calc_from_ses_rupture' ]
```
micheles: What's the point of having this module? You can just import `run_scenario_calc_from_ses_rupture` directly and it would be more explicit.

openquake/server/v1/calc_urls.py
```diff
@@ -80,3 +80,10 @@
         re_path(r'^validate_ini$', views.validate_ini),
         re_path(r'^validate_zip$', views.validate_zip),
     ])
+
```
micheles, line 83: The idea is to add the entrypoints below only in PAPERS_MODE, right? Then Paolo has to add it, similarly to the other application modes. However it can go in another PR.

nastasi-oq: Not yet, I prefer to avoid introduce a new MODE until we dacouple modes and features (that seems to me something that can became an hell very very (very) quickly).

micheles:
LGTM once you fix the conflict

openquake/server/papers/__init__.py
nastasi-oq: Fixed.

PR #11360, opened 2026-04-09
Add Alternative Characteristic MFD class

CB-quakemodel:
Alternative characteristic (AC) MFD is effectively back-to-back double truncated exponential distributions. The first zone covering the smaller earthquakes is the GR zone, and the second covering the larger earthquakes is the AC zone. The two zones are linked by the gamma value which is the fraction of the total seismic moment rate assigned to the AC zone according to:

{8F734726-B45E-4EE2-BB54-EE9E07A12E64}

This PR implements this capability as a new MFD class.

I've added unit tests and a QA test case (classical/case_10). The unit tests use parameters that are effectively equivalent (in terms of obtaining the same MFDs) when using either the regular AC MFD class or the `from_reference_rates` method (equivalent hazard curves are obtained).

I've updated where required throughout the engine to specify this MFD class within an XML and apply the associated uncertainties (some new ones added for AC zone b-value and modifying mmax and/or AC zone width).

- [x] Check which existing relative and absolute MFD-related uncertainties can be applied (e.g. `abGRAbsolute` should not work because we have multiple b-values (one per zone).
- [x] Add a class method allowing the instantiation of an AC MFD from the reference exceedance rates for given reference magnitudes which backcompute the total rate and gamma through inverting equation 1.2 of the BCHydro memo (see above equation) to support how the recurrence information is provided in the BCHydro input files.


openquake/qa_tests_data/classical/README.md
```diff
@@ -10,6 +10,7 @@
 |case\_07|Test DummyGMPE in event_based|
 |case\_08|Test override_vs30 with two values and disagg_by_src|
 |case\_09|Test ConditionalGMPE Macedo2019|
+|case\_10|Test use of alternative characterstic MFD|
```
micheles, line 13: Typo in `characterstic`

micheles (approved):

mmpagani (approved):
Great work! Thanks a lot Chris

PR #11365, opened 2026-04-10
Forbidded IMT-dependent weights for classical calculations with sampling

micheles:
Closes https://github.com/gem/oq-engine/issues/11350.
NB: we cannot break old event based calculations, so for those we log a warning as before.


CB-quakemodel (approved):
Thanks Michele, this looks like a good solution

PR #11375, opened 2026-04-14
Added parameter `keep_trt` to the job.ini

micheles:
So that we can restrict the gsim logic tree to the branchset for the specified TRT. Requested by @raoanirudh, to be tested more.


openquake/qa_tests_data/scenario/case_36/rupture.csv
```diff
@@ -1,3 +1,3 @@
-#,,,,,,,,,,"trts=['Stable Shallow Crust'], ses_seed=42"
+#,,,,,,,,,,"trts=['Active Shallow Crust', 'Stable Shallow Crust'], ses_seed=42"
```
raoanirudh, line 1: Why are there two TRTs in the comment header while there's only one TRT in the rupture?

openquake/hazardlib/gsim/nz22/nz_nshm2022_parker.py
```diff
@@ -452,7 +452,7 @@ class NZNSHM2022_ParkerEtAl2020SSlabB(NZNSHM2022_ParkerEtAl2020SSlab):
     For Cascadia and Japan where basins are defined (also require z2pt5).
     """
 
-    REQUIRES_SITES_PARAMETERS = {"vs30", "z2pt5"}
+    REQUIRES_SITES_PARAMETERS = {"vs30", "z2pt5", "backarc"}
```
raoanirudh, line 455: Is the addition of `backarc` here intentional?

raoanirudh (commented):
Since the TRT is an attribute of the rupture (and source), it might be cleaner in the long run to include it in the rupture.xml and rupture_dict (it's already included in the rupture.csv), rather than in the job.ini file.

openquake/hazardlib/gsim/nz22/nz_nshm2022_parker.py
micheles: Yes, I discovered a bug in the GSIM while implementing the test

openquake/qa_tests_data/scenario/case_36/rupture.csv
micheles: I wanted to see that happens if there are more trts in the header than in the file. This is a correct situation, happening when the rupture file has been exported by a calculation with two TRTs of which one with low occurrence rates, so that no ruptures are generated for it.

PR #11380, opened 2026-04-14
Add a h3 grid-lookup from hdf5-based GSIM adjustment utility class

CB-quakemodel:
Adds a utility GSIM class that applies h3-gridded location-based adjustments to the mean and sigma components of an underlying GSIM, resolved by either rupture hypocentre or site location.

This class can consider a h3-densifying grid, but the code works with only a single resolution too.

Configuration is stored at root level in the hdf5 as a `res_terms` dict attribute. Each key is a residual term (e.g. `dL2L`, `dS2S`) mapping to:

- **`location`**: `"hypo"` (uses `ctx.hypo_lon`/`ctx.hypo_lat` — e.g. `dL2L`) or `"site"` (uses `ctx.lon`/`ctx.lat` — e.g. `dS2S`) to determine which correction to apply
- **`sig_adjustment`** (optional, default `"none"`): whether to reduce (`"sub"`), inflate (`"add"`), or not modify the sigma component (`"none"`)
- **`sig_comp_modified`** (required when `sig_adjustment` is not `"none"`): which sigma component to adjust — `"sig"`, `"tau"`, or `"phi"` (e.g. for `dL2L` we would map to an adjustment to `tau` within this dict)

When `sig_adjustment` is `"none"` (the default), only the mean correction is applied — the `_sig` datasets in the hdf5 and the `sig_comp_modified` key are not required for that term.

The residual (corrective) terms stored in the `res_terms` dict are not fixed — the user can specify as they wish (e.g. only include `dS2S`). Checks ensure that if a residual term is mapped in this dictionary that it also is contained within the grid. 

The corrections must be IMT-dependent, and if an IMT lacks corrections for a given residual term, it is simply skipped. The h3 cell IDs (i.e. the grids) are on an IMT-by-IMT basis because the availability of data used to derive the corrections might vary with both period and term.

Unit tests and a new QA test (`classical/case_11`) are added for this capability.

The user can refer to `oq-engine\openquake\hazardlib\tests\gsim\mgmpe\data\test_grid_adjustments.hdf5` or the hdf5 in `classical/case_11` for an example of how the grid information is stored in the hdf5.

This class will evolve as we use it in real applications but right now it's **highly experimental**


openquake/hazardlib/gsim/mgmpe/grid_adjusted_gmpe.py
```diff
@@ -0,0 +1,294 @@
...
+    # Get adjustment data for given imt
+    entry = grid_data["grids"].get(str(imt))
+    
+    # Check if no data
+    if entry is None:
+        return 
+    
+    # Apply each correction
+    for term, cfg in grid_data["res_terms"].items():
+
```
micheles, line 162: Please add a comment here with an explicit example of `term` and `cfg`

mmpagani (approved):

micheles (approved):

PR #11381, opened 2026-04-15
Logging and storing the engine version

### LINKED ISSUE #11376: Log and store the correct engine version
Now we get the DbServer version, which is wrong. Hint: using something like
`cd /opt/openquake/oq-engine && git rev-parse --short HEAD` could work.

micheles:
Going back to the situation of 4 years ago (https://github.com/gem/oq-engine/pull/7924), since now git has finally fixed their bug forbidding the extraction of the git version for generic users. Closes https://github.com/gem/oq-engine/issues/11376

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,6 @@
+  [Michele Simionato]
+  * Logging and storing the engine version and not the DbServer version
+
   [Matteo Nastasi]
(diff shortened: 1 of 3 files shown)


raoanirudh (commented):
LGTM

PR #11383, opened 2026-04-15
Security: Untrusted remote archive download without integrity verification

tomaioo:
## Summary

Security: Untrusted remote archive download without integrity verification

## Problem

**Severity**: `High` | **File**: `openquake/commands/restore.py:L36`

When `archive` is a URL, the code downloads it via `requests.get(archive)` and immediately processes it as a trusted backup. There is no checksum/signature verification, no TLS pinning, and no timeout. This allows tampered archives (MITM/supply-chain risk) and can also cause hangs due to indefinite network waits.

## Solution

Require an expected digest/signature and verify it before extraction; enforce HTTPS-only URLs; set explicit timeouts (e.g., `timeout=(5, 30)`); and handle HTTP errors via `raise_for_status()`.

## Changes

- `openquake/commands/restore.py` (modified)

### DIFF
--- a/openquake/commands/restore.py
+++ b/openquake/commands/restore.py
@@ -23,12 +23,13 @@ import getpass
 import os.path
 import zipfile
 import sqlite3
+import hashlib
 import requests
 from openquake.baselib.general import safeprint
 from openquake.commonlib.dbapi import Db
 
 
-def main(archive, oqdata):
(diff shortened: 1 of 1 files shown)


micheles:
The `oq restore` command is internal, used only in our tests to download archives generated by us. As such, there are no security risks at all.

PR #11393, opened 2026-04-17
Refining `rupid` to be unique across mosaic models

micheles:
The idea is to solve the bug described in https://github.com/gem/oq-engine/pull/11391.
It does not work yet.


micheles:
Superseded by https://github.com/gem/oq-engine/pull/11394.

Issue #11392, opened 2026-04-17
Move `max_potential_paths` in openquake.cfg


micheles:
In cole and piaf has to be raised to 100_000_000_000_000. An alternative is to turn it into a warning.

raoanirudh:
The parameter that needs to be raised `event_based` calculations is `max_potential_gmfs`

PR #11400, opened 2026-04-21
Ignoring the max_potential_paths limit in event based calculations

micheles:
Perhaps sufficient for https://github.com/gem/oq-engine/issues/11392.


raoanirudh:
The issue in `event_based` calculations is with `max_potential_gmfs`

PR #11407, opened 2026-04-22
Fix beta function call in docstring with correct parameters

raoanirudh:
`rng.beta()` is called with parameters `a, b`, or `alpha, beta`: https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.beta.html

The actual call below is correct, so fixing only the docstring


raoanirudh:
Also check if the `Sampler` class is actually used anywhere, and if not, if it could be removed?
https://github.com/gem/oq-engine/blob/master/openquake/risklib/scientific.py#L134-L176

micheles:
The `Sampler` class is used all the time (both in scenario_risk and event_based_risk), it cannot be removed. Also, your fix breaks the doctest.

PR #11410, opened 2026-04-23
Add an epistemic uncertainty for aspect ratio in the SSC logic tree

CB-quakemodel:
I add in this PR a capability to apply either an absolute or relative aspect ratio to permit additional simplification of an SSC logic tree.

I add a QA test (logictree/case_32) and some unit tests too (in `logic_tree_tests.py`.

** Other **
I also moved the AC MFD epistemic uncertainty unit tests located here in `logic_tree_tests.py` into a separate class because they are long and class specific epistemic uncertainties.

Other some basic unit tests (I think) were missing for setting an MSR and the seismogenic depth uncertainties (only tested in QA tests), so I added some simple unit tests for these too. 

The absolute MSR epistemic uncertainty (dict of mag scaling relationships) was previously only supported for simple fault sources, but I think it makes sense to be able to use this uncertainty type in some other types too (still skip characteristic faults given inherently does not make sense here). The new unit tests for setting an MSR in `logic_tree_tests.py` test this on a point source.

### DIFF
--- a/openquake/commonlib/tests/logictree_test.py
+++ b/openquake/commonlib/tests/logictree_test.py
@@ -1400,13 +1400,7 @@ class BranchSetApplyUncertaintyTestCase(unittest.TestCase):
         self.assertEqual(inc_point_source.mfd.bin_width, 0.1)
         self.assertEqual(inc_point_source.mfd.occurrence_rates[0], 0.05)
         self.assertEqual(inc_point_source.mfd.occurrence_rates[1], 0.01)
-
-    def test_set_msr_absolute(self):
-        new_msr = openquake.hazardlib.scalerel.WC1994()
-        lt.apply_uncertainty('setMSRAbsolute', self.point_source, new_msr)
-        self.assertIsInstance(
-            self.point_source.magnitude_scaling_relationship,
-            openquake.hazardlib.scalerel.WC1994)
+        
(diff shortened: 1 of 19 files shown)


openquake/commonlib/tests/logictree_test.py
```diff
@@ -1401,96 +1401,126 @@ def test_absolute_incremental_mfd_uncertainty(self):
...
+        self.assertAlmostEqual(self.point_source.rupture_aspect_ratio, 2.0)
+
+    def test_aspect_ratio_relative_uncertainty(self):
+        lt.apply_uncertainty(
+            'setAspectRatioRelative', self.point_source, 0.5)
+        self.assertAlmostEqual(self.point_source.rupture_aspect_ratio, 1.5)
+
+
+class BranchSetApplyACMFDUncertaintyTestCase(unittest.TestCase):
+    def setUp(self):
```
micheles, line 1435: If this can be done only once per class it could become a classmethod:
```python
@classmethod
def setUpClass(cls):
   cls.source = ...
```

micheles: No, disregard that, the source.mfd is modified, so we need multiple instations.

micheles (approved):

PR #11412, opened 2026-04-23
Add ability to set a reference z1pt4 in job file

CB-quakemodel:
Small PR permitting specification of z1pt4 like already possible for vs30, z1pt0 and z2pt5.

We need this for bedrock calculations for JPN and KOR given they use the MF13 GMM which uses z1pt4.

Tested in an added QA test (classical/case_13).


micheles (approved):

micheles (approved):

raoanirudh:
The existing `oq prepare_site_model` command might need some refactoring to output the default values for z1pt0 (-999) and z2pt5 (-999) instead of using specific equations.

PR #11415, opened 2026-04-24
Add an epistemic uncertainty for setting a relative seismogenic depth

CB-quakemodel:
The existing capability was only supporting absolute uncertainties. The uncertainty is as expected applicable only to simple faults, area sources and point sources.

A new QA test (logictree/case_33) and unit tests are added.


mmpagani (approved):
LGTM thanks Chris

micheles (approved):

PR #11417, opened 2026-04-24
Add option of preparing a site model with sentinel z ref values instead

CB-quakemodel:
Update `prepare_site_model` command to permit assignment of sentinel z1pt0/z2pt5 instead if desired (default behaviour is still to use the NGAWest2 vs30 to z1pt0/z2pt5 relationships).

We now assign a sentinel value for z1pt0 or z2pt5 in the site models if we wish to tell the GMMs to compute the z1pt0 or z2pt5 from the considered GMM's vs30 to z ref relationship (see https://github.com/gem/oq-engine/pull/10594). 

This approach has been adopted in the 2026 Mosaic.

We need to permit this approach when the user is building a site model using the `prepare_site_model` command too.


openquake/commands/prepare_site_model.py
```diff
@@ -182,6 +193,7 @@ def main(
 main.z1pt0 = dict(help='build the z1pt0', abbrev='-1')
 main.z2pt5 = dict(help='build the z2pt5', abbrev='-2')
 main.vs30measured = dict(help='build the vs30measured', abbrev='-3')
+main.sentinel_z = dict(help='assign dummy -999 to z1pt0/z2pt5', abbrev='-D')
```
micheles: Does abbrev='-999' works? It would seem more intuitive to me.

CB-quakemodel: Yes it does, I've just pushed an updated version using -999 instead for the abbreviation

openquake/commands/tests/commands_test.py
```diff
@@ -696,6 +696,14 @@ def test(self):
         sc = sap.runline('openquake.commands prepare_site_model '
                          f'{vs30_csv} -s{vs30_csv} -12 -g10 -a5 -o {output}')
 
+        # test sentinel_z: z1pt0 and z2pt5 should be -999 for all sites
+        sc = sap.runline( # NOTE: -D is sentinel_z
+            'openquake.commands prepare_site_model '
+            f'{vs30_csv} -s{vs30_csv} -12 -999 -a5 -o {output}')
```
raoanirudh, line 702: What's `-12 -999 -a5` in this line?

CB-quakemodel: They are the inputs as described here for using the command - https://github.com/gem/oq-engine/blob/master/openquake/commands/prepare_site_model.py#L181

raoanirudh (approved):

micheles (approved):

micheles:
LGTM, but please update the changelog

CB-quakemodel:
> LGTM, but please update the changelog

Done!

PR #11423, opened 2026-04-27
Add ability to specify mag dependent aspect ratios

CB-quakemodel:
Add ability to specify magnitude dependent aspect ratios in the XML through predefined expressions that are evaluated as needed.

For now, I just add a simple linear piecewise following the basic form of:

{6E97DDF8-CF2D-4ADF-8349-CB84310F5C1B}

The syntax I implement in this PR to support a simple expression like this is:

{A2C476DA-60D2-4A23-B478-3AC3C918B978}

A new class called `MagDepAspectRatio` handles the expressions and is stored in a new hazardlib submodule called `aspect_ratio.py`.  We can add new expressions here as required.

Todo 
- [x] Add a QA test using all of the affected source typologies to check everything is working as intended.
- [x] Add unit tests that check expected values for interp/clamping behaviour +  writing mag-dep aratios back to xml node.
- [x]  Prevent specification of both a scalar aspect ratio and a mag-dep aratio for the same source in an XML.
- [x] Prevent use in kite faults because it uses profile sampling which I believe requires an aspect ratio prior to rup generation.
- [x] Prevent use of this capability in combination with epistemic uncertainties on aspect ratios (i.e. setting absolute values or relative increments).


openquake/hazardlib/source/base.py
```diff
@@ -37,6 +37,29 @@
...
 
+
+def linear_piecewise_aratio(mag, points):
+    """
+    Clamped piecewise-linear interpolation of aspect ratio from magnitude.
+
+    :param mag: Rupture magnitude (float)
+
+    :param points: List of (mag, aratio) pairs in ascending magnitude order
+
```
micheles: I don't like the name points, since they are not points. Perhaps `mag_aratios`?

openquake/hazardlib/geo/packager.py
```diff
@@ -93,8 +93,12 @@ def build_nodes(props):
     mfd_subnodes = [Node(tag, {}, scientificformat(dic[tag]))
                     for tag in dic.keys() if not tag.startswith('_')]
     msr = Node('magScaleRel', text=props['magscalerel'])
-    rar = Node('ruptAspectRatio',
-               text=scientificformat(props['ruptaspectratio']))
+    raw_rar = props['ruptaspectratio']
+    if isinstance(raw_rar, str) and raw_rar.startswith('{'):
+        from openquake.hazardlib.sourcewriter import build_aspect_ratio_node
```
micheles: Why we have an import here? To avoid a circular import? Then it is a code smell.

openquake/hazardlib/source/base.py
```diff
@@ -456,13 +480,39 @@ def modify_set_msr(self, new_msr):
...
+    def get_aspect_ratio(self, mag):
+        """
+        Return the rupture aspect ratio, which is either a float (the
+        "regular" version OR an expression of a PREDEFINED type which
+        is evaluated here).
+
+        More expressions can be added here in the future as required.
+        """
+        rar = self.rupture_aspect_ratio
+        if not isinstance(rar, dict):
```
micheles: I don't like that the magnitude-dependent aspect ratio is a dictionary. It should be its own class
```python
class MagDepAspectRatio:
    def __init__(self, ...):
           ...
    def get(self, mag):
          return ...

aratio = MagDepAspectRatio(...)
aratio.get(mag)
```

openquake/hazardlib/source/base.py
CB-quakemodel: Renamed throughout the PR to the phrase mag_points/mag_point

openquake/hazardlib/geo/packager.py
CB-quakemodel: I have fixed this by making a new hazardlib module for handling aspect ratio related functions and the mag-dependent aspect ratios. It makes sense I think given we will add more function types in the future to the `MagDepAspectRatio` class.

openquake/hazardlib/source/base.py
CB-quakemodel: I've added a class as suggested which is now hosted in the new `aspect_ratio.py`

mmpagani (approved):
Great work! LGTM

openquake/hazardlib/source/multi_point.py
```diff
@@ -33,7 +33,7 @@
 
 
 def get(arr, i):
-    if hasattr(arr, '__getitem__'):
+    if isinstance(arr, (list, numpy.ndarray)):
```
micheles: Now that MagDepAspectRatio is a class, I guess you can restore as it was (I prefer the `__getitem__` check since it works for any sequence).

openquake/hazardlib/aspect_ratio.py
```diff
@@ -0,0 +1,115 @@
...
+            mags = [m for m, _ in self.mag_points]
+            aratios = [a for _, a in self.mag_points]
+            if mag <= mags[0]:
+                return aratios[0]
+            if mag >= mags[-1]:
+                return aratios[-1]
+            for i in range(len(mags) - 1):
+                if mags[i] <= mag <= mags[i + 1]:
+                    t = (mag - mags[i]) / (mags[i + 1] - mags[i])
+                    return aratios[i] + t * (aratios[i + 1] - aratios[i])
```
micheles: If I am not mistaken, you are reinventing `np.interp(x, [xa, xb], [ya, yb])` i.e. a function constant outside the interval and linearly interpolating inside. You should be able to rewrite this more elegantly.

openquake/hazardlib/source/multi_point.py
CB-quakemodel: Restored to original

openquake/hazardlib/aspect_ratio.py
CB-quakemodel: Thanks, I have modified to use `np.interp`

micheles:
Impressive work, Chris!

micheles (approved):

PR #11428, opened 2026-04-29
Moved `max_potential_gmfs` to openquake.cfg

### LINKED ISSUE #11392: Move `max_potential_paths` in openquake.cfg

micheles:
Since it depends on the server. Closes https://github.com/gem/oq-engine/issues/11392


raoanirudh (commented):
LGTM

PR #11430, opened 2026-04-29
Add ability to specify a distribution of hypocentres in a simple fault source

CB-quakemodel:
Currently a PMF of hypo depths can only be specified for distributed seismicity but we need this capability for simple faults too for BCHydro SSC. 

When iterating over the floated ruptures the depth range of the given floating rupture is computed and then only the hypocentres within this depth range are retained, with the associated weights of the remaining depths renormalised. The computed down-dip fractions are then used to construct the hypocentres within the rupture object.

An optional value of `fixedDepFrac` permits a manual specification of the down-dip fraction used for assigning each hypocenter to the given floating rup, but if not present (optional column in `hypoDepthDist`) this is computed on the fly using the given floating rup (in this case the depths are still used to filter out against the given floating rup's depth range and then the ratio is fixed using the provided overriding value):

{2EBD1ECB-A3E3-48F8-ABA9-6C8900DD03B0}

For testing I add a QA test (classical/case_28) and some simple unit tests.

I split the existing `iter_rups` for clarity into helper functions for each hypocentre approach (centroid, hypo_depth_list, slip_list).

### DIFF
--- a/openquake/baselib/tests/flake8_test.py
+++ b/openquake/baselib/tests/flake8_test.py
@@ -52,8 +52,8 @@ def _long_funcs(module, maxlen):
         if isinstance(node, ast.FunctionDef):
             dotname = '%s.%s' % (module.__name__, node.name)
             args = node.args.args
-            if len(args) > 15:
-                raise SyntaxError('%s has more than 15 arguments: %s'
+            if len(args) > 16:
+                raise SyntaxError('%s has more than 16 arguments: %s'
                                   % (dotname, [a.arg for a in args]))
             doc = ast.get_docstring(node)
             # Adjust start line to skip the docstring if present
--- a/openquake/baselib/tests/flake8_test.py
+++ b/openquake/baselib/tests/flake8_test.py
@@ -52,8 +52,8 @@ def _long_funcs(module, maxlen):
         if isinstance(node, ast.FunctionDef):
             dotname = '%s.%s' % (module.__name__, node.name)
             args = node.args.args
-            if len(args) > 16:
-                raise SyntaxError('%s has more than 16 arguments: %s'
+            if len(args) > 15:
+                raise SyntaxError('%s has more than 15 arguments: %s'
                                   % (dotname, [a.arg for a in args]))
             doc = ast.get_docstring(node)
             # Adjust start line to skip the docstring if present
--- a/openquake/hazardlib/source/simple_fault.py
+++ b/openquake/hazardlib/source/simple_fault.py
@@ -175,14 +175,16 @@ class SimpleFaultSource(ParametricSeismicSource):
         sin_dip = math.sin(math.radians(self.dip))
         down_dip_delta = self.rupture_mesh_spacing * sin_dip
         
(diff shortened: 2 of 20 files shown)


openquake/baselib/tests/flake8_test.py
```diff
@@ -52,8 +52,8 @@ def _long_funcs(module, maxlen):
         if isinstance(node, ast.FunctionDef):
             dotname = '%s.%s' % (module.__name__, node.name)
             args = node.args.args
-            if len(args) > 15:
-                raise SyntaxError('%s has more than 15 arguments: %s'
+            if len(args) > 16:
+                raise SyntaxError('%s has more than 16 arguments: %s'
```
micheles: I really dislike functions with too many parameters. You can keep the same number of parameters by using an object combining hypo_slip_list and hypo_depth_list (an HypoData instance?)

CB-quakemodel: I now use this approach to group the alternative hypocentre specification approaches into one instance called `HypoData`

openquake/hazardlib/source/simple_fault.py
```diff
@@ -122,23 +148,112 @@ def __init__(self, source_id, name, tectonic_region_type,
...
             raise ValueError('mesh spacing %s is too high to represent '
                              'ruptures of magnitude %s' %
                              (rupture_mesh_spacing, min_mag))
 
+    def _hypo_list_from_depths(self, first_row, rup_rows):
+        """
+        Convert the hypo depth distribution to (dip_frac, weight) pairs
+        for a given floating rupture.
+
+        Depths outside [rup_top_depth, rup_bot_depth] are discarded
```
mmpagani: Wouldn't it be better to ask the user to provide the input distribution already in terms of a fraction of the fault width?

mmpagani (approved):
Thanks a lot Chris. This is extremely useful. Overall it looks good to me. The only note I left is about the use of absolute depth values as opposed to fractions of the fault width. The latter seems more flexibile to me but perhaps the use of absolute values is a requirement for the project.

openquake/hazardlib/source/simple_fault.py
CB-quakemodel: I've now added the option of specifying a down-dip fraction per depth within the `hypoDepthDist`. If not present, the current behaviour takes effect, and if a fraction is provided in the XML, this is used instead with the depths still being used but only to filter against the floating rupture's depth interval. Here is the syntax:

{2EBD1ECB-A3E3-48F8-ABA9-6C8900DD03B0}

CB-quakemodel:
> Thanks a lot Chris. This is extremely useful. Overall it looks good to me. The only note I left is about the use of absolute depth values as opposed to fractions of the fault width. The latter seems more flexibile to me but perhaps the use of absolute values is a requirement for the project.

Thanks Marco, I've addressed your comment by providing flexibility about this specification of either a depth or alternatively a fixed fraction, so I think both cases are now covered.

micheles (approved):

micheles:
Good job, Chris

PR #11436, opened 2026-05-04
Add raytracing utility to GridAdjustedGMPE

CB-quakemodel:
Alongside simple grid-based lookups for adjustments based on residual terms like dL2L, we might want to apply a distance-based correction (a classic example in a non-ergodic GMM being a residual attenuation rate (per km) for each grid cell, which is then applied to the mean ground-motion based on the distance through each grid cell by the given travel path).

This PR adds this feature and expands the existing QA (classical/case_11) and unit tests to consider it.


openquake/hazardlib/gsim/mgmpe/grid_adjusted_gmpe.py
```diff
@@ -164,24 +245,42 @@ def _apply_grid_corrections(grid_data, ctx, imt,
...
-        # The term can be selected based on hypo or site location
-        if cfg["location"] == "hypo":
-            # Hypo location-based
-            lats, lons = ctx.hypo_lat, ctx.hypo_lon
-        else:
-            # Site location-based
-            lats, lons = ctx.lat, ctx.lon
-
         # Check sigma adjustment configuration
         sig_action = cfg.get("sig_adjustment", "none")
```
micheles, line 254: A comment explaining the structure of the `cfg` dictionary would help

mmpagani (approved):

micheles (approved):

PR #11441, opened 2026-05-06
Refactored conditioned_gmfs/2

micheles:
```
USA on engine192
| calc_13743, maxmem=163.5 GB  | time_sec | memory_mb | counts |
| total get_mu_tau_phi         | 4_275    | 1_644     | 52     |
| EventBasedRiskCalculator.run | 476.7    | 749.0     | 1      |
| total ebr_from_gmfs          | 473.1    | 600.4     | 13     |
| EventBasedCalculator.run     | 413.7    | 573.4     | 1      |
| computing risk               | 233.0    | 0.0       | 20_024 |
```

### DIFF
--- a/openquake/hazardlib/calc/conditioned_gmfs.py
+++ b/openquake/hazardlib/calc/conditioned_gmfs.py
@@ -122,9 +122,9 @@ from openquake.hazardlib.geo.geodetic import geodetic_distance
 U32 = numpy.uint32
 F32 = numpy.float32
 
-Distance = namedtuple('Distance', 'YY YD DY DD')
+Precomputed = namedtuple('Precomputed', 'ctx_Y ctx_D YY YD DY DD mtp_args')
 
-def get_ctxs_dist(rupture, cmaker, inp):
+def get_precomputed(rupture, cmaker, inp):
     """
     :returns: (ctx_Y, ctx_D, dist)
     """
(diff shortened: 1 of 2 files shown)


openquake/hazardlib/calc/conditioned_gmfs.py
```diff
@@ -121,6 +122,33 @@
...
+    [ctx_Y] = cmaker.get_ctxs([rupture], inp.sites_Y)
+
+    # filter sites
+    mask_Y = numpy.isin(inp.sites_Y.sids, ctx_Y.sids)
+    inp.sites_Y = inp.sites_Y.filter(mask_Y)
+    mask_D = numpy.isin(inp.sites_D.sids, ctx_D.sids)
+    inp.sites_D = inp.sites_D.filter(mask_D)
+    inp.stations = inp.stations[mask_D].copy()
+
+    YY = compute_distance_matrix(inp.sites_Y, inp.sites_Y)
```
raoanirudh, line 146: This is the most expensive step, memory wise. The YY matrix is not used if `truncation_level = 0`, so we should avoid computing it at all costs if the job file has set `truncation_level = 0`.

In a scenario with 50,000 sites and 100 stations, with `truncation_level = 0`, the biggest matrix would be only 50,000 × 100 (the DY and YD matrices), and not 50,000 × 50,000 (YY).

PR #11443, opened 2026-05-06
Force the ASCE standards to be selected before picking a corresponding site class

ptormene:
I kept the rest of the chain logic unchanged (besides removing some duplication in the code): when the user selects the ASCE standards, the corresponding default site class is automatically pre-selected and the corresponding Vs30 value is set.

image


pslh (approved):
LGTM

micheles (approved):

PR #11458, opened 2026-05-13
Managed the change of name TWN->TEM plus ruptures not associated to any model

micheles:
Breaking the global risk model with errors like
```
  File "/home/michele/oq-engine/openquake/calculators/event_based.py", line 384, in get_allargs
    trt = trts[model][trt_smr // TWO24]
          ~~~~^^^^^^^
KeyError: '???'
```

### DIFF
--- a/openquake/hazardlib/countries.py
+++ b/openquake/hazardlib/countries.py
@@ -1,9 +1,12 @@
 import re
 from openquake.baselib import general
 
+# these are in one-to-one correspondence with the file mosaic.gpkg
 MODELS = sorted('''
 ALS ARB AUS CND CCA CEA CHN EUR GLD HAW IDN IND JPN KOR MEX MIE NAF NEA NWA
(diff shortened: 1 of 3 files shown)


openquake/hazardlib/countries.py
```diff
@@ -1,10 +1,13 @@
 import re
 from openquake.baselib import general
 
+# these are in one-to-one correspondence with the file mosaic.gpkg
 MODELS = sorted('''
 ALS ARB AUS CND CCA CEA CHN EUR GLD HAW IDN IND JPN KOR MEX MIE NAF NEA NWA
 NZL OAT OIN OPA PAC PHL PNG SAM SEA SSA TEM USA WAF ZAF PAN PAR'''.split())
 
+ALIASES = dict(TWN='TEM')  # manage change of name
```
raoanirudh, line 9: and `JPN='JPA'`

micheles: JPN is still working since in the list `MODELS` above is listed as `JPN` and not `JPA` where `TEM` is different

PR #11459, opened 2026-05-13
IMPACT: set the default `maximum_distance` based on trt and mag by interpolation

ptormene:


openquake/commonlib/oqvalidation.py
```diff
@@ -1138,7 +1148,8 @@ class OqParam(valid.ParamSet):
     steps_per_interval = valid.Param(valid.positiveint, 1)
     master_seed = valid.Param(valid.positiveint, 123456789)
     maximum_rupture_depth = valid.Param(valid.floatdict, {})  # km
-    maximum_distance = valid.Param(valid.IntegrationDistance.new)  # km
+    maximum_distance = valid.Param(valid.IntegrationDistance.new,
+                                   calc.filters.DEFAULT_INTEGRATION_DISTANCE)
```
micheles: For the engine used via the command-line, we DON'T want a default integration distance, we want to force the users to set it explicitly. It is only within OQImpact that we want to have a default integration distance.
Therefore setting the DEFAULT_INTEGRATION_DISTANCE at this level is a no-go.

openquake/hazardlib/calc/filters.py
```diff
@@ -39,6 +39,132 @@
...
+    "subduction intraslab":                            "SUBDUCTION_INTRASLAB",
+    "Subduction IntraSlab30":                          "SUBDUCTION_INTRASLAB",
+    "Subduction IntraSlab55":                          "SUBDUCTION_INTRASLAB",
+    "Subduction":                                      "SUBDUCTION_INTRASLAB",
+    "DeepSeismicity":                                  "SUBDUCTION_INTRASLAB",
+    # Volcanic
+    "Oceanic Crust with Volcanism":       "VOLCANIC",
+    "Volcanic":                           "VOLCANIC",
+}
+
```
micheles, line 166: This looks very fragile against future changes. At the minimum we need a test at the mosaic level to be able to maintain this mapping and I bet @mmpagani has an idea about the TRTs at the global scale.

PR #11460, opened 2026-05-13
Fix exporting of ruptures with gridded surfaces

CB-quakemodel:
A user reported an issue with exporting ruptures from an event-based calculation for the NEA model. I observe it locally too:

{AB88CB1B-B8F9-4D78-961C-D627C46ABEEE}

Upon inspection it was from slab non-parametric ruptures with gridded surfaces. Some of the ruptures have surfaces only comprising of 2 sets of coordinates and once they enter `plane_fit` (`hazardlib/geo/utils/`) the conditional check on dimensions of the `points` array inherently fails.

The strike of these surfaces can be crudely calculated without entering `plane_fit` (not applicable here given lack of points) by computing the horizontal azimuth of the two points, with this fix being implemented here to resolve the issue.

NB: I have checked the ruptures within the NEA hdf5 given it's initially confusing that only 2 points are being used to define a surface, but this is permitted in gridded surfaces and is presumably meant to represent very small (effectively point source) seismicity (in my small test calc using only the slab sources in NEA with an effective inv time of (only) 16,000 I see 385 ruptures have this 2 point defined surface mesh so it appears to be intentional).


mmpagani (approved):
Thanks Chris this is fine but I wonder of these ruptures should be actually filtered out (from the model)

micheles (approved):

PR #11462, opened 2026-05-14
Update MF13 to apply Phillipines intraslab correction term regardless of depth

CB-quakemodel:
An update to the MF13 GMM for the 2025 NSHM is that the `PH` correction term is applied to all intraslab events, rather than only those with a hypocentre depth less than 80 km like in the initial OQ implementation.


mmpagani (approved):

micheles:
Should we update the changelog? We have to do that if there is a change of numbers with respect to the last released version of the engine.

CB-quakemodel:
> Should we update the changelog? We have to do that if there is a change of numbers with respect to the last released version of the engine.

I have just updated it in the same line where I mention how we changed the xvf computation to a ctx-by-ctx basis.

micheles (approved):

micheles (approved):

PR #11467, opened 2026-05-14
Add option of down-dip fraction for hypo placement in hypoDepthDist

CB-quakemodel:
In this PR I add the ability to (optionally) specify a down-dip fraction for each depth in a `hypoDepthDist` for an area, point or multi-point source typology, which is used instead of the default of 0.5 (i.e., rupture centroid) if present when building a planar rupture in any distributed source typology.

The down-dip fraction places the hypocentre on the rupture (0=top, 1=bottom) with the rupture shifted so the hypocentre lands at that fraction along dip, then clamped to seismogenic depth range.

It's needed for BCHydro SSC implementation work (some overlapping code first implemented in https://github.com/gem/oq-engine/pull/11430 has been condensed into single helper functions in this PR in the source converter).

A new QA test (classical/case_31) and some simple unit tests are added.


micheles (approved):

micheles:
I have just checked that there is no performance penalty using performance.zip, so LGTM

mmpagani (approved):
Thank you, Chris. While I understand the logic of this new feature, I think it may create a conflict with the seismogenic layer definition, i.e., ruptures might never be able to fill the seismogenic layer OR they go beyond its limits. In the latter case, the ruptures generated might not adhere to the initial constraints on the hypocentral position on the rupture surface. I do not have a solution right now, but I believe we should consider a few checks to examine the interplay among the input parameters.

PR #11468, opened 2026-05-14
More Japan NSHM specific adjustments to the MF13 GMM

CB-quakemodel:
NIED NSHM implementation has more specifics we now know about:

- AI for SW region: applied only to deep PHP intraslab events in zones 4 and 5 of the NSHM

- PH term: applied to PHP intraslab earthquakes except for zones 3, 4, 5 of the NSHM

Update: @kejohnso has confirmed that these changes improve strongly the match to the NIED results.


mmpagani (approved):
Thank you Chris

micheles (approved):

PR #11470, opened 2026-05-15
Revert the mapping COUNTRY_CODE to Taiwan -> TWN

ptormene:
Oq-impact was failing for Taiwan because of the recent change TWN -> TEM.
Is Taiwan the only case in which we need to use a 3-letter code that is not the ISO code of the country, but an acronym of the corresponding earthquake model?


micheles (approved):

micheles:
Now, wait, this is wrong. We need to return "Taiwan" not "TWN"!
Study what is inside hazard/countries.py

micheles (approved):

PR #11472, opened 2026-05-16
MF13 PH corr zone adjustment

CB-quakemodel:
In our internal checks of the OQ implementation of the NIED model we see much better matches if omitting events in zone 3 from the `PH` correction, even though the NIED documentation states it should be applied to zones 3, 4 and 5.

UPDATE: NIED have confirmed the initial statement of application to zone 3 was incorrect, so the PR is correct implementation.


kejohnso (approved):
We have confirmation that this change is correct! LGTM, thanks Chris

micheles (approved):

micheles:
Please update the comment

PR #11482, opened 2026-05-20
Saved memory in the rupture sampling algorithm for fault sources

micheles:
This is crucial for the CHN model where we have a huge simple fault source called "HimalayanThrust" consuming 9+ GB of RAM per core with `rupture_mesh_spacing=2`:
```
# before/after
| calc_18486, maxmem=11.9 GB | time_sec | memory_mb | counts |
|----------------------------+----------+-----------+--------|
| total sample_ruptures      | 27.8629  | 9_397     | 1      |
| total sample_ruptures      | 23.9637  | 576.5     | 1      |
```
The memory saving is 16x, and we are also a bit faster.


mmpagani (approved):
LGTM

PR #11484, opened 2026-05-21
Inheriting `oqparam` correctly in calculations with SES.hdf5

### LINKED ISSUE #11483: Decide which `oqparam` to store in the workflow .hdf5 file
We are storing the first one
```python
            if wf_no == 0:  # at first step
                kw = wf.inis[0].copy()
                kw.update(calculation_mode='workflow')
                dstore['oqparam'] = OqParam(**kw)
```
This is then copied into the SES.hdf5 file by `build_ses`. Perhaps I should store a fake object instead to avoid confusing people.

micheles:
The correct logic is:

1. If the parameter is present in the job.ini file, it is used
2. If is it missing, it is inherited from the parent calculation

This is expecially important, since in the global risk calculations the `maximum_distance` is missing and must be inherited from the correct parent. This is possible since in ebrisk calculations the ruptures are partitioned by model, i..e. ruptures for JPN get the `maximum_distance` for JPN and ruptures from CHN get the maximum_distance for CHN.

Without this fix the `maximum_distance` for CHN was used everywhere, therefore discarding ruptures with mag > 8.

Closes https://github.com/gem/oq-engine/issues/11483


openquake/calculators/base.py
```diff
@@ -762,50 +763,48 @@ def pre_execute_from_parent(self):
...
+        if not isinstance(oqp, h5py.Group):  # from SES.hdf5
+            if 'weights' in parent:
+                weights = numpy.unique(parent['weights'][:])
+                if (oq.job_type == 'risk' and oq.collect_rlzs and
+                        len(weights) > 1):
+                    raise ValueError(
+                        'collect_rlzs=true can be specified only if '
+                        'the realizations have identical weights')
+            if oqp.imtls and oq.calculation_mode in ('classical', 'disagg'):
+                check_imtls(oq.imtls, oq.imtls)
```
raoanirudh: Should this be `check_imtls(oq.imtls, oqp.imtls)`?

openquake/calculators/event_based.py
```diff
@@ -365,17 +366,28 @@ def get_allargs(oq, sitecol, assetcol, sec_perils, dstore):
...
         events = dstore['events'][:]
         dstore['relevant_events'] = events[
             numpy.isin(events['rup_id'], filrups['id'])]
 
+        # populate oq_by when the parent is a SES.hdf5 file
+        grp = dstore.parent['oqparam']
+        if isinstance(grp, h5py.Group):
+            oq_by = {}
+            for model in grp:
+                oq_by[model] = oq.from_parent(dstore[f'oqparam/{model}'], True)
```
raoanirudh: Should this be 

```py
oq_by[model] = oq.from_parent(dstore.parent[f'oqparam/{model}'], True)
```

Can't debug easily on my computer, so I may be wrong!

openquake/calculators/base.py
micheles: indeed!

openquake/calculators/event_based.py
micheles: Yes, it works because `oqparam/{model}` is missing in the current dstore and it goes to the parent, but it is cleaner to look directly at the parent

raoanirudh:
Risk jobs starting from SES.hdf5 files that do not declare their own `maximum_distance` now fail with

```sh
  File "/opt/openquake/oq-engine/openquake/calculators/base.py", line 344, in run
    self.result = self.execute()
                  ^^^^^^^^^^^^^^
  File "/opt/openquake/oq-engine/openquake/calculators/event_based_risk.py", line 472, in execute
    raise InvalidFile('Missing maximum_distance in %s'
InvalidFile: Missing maximum_distance in job_risk.ini
```

due to the following check in `event_based_risk.py`:
https://github.com/gem/oq-engine/blob/c2f23c0542fb85c131f68c1ce1a90c2a692eef4b/openquake/calculators/event_based_risk.py#L471-L473

PR #11486, opened 2026-05-23
Fixed the check on maximum_distance in event_based_risk

### LINKED ISSUE #11483: Decide which `oqparam` to store in the workflow .hdf5 file

micheles:
Addresses https://github.com/gem/oq-engine/pull/11484#issuecomment-4522433025

While looking at this, I discovered another issue, right now all SES.hdf5 files contains the `oqparam` for Africa and not the right ones :-( This is due to the "include" functionality. I fixed that, but the SES.hdf5 have to be regenerated again :-( Really closes https://github.com/gem/oq-engine/issues/11483


raoanirudh (commented):
LGTM

PR #11490, opened 2026-05-25
fix numba issue in ebr ruptures with kite surfaces

CB-quakemodel:
Force lon/lat/depths in mesh to F32 to prevent numba decorator rejecting it when reconstructing a rupture with a kite surface from a CSV. Previously the following error was observed by risk team when using GEESE rups:

{D213270E-CC9D-4278-830A-DEBFD46D46AE}


CatalinaYepes (approved):
LGMT

micheles (approved):

PR #11494, opened 2026-05-26
Use the exposure.hdf5 generated in May

ptormene:
Only for this branch the CI action downloads the updated hdf5 (before merging we will have to remember to restore the action as it was and to replace the reference hdf5 file).
I have already fixed some expected files. Let's see if anything is still breaking using the updated data.


micheles (approved):

micheles:
Good work!

PR #11516, opened 2026-06-03
Update the script to build the hdf5 getting the correct values for the secondary perils

ptormene:
I made a mistake in the loop collecting site model data, therefore some variables were overwritten by values read from files with the same name format, that exist in different folders.
The result was that the variables for the secondary perils were collected, but contained unexpected zeros.
I regenerated the partial hdf5 and I am testing the updated file here. If tests are green, I will replace the old file with the new one and I will restore the action yml of this branch.

### DIFF
--- a/openquake/engine/build_exposure_hdf5.py
+++ b/openquake/engine/build_exposure_hdf5.py
@@ -139,16 +139,16 @@ def build_site_model(grm_dir):
     dfs = []
     # NOTE: assuming the site-models repository is cloned inside the grm_dir
     site_models_dir = os.path.join(grm_dir, 'site-models')
-    for cwd, dirs, files in os.walk(site_models_dir):
-        # skip os.walk subtrees that don't belong to a known region
-        if not any(cwd == os.path.join(site_models_dir, region) or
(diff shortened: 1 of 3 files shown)


openquake/engine/build_exposure_hdf5.py
```diff
@@ -139,16 +139,16 @@ def build_site_model(grm_dir):
...
-    for cwd, dirs, files in os.walk(site_models_dir):
-        # skip os.walk subtrees that don't belong to a known region
-        if not any(cwd == os.path.join(site_models_dir, region) or
-                   cwd.startswith(os.path.join(site_models_dir, region, ''))
-                   for region in REGIONS):
+    for region in REGIONS:
+        impact_dir = os.path.join(site_models_dir, region, 'Impact')
+        if not os.path.isdir(impact_dir):
+            logging.warning(
+                f'Folder "{impact_dir}" was not found. Skipping')
```
micheles: This should be a hard error and not a skip

ptormene: Indeed

PR #11520, opened 2026-06-04
numpy2.0 fixes in geo library

CB-quakemodel:
Fixing one more observed by risk team in GRM scenario work in `geodetic.py` and also pre-emptively catch another in `multiline.py`

Some event-based QA tests have very small value changes


micheles:
The dtype to use for lons and lats is `numpy.float32`, this is probably the reason why you are getting different numbers. Notice that lons and lats are kept at 32 bit to save memory and that is important.

PR #11524, opened 2026-06-05
spherical_to_cartesian numpy f32 fix + fix bug in check_aki_richards_convention

CB-quakemodel:
After a lot of digging, the `TypeError: loop of ufunc does not support argument 0 of type float which has no callable radians method` in `spherical_to_cartesian` error observed by risk team when using some rupture CSVs turns out to have a simple fix - in numpy 2.x, `array.flat` passed to a ufunc produces a `dtype=object` array, so replacing `.flat` with `.ravel()` in `Mesh.xyz` does the trick.

IMPORTANT
There was also a bug in `check_aki_richards_convention` where the bottom-left depth was used twice instead of bottom-left and upper-left, incorrectly causing QA test `logictree/case_20` to fail under F32 precision (we are back to F32, but the error still existed and has been fixed).

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,7 @@
+  [Christopher Brooks]
+  * Fix a bug observed in the Aki and Richard's convention check in
+    complex faults.
+
   [Michele Simionato]
   * Optimized reading and sampling multiFaultSources
   * Saved memory in classical calculations with large sources affecting many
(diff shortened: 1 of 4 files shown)


mmpagani (approved):

micheles (approved):

micheles:
Please update the changelog about the `check_aki_richards_convention` fix

PR #11533, opened 2026-06-09
Implementing secondary perils for ShakeMaps

micheles:


openquake/hazardlib/imt.py
```diff
@@ -47,6 +47,13 @@ def imt2tup(string):
     """
     s = string.strip()
     name, *rest = s.split('(')
+
```
micheles, line 51: It is essential to explain WHY we are doing this. From the traceback
```python
openquake/calculators/base.py:1387: in import_gmfs_csv
    missing = set(oqparam.imtls) - set(imts)
                  ^^^^^^^^^^^^^
openquake/commonlib/oqvalidation.py:1821: in imtls
    return DictArray(sort_by_imt(imtls)) if imtls else {}
                     ^^^^^^^^^^^^^^^^^^
openquake/hazardlib/imt.py:115: in sort_by_imt
    imts = sorted(imtls, key=lambda imt: from_string(imt).period)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
openquake/hazardlib/imt.py:115: in <lambda>
    imts = sorted(imtls, key=lambda imt: from_string(imt).period)
                                         ^^^^^^^^^^^^^^^^
openquake/hazardlib/imt.py:104: in from_string
    return IMT(*imt2tup(imt))
                ^^^^^^^^^^^^
openquake/hazardlib/imt.py:51: in imt2tup
    raise KeyError(name)
E   KeyError: 'AllstadtEtAl2022Landslides_LsProb'
```
it looks like the reason is to be able to sort the columns when importing the GMFs. Please add a comment about that.

ptormene: I added a comment inside the docstring

PR #11534, opened 2026-06-09
Fix Nowicki Jessee landcover coefficient assignment

catarinaquintela:
The landcover coefficient in the Nowicki Jessee model was being incorrectly assigned to the default value.

The issue occurred because the landcover input was converted to a string after being read as a float. For example, a landcover value of 60 in sitemodel.csv was parsed as the float 60.0 and then converted to the string "60.0". Since the landcover_values dictionary uses keys such as "60", the lookup failed and the default coefficient was used instead.

The bug was identified following a user report on the OpenQuake Users mailing list:
https://groups.google.com/g/openquake-users/c/yc2BfQEQtRA

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,6 @@
+  [Catarina Costa]
+  * Fixed landcover coefficient assignment on Nowicki Jessee landslide model
+
   [Michele Simionato]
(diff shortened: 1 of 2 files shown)


micheles:
Please update the changelog

LanaTodorovic93 (approved):

PR #11538, opened 2026-06-10
Saving `year` and `ses_id` in the events dataset of SES.hdf5 files

micheles:
It was missing.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -2,6 +2,7 @@
   * Fixed landcover coefficient assignment on Nowicki Jessee landslide model
 
   [Michele Simionato]
+  * Saving `year` and `ses_id` in the events dataset of SES.hdf5 files
   * Splitting multifault and multipoint sources with the colon convention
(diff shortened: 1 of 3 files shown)


raoanirudh (commented):
LGTM

raoanirudh:
Turns out this PR also inadvertently fixed another issue affecting the calculation of AEP and OEP aggregate loss curves.

Risk calculations using combined SES.hdf5 files created before this PR were producing only the generic EP curves, despite specifying in the risk job files:

`aggregate_loss_curves_types = ep, oep, aep`

`LossCurvesMapsBuilder.build_curve()` calculates AEP and OEP curves only when event years are available, grouping event losses by year, summing them for AEP and taking their maximum for OEP.

Since the original `year` and `ses_id` fields were being dropped while saving the combined SES.hdf5, downstream risk calculations were silently skipping AEP and OEP generation prior to this PR.

PR #11553, opened 2026-06-16
Fix BS24 PGA row coefficients

thuanycostadelima:
Fix a mistake in the PGA row of the BaylessSomerville2024 coefficient 
table. The pga row was incorrectly using T=0.01s values instead of the 
T=0 values from Table D-7 of the paper.

Corrected values:
- Cratonic pga: a1 = 1.4840 (was 1.5042)
- NonCratonic pga: a1 = 0.8943 (was 0.9145)


CB-quakemodel:
Hi @thuanycostadelima, for the GSIM's unit tests to pass you will need to regenerate the expected results CSVs given the coefficients have changed.

CB-quakemodel (approved):

micheles (approved):

PR #11554, opened 2026-06-16
Extract exposure by LSE

ptormene:
I am adding the possibility to extract exposure aggregated by secondary peril LSE tiers and tags using a new API endpoint `GET /v1/calc/:calc_id/exposure_by_lse`.
Corresponding buttons are added to the oq-impact job outputs page, so the same results can be visualized similarly to `Show exposure by MMI table`.


openquake/calculators/export/__init__.py
```diff
@@ -122,6 +122,7 @@
     'occupants_avg': 'Average number of occupants',
     'mmi': ('Macroseismic intensity (MMI) to which the given group'
             ' of assets is subjected'),
+    'embodied_carbon': 'Embodied carbon',
```
ptormene: @catarinaquintela what's the measurement unit we are using for embodied carbon?

catarinaquintela: @martinacaruso ?

catarinaquintela: It's Tonnes of CO2e

openquake/calculators/export/__init__.py
```diff
@@ -100,7 +100,7 @@
     'injured': 'Number of injured people',
     'affectedpop': ('Number of people living in buildings '
                     'with moderate or higher damage'),
-    'embodied_carbon': 'Embodied carbon loss',
+    'embodied_carbon': 'Embodied carbon loss (tCO<sub>2</sub>e)',
```
micheles: This string appears in the command-line, it should not contain HTML tags, rather the WebUI should do the translation work if desired.

openquake/risklib/asset.py
```diff
@@ -575,6 +575,137 @@ def get_mmi_values(self, aggregate_by, mmi_file, exposure_hdf5):
...
+            if True, discard from the output all tier bins with no assets
+            (default: True)
+        :returns:
+            a DataFrame with aggregated exposure metrics grouped by
+            geo-tags and the chosen LSE peril tiers
+        """
+        if secondary_peril == "liquefaction":
+            lse_col = "AllstadtEtAl2022Liquefaction_LSE"
+            tier_col = "liquefaction_lse_tier"
+            bins = [-numpy.inf, 0.005, 0.01, 0.02, 0.05, 0.1, 0.2, numpy.inf]
```
micheles: I guess you can start from 0 and not `-inf`

openquake/calculators/export/__init__.py
ptormene: I removed the html tags by using unicode characters in tCO₂e and m². It should be ok both when displayed in the browser and in the terminal.

openquake/risklib/asset.py
ptormene: Ok, if we are sure that it is impossible to observe any tiny negative values due for instance to numerical approximations.

micheles: Generating negative values would be an error, and it is always better not to hide it.

openquake/server/views.py
```diff
@@ -2383,10 +2435,17 @@ def can_extract(request, resource):
     return False
 
 
+def _decode(v):
+    return v.decode('utf-8') if isinstance(v, bytes) else v
```
micheles: We already have this function inside baselib.general

openquake/calculators/tests/scenario_risk_test.py
```diff
@@ -315,6 +315,15 @@ def test_case_shakemap(self):
         [fname] = export(('realizations', 'csv'), self.calc.datastore)
         self.assertEqualFiles('expected/realizations.csv', fname)
 
+        # exposure_by_lse
+        for secondary_peril in ['liquefaction', 'landslide']:
+            df = extract(self.calc.datastore,
+                         f'exposure_by_lse?secondary_peril={secondary_peril}')
```
micheles, line 321: There is a potential performance/memory issue here. Every time a user calls the API, the entire exposure and avg_gmf are read in memory and aggregated using pandas, which is very memory hungry. So the webserver can potentially explode (out of memory) and the user can potentially wait a lot of time even if the memory is enough, since the call is blocking. The solution is to use a "material view", i.e. performing the aggregation once at the end of the calculation and save it in the HDF5 file, similarly with how we save the PNG files instead of generating them at runtime in AELO. However, I would do that in a separate PR. We need a big calculation to test the performance.

micheles (approved):

openquake/server/views.py
```diff
@@ -2532,7 +2528,7 @@ def extract_html_table(request, calc_id, name):
 
     # Decode byte strings to plain str
     table_rows = [
-        list(zip(table_header, [_decode(v) for v in row]))
+        list(zip(table_header, [decode(v) for v in row]))
         for row in table_contents
```
micheles, line 2532: You can just use `decode(row)`, since `baselib.decode` is smart

PR #11567, opened 2026-06-18
Create and store exposure_by_lse right after creating avg_gmf

ptormene:
Addressing https://github.com/gem/oq-engine/pull/11554#discussion_r3429096645


openquake/calculators/extract.py
```diff
@@ -866,6 +866,23 @@ def extract_mmi_tags(dstore, what):
     return df
 
 
+def ensure_npy_serializable(df):
+    """
```
micheles, line 870: This is really inefficient. You just need a check, there is no need to convert the whole columns.

openquake/calculators/extract.py
```diff
@@ -868,18 +868,18 @@ def extract_mmi_tags(dstore, what):
...
+    is serializable with ``allow_pickle=False``.
+
+    :param df: a :class:`pandas.DataFrame`
+    :returns: the same DataFrame with object columns replaced in-place
+    """
+    for col in df.columns:
+        if df[col].dtype != object:
+            continue
+        df[col] = numpy.array(
+            [decode(v).encode('utf-8') for v in df[col]], dtype='S')
```
micheles: `decode(df[col])` should be enough

openquake/calculators/extract.py
```diff
@@ -878,8 +878,7 @@ def ensure_npy_serializable(df):
     for col in df.columns:
         if df[col].dtype != object:
             continue
-        df[col] = numpy.array(
-            [decode(v).encode('utf-8') for v in df[col]], dtype='S')
+        df[col] = df[col].astype('S')
```
micheles, line 881: I like this

micheles (approved):

PR #11578, opened 2026-06-29
Do not divide PGV values by 100 when calculating GMFs

ptormene:
After the fix we have no broken tests. Perhaps we need a test checking that numbers are correct.


openquake/hazardlib/shakemap/gmfs.py
```diff
@@ -169,6 +169,23 @@ def cholesky(spatial_cov, cross_corr):
...
+    """
+    # NOTE: dtype=float ensures the final gmfs to have max precision
+    # floating-point accuracy. We might prefer instead to avoid specifying it,
+    # so the default dtype would match the native precision of ShakeMap data.
+    scale = numpy.array([
+        numpy.full_like(
+            shakemap_std[str(im)],
+            1.0 if 'PGV' in str(im) else pctg_value,
+            dtype=float)
+        for im in imts]).flatten()
```
micheles: Instead of a `flatten` you could use a `reshape`

micheles:
Yes, we are missing a test, please add it

catarinaquintela (approved):
LGTM

openquake/hazardlib/shakemap/gmfs.py
```diff
@@ -302,7 +323,23 @@ def to_gmfs(shakemap, gmf_dict, vs30, truncation_level,
...
+    # Validate suspiciously large values per IMT
+    for i, im in enumerate(imts):
+        imt_str = str(im)
+        imt_gmf_max = out_gmfs[:, :, i].max()
+        if imt_str != 'PGV':
+            if imt_gmf_max > MAX_GMV:
+                logging.warning(
+                    'There are suspiciously large GMVs for %s of %.2fg',
+                    imt_str, imt_gmf_max)
+        else:
```
micheles: You can use an `elif` here

micheles:
Please update the changelog

PR #11598, opened 2026-07-07
Fix PGV period handling in the JB2009 spatial correlation model

### LINKED ISSUE #11597: Clarification on the intended treatment of PGV in the JB2009 spatial correlation model
Hello OpenQuake developers,

I would appreciate clarification regarding the intended treatment of PGV in the implementation of the Jayaram and Baker (2009) spatial correlation model.

In both OpenQuake Engine v3.25.1 and the current master branch, jbcorrelation contains:

period = 1.0 if imt.string == 'PGV' else imt.period

if period < 1:
    if not vs30_clustering:
        b = 8.5 + 17.2 * imt.period
    else:
        b = 40.7 - 15.0 * imt.period
else:
    b = 22.0 + 3.7 * imt.period

However, PGV() is constructed without an explicit period:

def PGV():
    return IMT('PGV')

and the default value of IMT.period is 0.0.

Consequently, for PGV, the local variable used to select the equation branch is:

period = 1.0

but the range parameter is calculated using:

imt.period = 0.0

The effective calculation is therefore:

b = 22.0 + 3.7 * 0.0
  = 22.0

By contrast, a complete numerical equivalence with SA(1.0) would give:

b = 22.0 + 3.7 * 1.0
  = 25.7

This caught my attention because commit 065358c1cfa5c8e8b15da042012e6d157f6bf00a is titled:

In JB2009 correlation model, treat PGV similar to SA(1.0)

The commit introduced the local period variable for selecting the equation branch, while the expressions used to calculate b continued to reference imt.period.

I understand that Jayaram and Baker (2009) did not directly calibrate their proposed model for PGV. The paper states only that PGV may be qualitatively compared with spectral accelerations at moderate periods, approximately 0.5–1.0 s. Therefore, I am not assuming that b = 25.7 is necessarily the uniquely correct physical value for PGV.

Could you please clarify the intended OpenQuake behavior?

Is PGV intentionally assigned to the T >= 1 s branch while retaining imt.period = 0, resulting in b = 22.0?
Or was the intention to treat PGV numerically as SA(1.0), in which case the local period variable would also be used in calculating b, resulting in b = 25.7?
Is the current behavior retained for backward compatibility?
Would adding a dedicated PGV unit test to correlation_test.py be appropriate?

If complete equivalence with SA(1.0) is intended, an internally consistent implementation might be:

period = 1.0 if imt.string == 'PGV' else imt.period

if period < 1:
    if not vs30_clustering:
        b = 8.5 + 17.2 * period
    else:
        b = 40.7 - 15.0 * period
else:
    b = 22.0 + 3.7 * period

Thank you for your clarification and for maintaining OpenQuake.

Shahmohamadian:
## Summary

This pull request fixes the treatment of PGV in the Jayaram and Baker (2009) spatial correlation model.

The implementation defines a local `period` variable so that PGV is treated as equivalent to `SA(1.0)`. However, the equations used to calculate the correlation range parameter `b` currently reference `imt.period`.

Since `PGV().period` defaults to `0.0`, the current implementation results in `b = 22.0` rather than the intended `b = 25.7`.

## Changes

- Replaced `imt.period` with the local `period` variable in the three equations used to calculate `b`.
- Added a dedicated PGV regression test.
- Added coverage for the equivalence of the JB2009 correlation matrices for `PGV()` and `SA(1.0)`, both with and without Vs30 clustering.

Fixes #11597

### DIFF
--- a/openquake/hazardlib/correlation.py
+++ b/openquake/hazardlib/correlation.py
@@ -137,13 +137,13 @@ def jbcorrelation(sites_or_distances, imt, vs30_clustering=False):
     if period < 1:
         if not vs30_clustering:
             # case 1, eq. (17)
-            b = 8.5 + 17.2 * imt.period
+            b = 8.5 + 17.2 * period
         else:
             # case 2, eq. (18)
-            b = 40.7 - 15.0 * imt.period
+            b = 40.7 - 15.0 * period
     else:
         # both cases, eq. (19)
-        b = 22.0 + 3.7 * imt.period
+        b = 22.0 + 3.7 * period
 
     # eq. (20)
     return numpy.exp((- 3.0 / b) * distances)
--- a/openquake/hazardlib/tests/correlation_test.py
+++ b/openquake/hazardlib/tests/correlation_test.py
@@ -17,7 +17,7 @@ import unittest
 
 import numpy
 
-from openquake.hazardlib.imt import SA, PGA
+from openquake.hazardlib.imt import SA, PGA, PGV
 from openquake.hazardlib.correlation import JB2009CorrelationModel, \
                                             HM2018CorrelationModel
 from openquake.hazardlib.site import Site, SiteCollection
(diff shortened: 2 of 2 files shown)


raoanirudh (approved):

micheles:
I will fix the failing test add you to the contributors list

micheles (approved):

PR #11616, opened 2026-07-14
A simple idea for correlated uncertainties

micheles:
To be discussed with @CB-quakemodel and @mmpagani


micheles:
I am marking the approach experimental, with the idea of applying it to the BCHydro project and see how it goes. It could be changed or even removed in the future.

PR #11628, opened 2026-07-17
BC Hydro alt2 epistemic uncertainties

CB-quakemodel:
Add epistemic uncertainties for alternative 2 of BC Hydro

Internal tests using full branches of alt2 show it works well (good match to BC Hydro expected + identical to the original runtime approach but instead using applyToSources):

plots_siteID_LAJ_mean_PGA


micheles:
Good job, Chris!

micheles (approved):

PR #11632, opened 2026-07-20
Add epistemic uncertainties for alt3 of BC Hydro NVA SSC LT

CB-quakemodel:
Adds support for BC Hydro NVA alternative 3 by extending `recurSet` with an optional per-source `mmaxOffset` (`bg mmax = fault mmax − 0.5`) and `recurRow` with optional `bg` or `fault`  prefixed row attributes (bg and faults get different but correlated recurrence params from one branch).

We get a good match to the expected results (and also identical to the OQ runtime approach):

plots_siteID_CMX_mean_SA(1 0)


micheles:
There is a docstring to fix

micheles (approved):

PR #11633, opened 2026-07-20
Fixed saving avg_gmf

micheles:
Bug discovered by @ZarrinGEM


micheles:
We also need a test for `oq plot avg_gmf?imt=PGA`

PR #11649, opened 2026-07-24
Installation updates from engine-3.26

pslh:
Updates to install.py to better handle post-install for external django applications such as django-gem-taxonomy, for user, devel, server and devel_server installations.

Updated installation documentation to include support for "Windows protected your PC" dialog caused by the use of a new MS certificate.  Updated Python versions.  Some minor EN language updates.

Updated installer tests including handling of AELO/IMPACT for server

Removed "Download job.zip" button from output.html - this makes sense only in IMPACT mode (which has a dedicated output html file)

Updated changelog


vot4anto (approved):
LGTM

debian/changelog
```diff
@@ -39,6 +39,23 @@
...
+  * Fixed bug in classical_risk from hazard_curves.csv manifesting as a
+    KeyError in U32([sid2idx[sid] for sid in df.sid])
+
+ -- Paul Henshaw (GEM Foundation) <paul@openquake.org>  Fri, 24 Jul 2026 09:28:04 +0000
+
+python3-oq-engine (3.26.1-1~xenial01) xenial; urgency=low
+  [Matteo Nastasi, Antonio Ettorre]
+  * Fix and improve of post installation scripts
+
+
```
nastasi-oq, line 61: Please add a footer release line as at 52 for the new one, here.

doc/getting-started/installation-instructions/development.rst
```diff
@@ -30,18 +30,18 @@ Linux
 -----
 
 Please check the documentation of your distribution to install one of
-the supported version of Python. At the moment we recommend Python version 3.12
-but Python 3.11 can also be used.
+the supported version of Python. At the moment we recommend Python 3.12,
```
micheles: Here we recommend Python 3.12 but below we recomment 3.13. I would recommend 3.13 always.

install.py
```diff
@@ -183,9 +214,12 @@ class devel(user):
     """
     Parameters for a devel installation (same as user)
     """
-
     exit = user.exit
 
+    @classmethod
+    def manage_py(cls):
+        return os.path.join('openquake', 'server', 'manage.py')
```
nastasi-oq, line 221: Is it correct and absolute this path? The other seems to be absolute and this not.

pslh: It is deliberately a relative path. install.py must be used in the git repo directory for devel and devel_server modes (and will exit if this is not the case) openquake/server/manage.py is the relative path as seen by install.py

micheles (approved):

nastasi-oq (approved):
LGTM

PR #11662, opened 2026-07-28
Fix `export_asce` taking into account the corner case using the `default` site class

ptormene:
Weekly tests were broken. I am running them here:
https://github.com/gem/oq-engine/actions/runs/30365478084

### DIFF
--- a/openquake/calculators/export/hazard.py
+++ b/openquake/calculators/export/hazard.py
@@ -805,15 +805,15 @@ def export_mce(ekey, dstore):
 @export.add(('asce07', 'csv'), ('asce41', 'csv'))
 def export_asce(ekey, dstore):
     sitecol = dstore['sitecol']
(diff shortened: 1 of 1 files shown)


openquake/calculators/export/hazard.py
```diff
@@ -806,7 +806,14 @@ def export_mce(ekey, dstore):
 def export_asce(ekey, dstore):
     sitecol = dstore['sitecol']
     for s, site in enumerate(sitecol):
```
micheles: I don't like this. Something like
```python
if len(sitecol) == 3 and len(dstore[ekey[0]) == 1:
   # default site class case
   sids = [0]
else:
   sids = sitecol.sids
for s in sids:
      ...
```
would be better

ptormene: Done, thanks

micheles (approved):

PR #11663, opened 2026-07-28
[WIP] making job file flexible for aelo command line run

kejohnso:


openquake/commands/mosaic.py
```diff
@@ -52,7 +52,7 @@ def engine_profile(jobctx, nrows):
 
 # NB: this is called by the action mosaic/.gitlab-ci.yml
 def from_file(fname, mosaic_dir, asce_dir, concurrent_jobs,
-              asce_version, vs30):
+              asce_version, vs30, jobcfg=None):
```
micheles: `jobcfg` in other places in the codebase is called `job_ini` so I would stick to that convention

kejohnso: okay, done !

ptormene: From the changes I see in this PR, it looks like not all occurrences of `jobcfg` and `jobconfig` have been replaced with `job_ini`. @kejohnso perhaps you forgot to push some changes?

kejohnso: you're right @ptormene, they are correct now

PR #11668, opened 2026-07-29
Github workflow test config update

pslh:
Set job timeout to 60min and add set -x to server job to assist debug


pslh:
Closing - used only for testing.

PR #11670, opened 2026-07-29
Testing engine_pr_test Action - starting from hash 7ef4cbb40fbe80cf19c4012d3ba32ef008d9a739 

pslh:
Branch started from hash 7ef4cbb40fbe80cf19c4012d3ba32ef008d9a739 - tests were green for PT PR [fix_export_asce](https://github.com/gem/oq-engine/tree/refs/heads/fix_export_asce)

Added Timeout and debug changes to engine_pr_workflow.

If the engine_pr_tests run to completion, this suggests the error is not (just) due to a github problem but is a change applied to the engine since the hash.   If not, it suggests the problem is with GH.

### DIFF
--- a/.github/workflows/engine_pr_test.yml
+++ b/.github/workflows/engine_pr_test.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   calculators:
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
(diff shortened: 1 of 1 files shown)


pslh:
Closing - used only for testing.

PR #11672, opened 2026-07-30
Improved `export_asce`

micheles:
Continuation of https://github.com/gem/oq-engine/pull/11662

### DIFF
--- a/openquake/calculators/export/hazard.py
+++ b/openquake/calculators/export/hazard.py
@@ -802,7 +802,7 @@ def export_mce(ekey, dstore):
     return [fname]
 
 
-def unique_sites(sitecol):
+def sites_by_xy(sitecol):
     by_xy = {}
     for site in sitecol:
         xy = site.location.x, site.location.y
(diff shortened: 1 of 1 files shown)


openquake/calculators/export/hazard.py
```diff
@@ -802,25 +802,27 @@ def export_mce(ekey, dstore):
...
         dic = json.loads(js)
         writer = writers.CsvWriter(fmt='%.5f')
         fname = dstore.export_path(ekey[0] + '-' + str(s) + '.csv')
         comment = dstore.metadata.copy()
-        comment['lon'] = sitecol.lons[s]
-        comment['lat'] = sitecol.lats[s]
-        comment['vs30'] = sitecol.vs30[s]
+        comment['lon'] = site.location.x
+        comment['lat'] = site.location.y
+        comment['vs30'] = site.vs30  # only the first vs30
```
ptormene: I suppose it would be more correct if we listed the multiple Vs30 values for the sites at the same location

micheles: Indeed, this is an initial attempt

PR #11676, opened 2026-07-30
Speedup Test Case 25 (BC Hydro)

CB-quakemodel:
It was very slow, so now use a single (greatly simplified) version of the full NVA model with sampling rather than a seperate calc for each alternative. It passes in about 33 seconds (which is more comparable to the other slower tests - previously it was taking 41 seconds).

This test is also now more representative of the full NVA model + it reduces the number of test files used.

It could be sped up more by using less than 50 samples (full LT is 144 branches), but using less than this would be insufficient I think given the complexity of the test model.


micheles (approved):

micheles:
Excellent!

PR #11682, opened 2026-08-03
Fix bug in CB14 USGS Basin Term

CB-quakemodel:
Fixing a bug present in the basin term of the CB14 GMM associated with application of the NSHM basin scaling adjustment reported by Mohsen Kohrangi (whom also kindly provided the correction required).

The existing QA tests for the USGS NSHM GMM modifications were broken (and updated in this PR), and thus a separate regression test was not required (I should note that the expected results in a provided test case were privately replicated as a sanity check).


micheles (approved):

micheles:
Please update the changelog

micheles (approved):

PR #11683, opened 2026-08-04
Fix bug in EMME site term

CB-quakemodel:
The site term should be applied after the backbone scaling factors have been applied. The site term in CY14 is computed using vs30 of 800 m/s and CY14 reference of 1130 m/s. Then following the backbone scaling, the EMME site model which uses a reference velocity of 800 m/s is applied. We get a good match to the updated MATLAB implementation provided by the model authors (max absolute difference over all test scenarios of 0.38 percent).

Also, within the EMME model, although CY14 is used as the backbone, they do not consider a conditional on whether to use PGA if T ≤ 0.3 and SA(T) is less than PGA, so this is fixed too - now if the EMME model is being used, this is skipped in the CY14 compute method.

NB: Also some refactoring of CY14 linear site term to clean up the Canada GMMs and the Stafford (2022) GMM a bit.


micheles:
No, you are adding 31k lines of CSV files, too much.

CB-quakemodel:
> No, you are adding 31k lines of CSV files, too much.

> No, you are adding 31k lines of CSV files, too much.

Fixed - now only same scenarios as in current master test files are regenerated

micheles:
Is this still Work-in-Progress?

CB-quakemodel:
> Is this still Work-in-Progress?

Yes we are just waiting for some independent checks to be performed by the authors of the model to confirm this is the appropriate fix.

micheles (approved):

PR #11684, opened 2026-08-04
Make the button "Download job.zip" available for all calculations

### LINKED ISSUE #11641: Make sure the button "Download job.zip" works for all calculations
In engine-3.26 we decided to hide the button outside of OQImpact, but it would be much nicer to have it work.
One should test all the demos, for instance in `bin/run-demos.sh` after running each demo one should call
`oq export job -e zip` and make sure there are no errors like
```python
  File "/home/michele/oq-engine/openquake/calculators/export/risk.py", line 954, in export_job_zip
    model = dstore['ruptures'][0]['model'].decode('ascii')
            ~~~~~~^^^^^^^^^^^^
  File "/home/michele/oq-engine/openquake/commonlib/datastore.py", line 555, in __getitem__
    raise KeyError('No %r found in %s' % (key, self))
KeyError: "No 'ruptures' found in <DataStore /home/michele/oqdata/calc_15379.hdf5 open>"
```
which obviously happens for classical calculations where there are no ruptures stored.

ptormene:
Fixes https://github.com/gem/oq-engine/issues/11641

See a run here: https://github.com/gem/oq-engine/actions/runs/31702137505/job/94453613787#step:7:3953


openquake/calculators/export/risk.py
```diff
@@ -954,28 +944,93 @@ def export_job_zip(ekey, dstore):
...
-        inputs['rupture_model'] = dest
+
+    assetcol_csv = None
+    if 'assetcol' in dstore or 'exposure' in dstore:
+        try:
+            [exposure_xml, assetcol_csv] = export_exposure(
+                ('exposure', 'zip'), dstore)
+            inputs['exposure'] = exposure_xml
+        except KeyError:
+            pass
```
micheles: This is terrible. How is it possible for an exposure to be present but not exportable?

openquake/commonlib/readinput.py
```diff
@@ -1158,7 +1158,16 @@ def read_consdict(oqparam, limit_states, perils):
             fnames = [fnames]
         # i.e. files collapsed.csv, fatalities.csv, ... with headers like
         # taxonomy,consequence,slight,moderate,extensive
-        df = pandas.concat([pandas.read_csv(fname) for fname in fnames])
+        dfs = []
+        for fname in fnames:
+            if os.path.exists(fname):
+                dfs.append(pandas.read_csv(fname))
+            else:
+                logging.warning(
```
micheles: No, missing consequences is the normal situation, it should not log a warning

openquake/calculators/export/risk.py
```diff
@@ -954,28 +944,93 @@ def export_job_zip(ekey, dstore):
...
+    if 'ruptures' in dstore:
+        try:
+            csv = extract(dstore, 'ruptures?slice=0&slice=1').array
+            if len(csv.splitlines()) > 2:  # comment + header + data
+                dest = dstore.export_path('rupture.csv')
+                with open(dest, 'w', encoding='utf8') as out:
+                    out.write(csv)
+                inputs['rupture_model'] = dest
+        except Exception:
+            pass
```
micheles: *Errors should never pass silently*
And we need a comment explain when and why this happens

micheles (approved):

PR #11700, opened 2026-08-11
Fix an error in lon array dtype when using gridded surface rupture from a CSV

CB-quakemodel:
`TypeError: loop of ufunc does not support argument 0 of type float which has no callable radians method` when loading a rupture from a CSV which uses `GriddedSurface` 

The dtype of the array was `object` so cleanest fix is to set dtype of float32 when constructing the surface (this is now consistent with the other surface types too).


LanaTodorovic93:
Grazie mille, Chris!

LanaTodorovic93 (approved):

raoanirudh (approved):

PR #11703, opened 2026-08-12
Speed up `truncation_level = 0` case for the conditioned gmfs calculator

raoanirudh:
Introducing a faster route for the case of `truncation_level = 0` for the conditioned gmfs calculator. In this case, the target covariance matrices need not be built as only the conditioned median is required. As a consequence, we can also skip the computation of the YY (target sites) distance matrix, which will not be used in this case. Computation of these matrices is very expensive and avoiding them for the  `truncation_level = 0` where they are not used can lead to significant improvements in runtimes and reduces the memory requirements for such calculations.

For @LanaTodorovic93 's job cole:/home/risk/global_risk_model/scenarios/events/South_America/2026_08_10_M7.4_Colombia/oq/job_wip_stations.ini

master: 14462 seconds
cond-tl-zero: 773 seconds

Addresses the review comment in https://github.com/gem/oq-engine/pull/11441.

### DIFF
--- a/openquake/calculators/event_based.py
+++ b/openquake/calculators/event_based.py
@@ -38,7 +38,7 @@ from openquake.hazardlib.contexts import (
     ContextMaker, FarAwayRupture, get_cmakers)
 from openquake.hazardlib.calc.filters import (
     close_ruptures, magstr, nofilter, getdefault, get_distances, SourceFilter)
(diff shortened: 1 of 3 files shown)


CB-quakemodel (approved):
LGTM

micheles:
Good job, but you need to update the changelog!

Issue #11706, opened 2026-08-14
Generate ShakeMap GMFs from `shake_result.hdf` instead of `grid.xml` and `uncertainty.xml`


raoanirudh:
## Motivation

The engine currently obtains ShakeMap ground motions primarily from `grid.xml` and `uncertainty.xml`. The resulting data contain the conditioned median and conditional total standard deviation at each grid point.

The `Silva&Horspool` implementation then constructs a new covariance matrix by applying spatial and cross-IMT correlation models to those total standard deviations. In particular, the spatial covariance is constructed from `shakemap['std']` in [hazardlib/shakemap/gmfs.py](https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/shakemap/gmfs.py).

This does not distinguish between within-event and between-event residual components. Consequently, a spatial correlation model calibrated for within-event residuals (Jayaram and Baker, 2009) is effectively applied to the total uncertainty, while the spatially common between-event contribution is not sampled separately. A hardcoded cross-IMT correlation model is also applied to the total residuals.

This was a reasonable approximation when only total ShakeMap uncertainty was available through `uncertainty.xml`, but modern ShakeMap products provide the information needed for uncertainty component-aware, ShakeMap-consistent realizations.

As opposed to `grid.xml` and `uncertainty.xml`, the `shake_result.hdf` product contains several additional realization bookkeeping for each IMT, including:

- `C`
- `Sigma_HH_YD`
- `add_uncertainty`
- `sta_per_ix`
- `sta_phi`
- station coordinates

These arrays were extended to all IMTs in `shakemap-modules` 1.1.4 and are described in the source as the information used to reconstruct realizations. See:

- [ShakeMap output documentation](https://ghsc.code-pages.usgs.gov/esi/shakemap/docs2020/manual4_0/ug_products.html#shake-result-hdf)
- [ShakeMap realization-data change](https://code.usgs.gov/ghsc/esi/shakemap-modules/-/commit/2baeeb98f78c1a74c94cf1e8cc7ff4bc6dd56d69)
- [Current HDF storage implementation](https://code.usgs.gov/ghsc/esi/shakemap-modules/-/blob/main/src/shakemap_modules/modeling/storage.py#L196)
- [Current MVN conditioning implementation](https://code.usgs.gov/ghsc/esi/shakemap-modules/-/blob/main/src/shakemap_modules/modeling/mvn.py#L196)

## Proposed change

[...]

Issue #11230, opened 2026-02-19
Refactor correlation modules to support combined spatial-cross-correlation models


raoanirudh:
## Motivation

The engine currently supports:
* **Same-IMT spatial correlation** in [hazardlib/correlation.py](https://github.com/gem/oq-engine/blob/engine-3.24/openquake/hazardlib/correlation.py) – this applies to the within-event residuals
* **Cross-IMT (non-spatial / same-site) correlation** in [hazardlib/cross_correlation.py](https://github.com/gem/oq-engine/blob/engine-3.24/openquake/hazardlib/cross_correlation.py) – these models might be either parametric or empirical, might apply to the within-event residuals or between-event residuals or total residuals, and might have been derived for different intensity measure components (IMCs)

GMF generation not involving conditioning on station data, with both spatial and cross-IMT correlation turned on currently applies:
  * spatial correlation to **within-event** residuals **per IMT**
  * cross-IMT correlation to **between-event** residuals

The conditioned GMF calculator also supports cross-IMT correlation of **within-event** residuals in addition to the above. The full correlation structure is obtained by composing together the cross-IMT correlation of **between-event** residuals, cross-IMT correlation of **within-event** residuals, and spatial correlation to **within-event** residuals, which might be provided by separate models.

Models such as Loth and Baker (2013)[^1], Wang and Du (2013)[^2], Markhvida et al. (2018)[^3], and Du and Ning (2021)[^4], which provide joint spatial-cross-correlation models (for within-event residuals) cannot be readily implemented in the engine given the current separation into two modules handling correlation.

Monteiro and O’Reilly (2026)[^5] provide a comprehensive overview of 45 ground motion correlation models available in the literature, with Python implementations of each available at https://github.com/vitorazevedomonteiro/correlation-models. Ideally, the OpenQuake engine should be able to support *all* of these model classes via a unified and extensible architecture.

## Goals
* Unify correlation handling into a framework that can provide correlation/covariance for:
  * **within-event**, **between-event**, and **total** residuals (as applicable)
* Support:
  * joint spatial-cross correlation models (IMT×site jointly)
  * non-spatial (single-site) cross-IMT correlation, with the conditional spectra calculator being a primary use case (sep. dist. `h = 0`)
  * same-IMT spatial correlation
  * the latter two cases should fall out as special cases of the first
* Keep backward compatibility with existing engine configs and outputs as far as possible .

### 1) Unified correlation framework

[...]

PR #11708, opened 2026-08-14
Correlation modules refactor #1 – architectural base

raoanirudh:
Establishing the architectural base for #11230. 

This PR:
- introduces common interfaces and a registry for spatial, cross-IMT, and future joint spatial-cross-IMT correlation models;
- moves each existing model into its own module with standardized class, function, and file names, bringing the correlation module more in line with the gsim module;
- retains `openquake.hazardlib.correlation` and `cross_correlation` as backward compatibility modules, to be retained for a few releases at least until the new correlation module is well tested and functioning efficiently;
- standardizes job configuration temporarily on (might change again in successive PRs which will bring in joint spatial-cross-correlation models):
  - `spatial_correlation_model`
  - `spatial_correlation_params`
  - `cross_imt_correlation_model`
  - `cross_imt_correlation_params`
- migrates maintained job files and documentation while preserving legacy aliases;
- makes 'ShakeMap to gmfs' use the same named-correlation-model configuration as regular scenario calculations, with omission disabling correlation, instead of the current `yes`|`no`|`full` syntax which deviates from regular scenario calculations;
- adds `BakerCornell2006` to preserve the historical ShakeMap cross-IMT behavior for now, which seems to be the provenance of the current hard-coded cross-correlation model in [hazardlib/shakemap/gmfs.py](https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/shakemap/gmfs.py), to be updated soon in upcoming PRs related to #11706.

Existing sampling algorithms, truncation level behavior, and numerical results are preserved in this PR.

This PR deliberately does not yet implement a joint spatial-cross-IMT model or new efficient sampling algorithms; those will follow in smaller PRs to keep reviews manageable.


openquake/commonlib/oqvalidation.py
```diff
@@ -1075,7 +1077,13 @@ class OqParam(valid.ParamSet):
...
                'quantile_hazard_curves': 'quantiles',
                'mean_hazard_curves': 'mean',
-               'max_hazard_curves': 'max'}
+               'max_hazard_curves': 'max',
+               'ground_motion_correlation_model':
+               'spatial_correlation_model',
+               'ground_motion_correlation_params':
+               'spatial_correlation_params',
+               'cross_correlation': 'cross_imt_correlation_model',
+               'spatial_correlation': 'spatial_correlation_model'}
```
micheles, line 1087: I like that you are keeping backward compability

micheles:
Overall I like the plan and also the implementation. LGTM if all tests are green. Good job!

CB-quakemodel (approved):
LGTM! I think the changelog will need a quite substantial entry!

PR #11712, opened 2026-08-18
Correlation modules refactor #2 – residual-aware joint correlation

raoanirudh:
## Summary

Part of #11230. Following the architectural base laid in #11708, this PR now introduces residual-specific correlation configuration and further architectural support needed for combined spatial and cross-IMT correlation models.

The new settings are:

- `within_event_correlation_model` and `_params`
- `between_event_correlation_model` and `_params`
- `total_residual_correlation_model` and `_params`

The settings are named after the residual component being correlated rather than the dimensions of the correlation. “Spatial” and “cross-IMT” describe covariance structure, but a model may handle both jointly, so separate settings would implicitly assume separability and would not represent models such as Loth and Baker (2013) cleanly.

The names `within_event`, `between_event`, and `total_residual` instead identify where a model is applied and the residual component for which it should be calibrated. A within-event correlation model may be spatial-only or jointly spatial–cross-IMT, while between-event and total-residual correlation models generally describe cross-IMT correlation. We could be more explicit by using `within_event_spatial_cross_imt_correlation_model`, `between_event_cross_imt_correlation_model`, and `total_residual_cross_imt_correlation_model`, but this seems excessive and hopefully the chosen settings should be clear enough.

This PR also:

- validates correlation models against the residual component for which they were calibrated
- adds a covariance (and factorization) interface for joint spatial–cross-IMT models
- enables joint within-event sampling across sites and IMTs
- supports the rectangular covariance blocks required by station-conditioned calculations
- maintains the existing spatial-only sampling path and aliases
- makes correlation choices explicit in regular gmfs, conditioned gmfs, conditional-spectrum, and ShakeMap workflows, so that model choices and residual components are no longer hardcoded or implicitly assumed

## ShakeMap to gmfs limitation

The ShakeMap to gmfs workflow still applies cross-IMT correlation to total residuals, as in the original implementation. This is not theoretically correct for ShakeMap-consistent sampling, but the behavior is intentionally retained in this PR to avoid changing legacy numerical results during this intermediate refactoring stage.

The historical choices of Jayaram and Baker (2009) for spatial correlation and Baker and Cornell (2006) for total-residual cross-IMT correlation are now required to be explicitly configured instead of being hardcoded as in the original implementation.

A follow-up will replace both with the joint spatial–cross-IMT model of Loth and Baker (2013), which is the within-event correlation model used internally by ShakeMap. The broader transition of the ShakeMap workflow is tracked in #11706.

## Critical items for review

Although the diff for this PR includes many files, most changes are simply configuration renames in QA test jobs and documentation. The critical files for review include:

1. residual-component configuration, validation, and legacy aliases in `oqvalidation.py`
2. the covariance/factorization interfaces in `correlation_models/base.py`
3. joint sampling and preservation of the spatial-only correlation option in `calc/gmf.py`
4. joint and rectangular covariance handling in `calc/conditioned_gmfs.py`
5. routing of total-residual correlation in conditional spectra and the intentionally unchanged ShakeMap to gmf behavior

## What to expect in following PRs

Subsequent PRs will incrementally:

- introduce Loth and Baker (2013) as the first joint spatial–cross-IMT model
- add further correlation models individually with support from https://github.com/vitorazevedomonteiro/correlation-models and https://github.com/vitorazevedomonteiro/oq-engine
- add dedicated documentation pages covering correlation concepts, model selection, configuration, assumptions, and limitations
- separately update the ShakeMap workflow under #11706, replacing the retained legacy total-residual treatment with Loth and Baker (2013)
- add model tests and performance benchmarks in `oq-risk-tests`
- audit each correlation model’s supported IMTs against its original calibration and remove legacy proxy mappings or extrapolations, such as treating PGV as SA(1.0), or at least add clear warnings in such cases that the selected model is being used outside its original domain
- evaluate efficient and scalable sampling methods including circulant embedding for large simulations


CB-quakemodel (approved):
Thanks Anirudh, I have only reviewed the outlined priority files, but LGTM.

Also, I agree that the use of keys in the job file such as `within_event_spatial_cross_imt_correlation_model` would be too cumbersome, and I prefer the naming convention you propose.

PR #11713, opened 2026-08-18
Correlation modules refactor #3 – early documentation

raoanirudh:
This is a documentation-only PR linked to #11230. 

Adds the basic documentation structure for ground-motion correlation models, including some dedicated pages:

- an underlying-science page covering 
  - residual components
  - spatial and cross-IMT dependence
  - joint covariance, conditioning
  - separability assumptions
  - computational cost
  - the current ShakeMap to gmfs legacy limitation
- a contributor guide for implementing and deterministically testing new models, closely based on the guidelines page for implementing new GSIMs
- API documentation for the correlation interfaces, registry, and existing models
- updated conditioning and GSIM documentation with appropriate cross-references

Previous PRs in this sequence include:
- #11708
- #11712

PR #11714, opened 2026-08-18
Correlation modules refactor #4 – Loth & Baker (2013)

raoanirudh:
## Loth and Baker (2013) spatial-cross-IMT correlation model

Part of #11230. This PR adds `LothBaker2013`, the first combined spatial and cross-IMT correlation model built on the refactored residual-aware correlation architecture introduced previously in
- #11708
- #11712 and documented in
- #11713. 

This implementation:

- Computes within-event correlations jointly across sites and spectral periods using Equation 42 of [Loth and Baker (2013)](https://doi.org/10.1002/eqe.2212)
- Uses the corrected coefficient tables from the [2020 erratum](https://doi.org/10.1002/eqe.3233)
- Follows the authors' [current Matlab implementation](https://github.com/bakerjw/GMMs/blob/master/correlations/lb_2013_spatial_corr.m), including:
  - the April 2022 interpolation refinement that preserves the diagonal ridge and avoids saddle points
  - the revised, full-precision B3 coefficients added in May 2022 to preserve positive definiteness
- Supports 5%-damped SA from 0.01 to 10 seconds, matching the model's calibrated domain
- Produces rectangular, IMT-major correlation blocks using `float64`
- Registers the model as a combined spatial and cross-IMT model for within-event residuals

The focus of this PR is initial model implementation and verification of scientific correctness. Performance and memory improvements will be addressed separately.

## Verification

Deterministic verification tests cover:

- Values generated independently from the authors' Matlab implementation
- Near-diagonal and off-diagonal interpolation
- Symmetry, unit diagonal, and positive definiteness
- Rectangular IMT-major block ordering
- Covariance construction and factorization
- Checking unsupported IMTs, damping values, periods, residual components, and invalid distances

The reference values were generated on 18 August 2026 by running the authors' [Matlab implementation](https://github.com/bakerjw/GMMs/blob/d7c3241/correlations/lb_2013_spatial_corr.m) at revision `d7c3241` with GNU Octave 8.4.0, independently of hazardlib.

The OpenQuake implementation in this PR agrees with the authors' Matlab implementation for all reference cases to floating-point precision. The largest observed difference was approximately `2.8e-17`.

## Comparison with USGS ShakeLib

The table below illustrates differences between the authors' current implementation and [USGS ShakeLib](https://code.usgs.gov/ghsc/esi/esi-shakelib/-/blob/main/src/esi_shakelib/correlation/loth_baker_2013.py):

| T1 (s) | T2 (s) | Distance (km) | Author Matlab and this PR | USGS ShakeLib |
| ------: | ------: | ------------: | ------------------------: | --------------: |
| 1.00 | 2.00 | 10.0 | 0.320641252023 | 0.320641252023 |
| 0.15 | 1.00 | 12.5 | 0.135311973068 | 0.135311973068 |
| 0.15 | 0.18 | 25.0 | 0.149424292835 | 0.145002397537 |
| 0.15 | 0.18 | 0.0 | 0.930999913169 | 0.885000000000 |
| 0.35 | 3.00 | 7.0 | 0.163501415622 | 0.168955195471 |

ShakeLib provides a valid implementation of the published formulation. It uses the rounded tables provided in the erratum and general bilinear interpolation. The paper permits linear interpolation but does not prescribe the later ridge-preserving procedure.

The differences with the ShakeLib implementation above likely arise because the authors refined their Matlab implementation in 2022 with:

- Special interpolation near the diagonal
- Local piecewise-linear interpolation away from the diagonal
- A revised, higher-precision B3 matrix

## Notes

**Note 1**: The implementation in this PR follows the current author-maintained implementation as its primary reference since the values computed by the engine can then be directly compared against the authors' own implementation.

**Note 2**: PGA is supported using SA(0.01) as its correlation proxy, following the same convention in ShakeLib, since this model is likely going to be used for calculations involving PGA. ShakeMap also maps PGV and MMI to SA(1.0) operationally, but unlike the PGA ≅ SA(0.01) approximation, that lacks a direct physical equivalence and is not supported in this PR. Calculations involving PGV _and_ SA/PGA must use a different correlation model that natively supports all of them.


CB-quakemodel (approved):
LGTM

PR #11720, opened 2026-08-22
Avoid animations/transitions in playwright UI tests…

ptormene:
…to prevent Bootstrap modal backdrop timeouts.

Sometimes (rarely but annoyingly) the Playwright UI tests fail due to race conditions (e.g. https://github.com/gem/oq-engine/actions/runs/32467349360/job/96726683861)
Let's see if this change will prevent those random failures.

See also https://github.com/gem/oq-engine/actions/runs/32583462483


ptormene:
It looks like it had undesired side-effects, so it is not a viable solution.

Issue #11, opened 2010-12-25
bin/openquake --server still tries to run memcached


larsbutler:
This got missed in the transition from memcached to redis.

PR #11724, opened 2026-08-25
Correlation modules refactor #11 – conditioned GMFs calculator

raoanirudh:
## Correct joint conditioning of ground-motion fields across sites and IMTs

Part of #11230. This PR establishes correct joint conditioning of ground-motion fields across sites and intensity measures. It allows the newly added spatial-cross-IMT within-event correlation models to be used without resorting to the historical per-IMT, separable approximation. It also fixes several correctness issues uncovered while checking the conditioned-GMF workflow and substantially expands the test coverage. It also begins to lay the groundwork for efficient memory management and performance improvements for calculations involving conditioning on observations.

### Motivation

The conditioned-GMF calculator historically processed each target IMT independently. Cross-IMT dependence was approximated by multiplying a spatial correlation with a separate cross-IMT coefficient, and realizations for different target IMTs were not drawn from one joint posterior. The reasons for this were both performance considerations, as well as the absence of any joint spatial-cross-IMT correlation models in the engine.

With the recent addition of joint spatial-cross-IMT models such as [Loth and Baker (2013)](https://github.com/gem/oq-engine/pull/11714), [Markhvida et al. (2018)](https://github.com/gem/oq-engine/pull/11716), or [Du and Ning (2021)](https://github.com/gem/oq-engine/pull/11723), the historical approximations are no longer necessary, and the calculator is rewired in this PR to be able to use the joint spatial-cross-IMT models that are now available in the engine. The existing approach also made it difficult to introduce scalable sampling algorithms because the conditioning logic and historical sampling path are tightly coupled.

During this work, several other pre-existing correctness issues were identified:

- A regression introduced during a 2023 refactor (#9093) caused multi-IMT calculations to reuse the GSIM mean, tau, and phi of the first target IMT for all other IMTs — also flagged by Jia-Sheng Hung from Taiwan Earthquake Model.
- Station observation-error variances could be assigned to the wrong covariance diagonal entries because a column vector was passed through NumPy broadcasting.
- PGV was always excluded from conditioning, because previously the engine did not have any correlation models that natively supported PGV. Now that there are correlation models supporting PGV, this needs revisiting.
- Chunked site collections were sliced using global site IDs rather than local indices.

These issues are corrected here and covered by focused regression tests.

### Joint spatial-cross-IMT conditioning architecture

For a genuine spatial-cross-IMT within-event model, the calculator now constructs one station system containing all usable observed IMTs and station sites. This system combines:

- the model-defined joint within-event covariance
- the GSIM-provided within-event standard deviations
- station observation-error variances, and
- the between-event covariance represented in the smaller latent IMT space

The target IMTs and sites are then treated as a single IMT-major Gaussian vector. The calculator constructs the total station, target-to-station, and target covariance blocks and applies the standard conditional multivariate-normal equations.

This is algebraically equivalent to the partitioned formulation in equations B8–B9 and B16–B17 of Engler et al. (2022), generalized to multiple jointly correlated target IMTs. A deterministic matrix test independently evaluates the Engler partition and confirms that it produces the same posterior mean and covariance as the new joint formulation.

The historical spatial-only path remains available and retains its existing per-IMT and finite truncated-normal behavior, but should be removed in a dedicated follow-up PR. This was not done in the current PR since the numbers for several tests would change.

### Bugfixes and addition of numerical safeguards

The PR additionally:

- computes target GSIM statistics separately for every target IMT (fixing the bug described above)
- includes PGV when supported, while continuing to exclude MMI
- applies each station uncertainty to its corresponding covariance entry
- handles incomplete station records using one consistent IMT-major mask
- verifies that target-to-station covariance blocks are compatible with the station system
- rejects finite truncated joint sampling explicitly rather than reverting to an incorrect per-IMT calculation

Since this module has undergone several changes since its original implementation, the definitions of the notations used in the module have been updated. The module documentation now also provides an overview of the basic calculation flow.

### Memory and performance groundwork

Deterministic joint calculations no longer construct the complete target-to-station or target-to-target matrices. The exact posterior mean is evaluated in target-site chunks.

The conditioned-calculation now distinguishes between:

- chunked deterministic conditioning
- dense stochastic joint conditioning
- station-scale covariance systems
- target-to-station covariance blocks
- random samples, and
- final float32 GMF arrays

It also initializes the event and realization arrays before estimating memory, so the estimate uses the actual number of fields instead of the rougher approximation used previously.

Oversized calculations now fail fast before allocating the full posterior rather than exhausting system memory.

For deterministic joint conditioning, the chunk size is derived from the configured `memory.conditioned_gmf_gb` workspace budget. The planner first reserves memory for the station covariance system and final result arrays, divides the remaining budget across GSIM tasks, and then determines how many complete target sites can fit while accounting for three temporary float64 covariance arrays. The calculated block size is passed to the worker, ensuring that memory estimation and execution use exactly the same plan.

If even a single target-site chunk cannot fit, the calculation is rejected with an error reporting the required and configured memory. The user can then reduce the number of target or station sites or adjust `memory.conditioned_gmf_gb` for the machine running the calculation.

Note that the configuration `memory.conditioned_gmf_gb` represents the workspace available to conditioned-GMF calculations, not the machine’s total physical memory.

### QA changes

Expected results are updated for four existing conditioned QA cases. These changes are intentional consequences of the bug fixes described above:

1. using the correct mean, tau, and phi for each target IMT, and
2. assigning unequal station uncertainties to their correct observations

### Current limitations

- Truncated multivariate-normal sampling is not yet implemented for joint spatial-cross-IMT models
- Stochastic joint conditioning still constructs and factorizes the dense target posterior covariance – worth revisiting in future PRs
- The joint path currently returns the combined posterior field rather than preserving separate posterior within- and between-event fields
- The historical per-IMT separable path remains temporarily for compatibility and to avoid bloating this PR further, but should be removed in a dedicated follow-up PR

### What should come next

This work also prepares for scalable conditional-simulation methods. The immediate next step should be to investigate conditional simulation approaches for replacing construction of the dense posterior covariance with an unconditional joint field followed by a station-based correction. We'd need to evaluate scalable methods for generating that unconditional field, including circulant embedding as proposed by [Bailey et al. (2021)](https://doi.org/10.1002/sta4.446), alongside related memory-efficient approaches such as the successive conditional simulation algorithm of [Verros et al. (2017)](https://doi.org/10.1016/j.cageo.2016.11.004).

Performance could be evaluated using an event such as 2014 M6.0 South Napa with a smaller footprint, and a stretch case such as the 2023 M7.8 Kahramanmaraş earthquake with a much larger footprint in `oq-risk-tests`, targeting approximately:

- 10,000 sites for the smaller M6.0 event up to 250,000 sites for the M7.8 event
- four or five IMTs, including PGV where supported, and
- 500–1,000 conditioned ground-motion field realizations

Once the joint sampling path is scalable and stable, the historical per-IMT conditioning implementation and its separable approximation can be removed.


CB-quakemodel (approved):
LGTM

micheles:
Impressive work

PR #11726, opened 2026-08-28
Logic trees for site amplification model

CB-quakemodel:
This PR adds the ability to specify amplification model logic trees using the regular NRML XML logic. The user can specify an amplification LT XML file for the `amplification_csv` key in the .ini within which each branch points to a different amplification model CSV.

Adding epistemic uncertainty on the amplification through a proper logic tree is important for supporting recent site-response methodologies (e.g., Rodriguez-Marek et al. 2021 - see their Figure 6 for a simple visualisation). These recent methodologies produce multiple weighted AF branches (i.e., logic trees of amplification models/factors), and each branch may carry its own σ_AF per IMT to represent within-branch aleatory variability (i.e., we have several amp factor branches each with aleatory variability, akin to how we have multiple weighted GMMs each with sigma). Properly incorporating sigma to the AF requires the convolution approach rather than scalar factors applied in a ModifiableGMPE. The engine already supports the convolution method for a single amplification model (with uncertainty) - this PR extends it to a logic tree of such models, helping to support advanced site-response workflows.

Each branch must define AFs for all sitecol ampcodes and for all IMTs in `intensity_measure_types_and_levels`, but beyond that all parameters (the actual AF and sigma values, the rock-IML `level` grid, and the `from_mag`/`from_rrup` grids) are permitted to differ across branches so users can represent alternative discretisations and value sets as epistemic choices for site amplfication.

Full enumeration and sampling of amp branches are supported. Under full enumeration realizations become the product of `R_SSC × R_GMM × R_amp`; when sampling all three legs are drawn `num_samples` times independently (the amp leg uses the same `lt.random` / `lt.sample` methods as the SSC and GSIM parts so both `early_weights` and `late_weights` are supported).

A new hazardlib module `amp_lt.py` provides `AmplificationLogicTree`, which parses and validates the amp LT XML. The parsed branches are held by `AmplificationModel` (in `site_amplification.py`), a single class covering both the plain single-CSV case and the amp-LT case: it holds the per-branch metadata (names, weights, DataFrames, filenames), the `Amplifier` instances built from those DataFrames, and the per-realization branch index, and provides `get_realizations()` for full enumeration and `sample()` for sampling.

QA tests are added (`logictree/case_26`) for a 3-branch amp LT (low, med, high AF) covering 2 IMTs (PGA + SA(0.2)) each with 2 rock levels plus `from_mag`/`from_rrup` columns:

- Classical with full enumeration
- Classical with sampling

Example of results for the classical with full enumeration:

{32BA5CE3-ABC7-47DE-8A80-996309B2619E}

---------------------------------------------------------------

Extensive unit tests are also added for the new `amp_lt.py` hazardlib module which test:

- The `AmplificationLogicTree` XML parser rejects non-`amplificationModel` branchsets
- The `AmplificationLogicTree` XML parser rejects weights that don't sum to 1
- The `AmplificationLogicTree` XML parser rejects duplicate branchIDs within a branchset
- The `AmplificationLogicTree` XML parser rejects a `<logicTree>` with no `amplificationModel` branchset
- `AmplificationLogicTree.is_amp_lt` returns True for a valid amplification LT XML and false for an invalid amplification LT XML
- The `AmplificationModel` container assigns every branch a unique BASE183 short character
- `FullLogicTree.__toh5__`/`__fromh5__` roundtrip preserves the amp LT metadata
- `FullLogicTree.get_realizations` under full enumeration builds the `SSC × GMM × amp` product with correctly multiplied weights and a three-leg path
- `FullLogicTree.get_realizations` under `early_weights` sampling gives uniform `1/num_samples` weights
- `FullLogicTree.get_realizations` under `late_weights` sampling draws branches by weight

-----------------------------------------------------

Documentation is added for this feature. Also, support is added for `oq show rlz <rlz_id>`, `oq show branches` and `oq show branchsets`:

{4997425A-2A36-4BB0-B62F-9B7188F1F3D2}

---------------------------------------------------------------

Currently only supported in classical calculations with the convolution approach. Errors are raised if calculation mode is not classical, and if amplification method is not convolution (i.e., kernel).


openquake/hazardlib/amp_lt.py
```diff
@@ -0,0 +1,194 @@
...
+from openquake.hazardlib import InvalidFile, nrml
+from openquake.hazardlib import lt
+from openquake.hazardlib.lt import Realization
+
+
+F32 = numpy.float32
+
+
+def _iter_branchsets(lt_node):
+    # Tolerate both flat and older <logicTreeBranchingLevel>-wrapped layouts
```
micheles: I am not sure if we want to support `logicTreeBranchingLevel`; it is deprecated in spirit, why to support it for new features like the amplification lt?

CB-quakemodel: I have now addressed this (plus the handling of `amplifier` and `amplifier_lt` in a single container object called `AmplificationModel` instead)

micheles (approved):

micheles:
Please update the changelog!

PR #11734, opened 2026-09-04
Correlation modules refactor #13 – unconditioned correlated GMFs using circulant embedding

raoanirudh:
## Efficient (unconditioned) spatially-cross-correlated GMF simulation using circulant embedding

Related to #11230 and #10833. This PR connects the circulant-embedding implementation introduced in #11728 to unconditioned scenario and event-based ground-motion field (GMF) generation.

Previously, as described in #11728, correlated GMFs required constructing and factorizing a dense covariance matrix whose dimensions grew with the number of sites and IMTs. This made realistic calculations prohibitively expensive and would inevitably exhaust available memory before any GMFs were generated. Circulant embedding (CE) instead represents a stationary correlation model on a periodic grid and uses FFTs with small cross-IMT spectral factors, avoiding the dense site-by-site covariance matrix. CE takes advantage of the fact that for regular grids (equidistant points), assuming that the spatial correlation between points depends solely on the distance between them, the correlation matrix will have a highly regular structure, in fact forming a Block Toeplitz matrix with Toeplitz blocks.

### Implementation

Large regular geographic grids (eg. `n_sites > 1,000`) are detected automatically and mapped to a locally projected lattice. Circulant embedding is currently enabled explicitly for the stationary `JayaramBaker2009` and `DuNing2021` models, whereas other models continue to use the existing dense path when that path fits within the configured memory budget.

The new `memory.correlated_gmf_gb` setting defines a per-worker memory budget. Both the FFT operations and the resulting GMF arrays are divided into batches, which are streamed incrementally to the datastore rather than assembling the complete `site × IMT × realization` cube in memory, so that we avoid blowing up the available memory at any step. Calculations that cannot fit within the budget are made to fail early with an informative message.

Separate reproducible random streams are used for within-event residuals, between-event residuals, and amplification. Changing the batch size therefore should not change seeded realizations.

The `truncation_level_within` values are honoured approximately by truncating the FFT draws before applying circulant embedding, somewhat similar to the existing approach to handling within event residual truncation. The resulting correlated field is thus not an exactly component-wise truncated multivariate normal, as individual correlated residuals may exceed the specified bound. Exact multivariate truncation at these dimensions would not be computationally feasible and is outside this PR’s scope. The approximate truncation should still achieve the intended objective of `truncation_level_within` in spirit, even though not exact.

Efficient handling of streaming arrays is outside my expertise, so I'm not sure if any changes are needed in [baselib/parallell.py](https://github.com/gem/oq-engine/blob/master/openquake/baselib/parallel.py) or [baselib/zeromq.py](https://github.com/gem/oq-engine/blob/master/openquake/baselib/zeromq.py) to ensure these calculations run smoothly on different kinds of machines @micheles?

### Benchmark test cases

Two performance benchmarking test cases have been added to `oq-risk-tests` (private repo to avoid bringing in potentially expensive test cases into the engine repo), each generating 500 unconditioned spatially-cross-correlated fields for PGA, SA(0.3), SA(1.0), and SA(3.0) with `DuNing2021` correlation and a truncation level of 3:

- the 2014 M6.0 South Napa scenario, with 8,864 sites within 50 km of the rupture
- the 2023 M7.8 Kahramanmaraş scenario, with 203,775 sites within 180 km of the rupture

Both calculations complete successfully on my MacBook Pro (36 GB RAM). The South Napa scenario completes in 30 seconds, while the Kahramanmaraş calculation also runs to completion without running out of memory in under 2 minutes. Previously, it was simply not feasible to run the 2023 Kahramanmaraş scenario to completion even on machines with 128 GB memory. Both of these test cases are meant to be manual performance benchmarks and are not registered for automatic CI execution. 

We are not testing the correctness of the numbers in these two test cases, as these are purely performance benchmarks, and should ideally be run using identical environments for each run to ensure comparability of runtimes and memory consumption across runs.

In terms of correctness, we can only undertake statistical validation of a large number of generated fields and check that the residual means are approximately zero across realizations, and that the marginal variances have the expected values. This is not done in the current PR.

### Scope and follow-up

This PR addresses _unconditioned_ correlated GMFs on _regular_ grids. Station-conditioned calculations would be the next step. Irregular site collections will likewise require a follow-up PR, perhaps looking at the CE adaptation method proposed by Bailey et al. (2022) as described in #11728.

### DIFF
--- a/openquake/baselib/parallel.py
+++ b/openquake/baselib/parallel.py
@@ -544,7 +544,12 @@ def safely_call(func, args, task_no=0, mon=dummy_mon):
         args += (mon,)
     sentbytes = 0
     if isgenfunc:
-        with Socket(mon.backurl, zmq.PUSH, 'connect') as zsocket:
+        # Keep a worker from queuing many large generator results faster
+        # than the master can persist them. The synchronous ``no`` mode
+        # cannot use backpressure because it sends before reading begins.
+        hwm = 1 if getattr(mon, 'distribute', 'no') != 'no' else None
+        with Socket(
+                mon.backurl, zmq.PUSH, 'connect', hwm=hwm) as zsocket:
             it = func(*args)
             while True:
                 res = Result.new(next, (it,), mon, sentbytes)
@@ -864,6 +869,7 @@ class Starmap(object):
             if self.distribute == 'slurm':
                 self.init_slurm()
         dist = 'no' if self.num_tasks == 1 else self.distribute
+        self.monitor.distribute = dist
         if dist not in ('no', 'threadpool'):
             pickled = isinstance(args[0], Pickled)
             if not pickled:
--- a/openquake/baselib/parallel.py
+++ b/openquake/baselib/parallel.py
@@ -544,12 +544,7 @@ def safely_call(func, args, task_no=0, mon=dummy_mon):
         args += (mon,)
     sentbytes = 0
     if isgenfunc:
-        # Keep a worker from queuing many large generator results faster
-        # than the master can persist them. The synchronous ``no`` mode
-        # cannot use backpressure because it sends before reading begins.
-        hwm = 1 if getattr(mon, 'distribute', 'no') != 'no' else None
-        with Socket(
-                mon.backurl, zmq.PUSH, 'connect', hwm=hwm) as zsocket:
+        with Socket(mon.backurl, zmq.PUSH, 'connect') as zsocket:
             it = func(*args)
             while True:
                 res = Result.new(next, (it,), mon, sentbytes)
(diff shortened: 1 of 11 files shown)


raoanirudh:
5 / 500 simulated unconditioned spatially-correlated PGA fields for the 2014 M6.0 South Napa earthquake
PGA_0
PGA_1
PGA_2
PGA_3
PGA_4

raoanirudh:
5 / 500 simulated unconditioned spatially-correlated SA(0.3) fields for the 2023 M7.8 Kahramanmaraş earthquake

SA(0 3)_0
SA(0 3)_1
SA(0 3)_2
SA(0 3)_3
SA(0 3)_4

micheles:
@raoanirudh please send me the biggest calculation, I want to run it on a large machine with 160 cores to check the performance and the data transfer and also possible issues with parallelism (i.e. we must make sure that the random
numbers generated are the same independently from the number of cores used, particularly for event_based_risk, if we have and example for it).

micheles (approved):

PR #11737, opened 2026-09-05
Correlation modules refactor #15 – station-conditioned correlated GMFs

raoanirudh:
## Efficient station-conditioned spatially-cross-correlated GMF simulation using circulant embedding and local Kriging

Related to #11230 and #10833. #11734 connects the circulant-embedding implementation introduced in https://github.com/gem/oq-engine/pull/11728 to unconditioned ground-motion field (GMF) generation. #11735 implemented local Kriging to obtain unconditioned gmf values at off-grid station locations.

This is the second of two PRs stacked on top of #11734 following #11735, extending circulant embedding from unconditioned regular-grid fields to station-conditioned GMFs.

The existing conditioning algorithm in the engine constructs a posterior covariance whose dimensions are the number of target sites multiplied by the number of IMTs. Storing this matrix requires quadratic memory, and factorizing it has cubic computational cost. This makes station-conditioned GMF sampling infeasible for realistic regional calculations.

This PR replaces that dense target-posterior operation, for eligible regular grids and joint correlation models, with paired circulant-embedding realizations and Matheron substitution.

### How the scalable conditioning works

The method first generates an unconditioned realization at the target grid and corresponding values at the station locations:

- circulant embedding generates the large joint within-event field on the regular grid ([#11734](https://github.com/gem/oq-engine/pull/11734))
- exact extraction or local Kriging provides paired within-event values at the stations ([#11735](https://github.com/gem/oq-engine/pull/11735))
- one low-dimensional between-event residual is shared by every site and correlated across IMTs using the configured between-event cross-IMT correlation model
- station observation uncertainty is added where specified

The target and station values must be generated from the same random realization. If $U_Y$ and $U_D$ are these paired zero-mean prior values at the targets and stations, respectively, the conditioned field is

```math
Y_{|D}
=
\mu_Y + U_Y
+
\Sigma_{YD}\Sigma_{DD}^{+}
\left(\zeta_D-U_D\right).
```

Here, $\zeta_D$ contains the observed station residuals, $\Sigma_{YD}$ is the target-to-station prior covariance, and $\Sigma_{DD}$ is the station covariance including the within-event, between-event, and observation-error contributions.

Essentially, the calculation generates an unconditioned field, compares its simulated station values with the observations, and propagates that difference across the target grid using the user-specified covariance model. This is also called Matheron substitution in spatial statistics / geostatistics. It produces the conditional Gaussian field without constructing the dense target posterior covariance.

The target-to-station regression weights are built once in memory-bounded site chunks, retained in `float64`, and reused for every realization. The realizations are then generated, conditioned, converted to GMFs, and returned in batches governed by `memory.conditioned_gmf_gb` and `memory.max_gmvs_chunk` (first stab, there might be more efficient ways of doing this!).

### Efficiency of the circulant embedding + local Kriging approach

The 2023 M7.8 Kahramanmaraş benchmark scenario added in #11734 contains 112 stations and four IMTs. For a rounded target grid of ~250,000 sites, the joint target field contains $4 \times 250{,}000 = 1{,}000{,}000$ random variables, while the station data contain $4 \times 112 = 448$ observed variables.

| Operation | Dimensions | `float64` storage |
|---|---:|---:|
| Spatial covariance for one IMT | 250,000 × 250,000 | ~ 500 GB |
| Joint target covariance for four IMTs | (250,000 × 4) × (250,000 × 4) | ~ 8 TB |
| Joint target-to-station covariance | 1,000,000 × 448 | ~ 3.6 GB |
| Local station weights using 64 grid cells | (112 × 4) × (64 × 4) | ~ 1 MB |

The 8 TB estimate is for the joint covariance matrix alone. Its factor and numerical workspaces would require still more memory, which is why this calculation was doomed to fail by running out of memory on any machine.

Circulant embedding avoids constructing the massive target-by-target covariance matrix, replacing it with FFTs and small cross-IMT spectral matrices. Local Kriging then helps avoid coupling every off-grid station to the _complete_ target grid when generating the paired station prior. Each station uses _only 64_ nearby grid cells, reducing that particular operation from approximately 3.6 GB of covariance values to about 1 MB of retained local weights, plus a roughly 0.5 MB neighborhood covariance, with barely any loss of usable information.

The final Matheron correction still requires global target-to-station regression weights. These are constructed in memory-bounded target chunks and then retained for reuse across all realizations. Thus, local Kriging removes an additional full-grid station coupling, although it does not eliminate the scientifically necessary target-to-station conditioning relationship.

### Relationship to the Engler et al. (2022) approach

The existing OpenQuake engine implementation for station-conditioned GMFs (#8318) follows the partitioned formulation of Engler et al. (2022)[^1]. It first derives the posterior between-event residual and then kriges the remaining within-event station residuals.

If $\widehat H$ is the posterior mean of the normalized between-event residual, its target contribution is $T_Y\widehat H$. This is the target-specific form of the quantity described as the nominal bias. The partitioned posterior mean can be written as

```math
\mu_{Y|D}
=
\mu_Y
+
T_Y\widehat H
+
\Sigma_{W_YW_D}\Sigma_{W_DW_D}^{+}
\left(\zeta_D-T_D\widehat H\right).
```

The joint formulation used here instead writes

```math
\mu_{Y|D}
=
\mu_Y+\Sigma_{YD}\Sigma_{DD}^{+}\zeta_D,
```

with

```math
\Sigma_{YD}
=
\Sigma_{W_YW_D}
+
T_Y\Sigma_{HH}T_D^{\mathsf T}.
```

Expanding the joint regression recovers the between-event nominal-bias term and the within-event Kriging term in the partitioned expression. Under the joint-Gaussian assumptions, the two formulations therefore produce the same total posterior distribution. 

The main organizational difference is that the new CE + local Kriging method does not retain separately labelled posterior between- and within-event fields. It directly samples their jointly conditioned total contribution.

This also follows the broad computational approach described by Engler et al. (2025)[^2]: generate an unconditioned within-event realization using circulant embedding and then condition that realization on the observations through Kriging. The current PR generalizes this to multiple jointly correlated IMTs.

One of the advantages of the Engler et al. (2022) partitioning is that it can report the posterior event term, its uncertainty and the nominal bias separately from the within-event field. That can be quite useful for diagnostics, comparing the inferred event bias between GMMs, or for applications that want to fix or sample the event term independently. 

The CE approach, on the other hand, provides a major performance advantage. Its limitations are the regular-grid and stationarity requirements, the local approximation for off-grid stations, and the absence of separately retained posterior between- and within-event realizations.

Given the valuable diagnostic utility of the Engler et al. (2022) partitioning, we continue to report the posterior mean and standard deviation of the normalized between-event residual and the corresponding target-averaged nominal-bias statistics. These diagnostics require only the small station covariance system and do not change the sampled GMFs, random-number sequence, or memory scaling. Separate posterior component realizations are not retained, but could be added later if needed.

### In this PR

The PR adds:

- a small joint station covariance system containing the within-event, between-event, and observation-error contributions
- paired target and station prior draws using CE and local Kriging
- a dense joint implementation retained as the reference path for small calculations
- reusable target-to-station regression weights constructed in bounded chunks
- memory-aware streaming of conditioned GMF realizations
- calculator integration using the existing task and datastore interfaces, without modifying `parallel.py` or `zeromq.py`
- underlying-science and API documentation covering CE, local kriging, Matheron substitution, nominal-bias equivalence, and current limitations

The scalable path introduced in this PR is currently enabled for sufficiently large regular grids using the `DuNing2021` spatial-cross-correlation model. Other models continue to use the old path until their structured embeddings have been validated. 

**Note**: Joint finite truncated multivariate-normal sampling is not yet supported.

### Verification and benchmarks

The numerical tests verify the paired target-and-station prior covariance, shared between-event terms, observation error, colocated stations, Matheron substitution, posterior means, and invariance to covariance and realization batch sizes. The dense joint implementation is assumed to provide the exact values for testing the scalable formulation.

My 36 GB MacBook Pro is successfully able to generate 500 conditioned GMFs for four IMTs (for a single GMM in each case):

- the 2014 M6.0 South Napa scenario, with approximately 9,000 sites, completed in about 20 seconds
- the 2023 M7.8 Kahramanmaraş scenario, with approximately 204,000 sites, completed in about 80 seconds

These cases, especially the latter, clearly demonstrate that the method can handle calculation sizes for which constructing the dense posterior covariance was simply impractical.

The review could particularly check the memory accounting and batching.

### Possible future reorganization

`conditioned_gmfs.py` now contains both the historical per-IMT partitioned algorithm based on Engler et al. (2022) and the new joint Gaussian implementation with dense and CE sampling backends based on Bailey et al. (2022)[^3]. A later structural PR could separate out common conditioned calculation orchestration, the Engler et al. (2022) formulation, and the Bailey et al. (2022) CE-specific sampler into clearly delineated modules.

### References

[^1]: Engler, D. T., Worden, C. B., Thompson, E. M., & Jaiswal, K. S. (2022). Partitioning ground motion uncertainty when conditioned on station data. *Bulletin of the Seismological Society of America*, 112(2), 1060–1079. https://doi.org/10.1785/0120210177
[^2]: Engler, D. T., Jaiswal, K. S., & Ganesh, M. (2025). Evaluating the impact of uncertainty in ground motion forecasts for post-earthquake impact modeling applications. *Earthquake Spectra*, 41(1), 524–546. https://doi.org/10.1177/87552930241283201
[^3]: Bailey, M. D., Bandyopadhyay, S., & Nychka, D. (2022). Adapting conditional simulation using circulant embedding for irregularly spaced spatial data. *Stat*, 11(1), e446. https://doi.org/10.1002/sta4.446

### DIFF
--- a/doc/api-reference/openquake.hazardlib.correlation_models.rst
+++ b/doc/api-reference/openquake.hazardlib.correlation_models.rst
@@ -19,6 +19,22 @@ Registry
     :undoc-members:
     :show-inheritance:
 
+Scalable sampling
+-----------------
+
+.. currentmodule:: openquake.hazardlib.correlation_models.circulant_embedding
+
+.. autoclass:: RegularGridLayout
+    :members:
+
+.. autoclass:: CirculantEmbeddingFactor
+    :members:
+
+.. currentmodule:: openquake.hazardlib.correlation_models.local_kriging
+
+.. autoclass:: LocalKrigingFactor
+    :members:
+
 Spatial models
 --------------
 
--- a/doc/underlying-science/conditioning-gmf.rst
+++ b/doc/underlying-science/conditioning-gmf.rst
@@ -1,9 +1,9 @@
 Conditioning Ground Shaking
 ===========================
 
-Given an earthquake rupture model, a site model, and a set of ground shaking intensity models (GSIMs), the OpenQuake scenario calculator simulates a set of ground motion fields (GMFs) at the target sites for the requested set of intensity measure types (IMTs). This set of GMFs can then be used in the :ref:`scenario-damage-intro` and :ref:`scenario-risk-intro` calculators to estimate the distribution of potential damage, economic losses, fatalities, and other consequences. The scenario calculators are useful for simulating both historical and hypothetical earthquakes.
+Given an earthquake rupture model, a site model, and a set of ground shaking intensity models (GSIMs), the OpenQuake engine scenario calculator simulates a set of ground motion fields (GMFs) at the target sites for the requested set of intensity measure types (IMTs). This set of GMFs can then be used in the :ref:`scenario-damage-intro` and :ref:`scenario-risk-intro` calculators to estimate the distribution of potential damage, economic losses, fatalities, and other consequences. The scenario calculators are useful for simulating both historical and hypothetical earthquakes.
 
-For historical events, ground motion recordings and macroseismic intensity data ("ground truth") can provide additional constraints on the ground shaking estimates in the region affected by the earthquake. The U.S. Geological Survey ShakeMap system (Worden et al. 2020; Wald et al. 2022) provides near-real-time estimates of ground shaking for earthquakes conditioned on station and macroseismic data where available. Support to read USGS ShakeMaps and simulate GMFs based on the mean and standard deviation values estimated by the ShakeMap — for peak ground acceleration (PGA) and spectral acceleration (SA) at 0.3s, 1.0s, and 3.0s — was added to the OpenQuake-engine in 2018, and a link to the `scenario_damage` and `scenario_risk` calculators was added based on the methodology described by Silva & Horspool (2019). Further support for reading ShakeMaps from sources other than the USGS, support for MMI, and performance improvements were introduced to the engine in 2021.
+For historical events, ground motion recordings and macroseismic intensity data ("ground truth") can provide additional constraints on the ground shaking estimates in the region affected by the earthquake. The U.S. Geological Survey ShakeMap system (Worden et al. 2020; Wald et al. 2022) provides near-real-time estimates of ground shaking for earthquakes conditioned on station and macroseismic data where available. Support to read USGS ShakeMaps and simulate GMFs based on the mean and standard deviation values estimated by the ShakeMap — for peak ground acceleration (PGA) and spectral acceleration (SA) at 0.3s, 1.0s, and 3.0s — was added to the OpenQuake engine in 2018, and a link to the `scenario_damage` and `scenario_risk` calculators was added based on the methodology described by Silva & Horspool (2019). Further support for reading ShakeMaps from sources other than the USGS, support for MMI, and performance improvements were introduced to the engine in 2021.
 
 Therefore, starting from OpenQuake engine v3.16, it is possible to condition the ground shaking to observations, such as ground motion recordings and macroseismic intensity observations. The implementation is tightly coupled with the existing scenario calculator.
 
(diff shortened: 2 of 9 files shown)


raoanirudh:
5 / 500 simulated station-conditioned spatially-correlated PGA fields for the 2014 M6.0 South Napa earthquake
PGA_0
PGA_1
PGA_2
PGA_3
PGA_5

raoanirudh:
5 / 500 simulated station-conditioned spatially-correlated SA(0.3) fields for the 2023 M7.8 Kahramanmaraş earthquake

SA(0 3)_0
SA(0 3)_1
SA(0 3)_2
SA(0 3)_3
SA(0 3)_4

raoanirudh:
1 / 500 simulated station-conditioned spatially-cross-correlated PGA and SA(0.3) fields for the 2023 M7.8 Kahramanmaraş earthquake

**PGA**
PGA_0

**SA(0.3)**
SA(0 3)_0

micheles (approved):

PR #11739, opened 2026-09-07
Supporting aftershock terms in GMMs when using delta_rates input option

CB-quakemodel:
A handful of GMMs in the engine have aftershock terms (e.g., CY08, Bradley 2013), but currently, they are all turned off (e.g., the `AS` term in CY08 is hardcoded to false despite influencing the GMM's behaviour if set to true).

This PR modifies the existing `delta_rates`-based aftershock calculation approach (applying rate adjustments to ruptures labelled as aftershocks) by setting `ctx.is_aftershock=True` for such ruptures, so that GMMs can be informed of if an event is to be modelled an aftershock (e.g., setting `AS=True` in CY08 when needed), and otherwise a mainshock as usual (default of `ctx.is_aftershock=False`).

The existing test case (aftershock/case_1) results are updated - note that I now include the Bradley 2013 GMM to provide test coverage for the aftershock term too in this GMM too.

For context, here is the aftershock/case_1 hazard curves with ONLY CY08 (rather than also Bradley 2013 as now included), but with and without the change introduced in this PR of using aftershock term when required. The changes make sense - some of the ruptures to which rates deltas are applied now also generate lower ground-motion intensities because the aftershock term is being applied when appropriate now:
observed_changes_case_1_aftershock

--------------------------------------------

The ASK14 GMM includes an aftershock term (not currently present), but it will also require a "centroid joyner-boore" distance to be available for each rupture to use it properly - I will add this in a sequential PR.

This work (part 1/2 on engine side) is to support Zainab's aftershock residual analysis - I will add the required changes to the SMT too (reading in aftershock flag + Crjb parameters from a flatfile).


micheles (approved):

openquake/hazardlib/contexts.py
```diff
@@ -1534,10 +1538,15 @@ def gen_ctxs(self, src):
...
+                # Every rup listed in delta_rates is an aftershock, so
+                # shift its occurrence_rate and set is_aftershock as True
+                # so aftershock terms are applied in GMMs containing them
                 with self.cmaker.delta_mon:
                     delta = self.cmaker.deltagetter(src.id)
-                    ctx.occurrence_rate += delta[ctx.rup_id]
+                    d = delta[ctx.rup_id]
+                    ctx.occurrence_rate += d
+                    if 'is_aftershock' in ctx.dtype.names:
+                        ctx.is_aftershock = d != 0
```
micheles, line 1549: Notice that `is_aftershock` should be a bool but is a float64 instead, because you forgot to specify the type in `site_param_dt`. Apart from that, LGTM to me.

CB-quakemodel: Thanks, I have just fixed this and updated changelog

PR #11742, opened 2026-09-08
Add the aftershock term to ASK14 GMM + specification of Crjb in delta_rates CSV for each aftershock

CB-quakemodel:
The ASK14 GMM did not have its aftershock term implemented, so I add this here.

For this term, the centroid Joyner-Boore (Crjb) distance is required. Crjb is the shortest horizontal distance from the aftershock rupture centroid to the edge of the main shock rupture.

Given the use of a `delta_rates` CSV provides a per aftershock rupture delta to the associated rate of occurrence (and it is how aftershock calcs are ran currently in the engine), this is the ideal place to specify the associated Crjb for each aftershock too (we don't need any info on the mainshock here, just the Crjb distance which would also be computed in a preprocessing).

I add a unit test (aftershock/case_2) to test this feature.


openquake/hazardlib/site.py
```diff
@@ -323,7 +323,8 @@ def _extract(array_or_float, indices):
     'custom_site_id': (numpy.bytes_, 8),
     'region': numpy.uint32,
     'in_cshm': bool,  # used in mcverry
-    'is_aftershock': bool
+    'is_aftershock': bool,
```
micheles: please add a comment with where it is used

micheles (approved):

Issue #11740, opened 2026-09-08
Investigate replacing the DbServer with a FastAPI server


micheles:
Pilot project. If it works well, the next step is migrating the scientific API from Django to FastAPI.

PR #11747, opened 2026-09-09
Introduced a minimal FastAPI server

micheles:
Part of #11740


vot4anto:
@micheles we do not need to add the wheel on our wheelhouse?

micheles:
Yes, @vot4anto we need to add the wheels to our wheelhouse, this is why I set you up as a reviewer

PR #11749, opened 2026-09-09
Prevent SQLite from reusing IDs of restored jobs

ptormene:
In the QGIS plugin integration tests we run oq restore to have the possibility to test loading in QGIS the outputs of the oq demos. Since yesterday, without this fix, oq restore updated database paths for the restored files but did not ensure SQLite’s internal auto-increment counter knew about their IDs. If the database sequence was stale, the next job could receive an ID already used by a restored calc_<id>.hdf5 file.

The change:

  1. Collects all IDs found in restored datastore filenames.
  2. Finds the largest restored ID.
  3. Updates SQLite’s sqlite_sequence for the job table to at least that value.
  4. Creates the sequence entry if it does not exist.

  Therefore, the next automatically created job gets an ID greater than every restored datastore ID.

To reproduce the error:
```
oq reset -y
oq engine --run demos/hazard/AreaSourceClassicalPSHA/job.ini
oq dump /tmp/demo.zip
oq reset -y
oq restore /tmp/demo.zip ~/oqdata
oq engine --run demos/hazard/AreaSourceClassicalPSHA/job.ini
```
Without this fix, it causes:
`RuntimeError: There is a pre-existing file /home/ptormene/oqdata/calc_1.hdf5`

Checking out commit 3367c018d4 from 2 days ago, the error does not occur.

### DIFF
--- a/openquake/commands/restore.py
+++ b/openquake/commands/restore.py
@@ -52,15 +52,27 @@ def main(archive, oqdata):
     db = Db(sqlite3.connect, dbpath, isolation_level=None,
             detect_types=sqlite3.PARSE_DECLTYPES)
     n = 0
+    calc_ids = []
     for fname in os.listdir(oqdata):
         mo = re.match(r'calc_(\d+)\.hdf5', fname)
         if mo:
             job_id = int(mo.group(1))
+            calc_ids.append(job_id)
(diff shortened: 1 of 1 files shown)


ptormene:
Superseded by https://github.com/gem/oq-engine/pull/11752

PR #11751, opened 2026-09-09
Correlation modules refactor #16 – Abrahamson et al. (2014)

raoanirudh:
## Abrahamson et al. (2014) between-event and within-event cross-IM correlation models

Related to #11230. This PR adds separate cross-IMT correlation models for both between-event residuals and within-event residuals based on Abrahamson, Silva, and Kamai (2014)[^1], aka ASK14:

- `AbrahamsonEtAl2014BetweenEvent`
- `AbrahamsonEtAl2014WithinEvent`

Prioritizing this model because of the need for a between-event cross-IM correlation model supporting the commonly used combination of PGA, SA, and PGV. Published residual-specific between-event models with this coverage are scarce, while the ASK14 electronic supplement provides separate correlation tables for both between-event (`tau`) and within-event (`phi`) residuals derived from NGA-West2 data. Eg. the existing model in the engine of Goda and Atkinson (2009)[^2] is regionally calibrated for Japan, supports only SA from 0.1–5 s, does not support PGV, and does not explicitly support PGA either – the engine relies on an out-of-domain SA(0.05) proxy to handle PGA with this model. Baker and Bradley (2017)[^3] does cover PGA, SA, and PGV but the publication provides only total-residual correlation, it does not provide separate tabulated values for the between-event correlations.

Adding an explicit cross-IM correlation model for between-event residuals helps avoid the approximation of substituting a total-residual cross-IM correlation model for between-event cross-IM correlation.

### Implementation

The reference used for the implementation is the spreadsheet [ASK14_coeffs.xls](https://onlinelibrary.wiley.com/action/downloadSupplement?doi=10.1193%2F070913EQS198M&file=esp4bf00365-sup-0001.xls), which is provided as an electronic supplement to Abrahamson et al. (2014). The cross-IM correlation models support RotD50 PGA, PGV, and 5%-damped SA at periods from 0.02 to 10 seconds. The published tables contain 22 SA periods. Intermediate periods such as SA(0.6) are supported by interpolation in log-period correlation-vector space, followed by normalization. This reproduces the tables exactly at the periods listed in the electronic supplement, while preserving symmetry, a unit diagonal, and positive-semidefinite correlation matrices if interpolated periods are requested.

The published within-event matrix is already positive definite and is kept unchanged from the electronic supplement. The published between-event matrix, however, has a minimum eigenvalue of approximately −0.01972 and therefore cannot be sampled directly. It is adjusted deterministically using the nearest-correlation procedure of Higham (2002)[^4] with a small positive eigenvalue floor. The maximum change to any published coefficient is approximately 0.00389. Both the original table and the adjustment are tested.

Both classes are added in the same module because they come from the same publication and share their IMT grid. They are registered as separate models so that their residual components cannot be confused.

The common correlation interface now permits a cross-IMT-only model to be selected as a within-event model. Such models provide rectangular same-location covariance blocks for conditioning and factor only their small $$M \times M$$ IMT matrix. The factor is reused over all $$N$$ sites, avoiding construction of an unnecessary $$MN \times MN$$ matrix.

**Note 1**: The within-event class represents same-site cross-IMT correlation only. It applies the same IMT correlation independently at every site and does not introduce spatial correlation between sites.

**Note 2**: Although these correlations were derived from ASK14 residuals and their ideal application is therefore with the `AbrahamsonEtAl2014` GMM, the engine does not warn about or prohibit their use with other GMMs as the impact of using them with other GMMs is not expected to be significant.

### Verification tests

Reference coefficients were extracted independently from the publisher’s [electronic supplement](https://onlinelibrary.wiley.com/action/downloadSupplement?doi=10.1193%2F070913EQS198M&file=esp4bf00365-sup-0001.xls).

In addition to testing the extracted published coefficients, the tests also check the residual-component and IMT metadata, perform IMT validation, check interpolation at SA(0.6) and symmetry, unit diagonal, and positive definiteness

### References
[^1]: Abrahamson, N. A., Silva, W. J., & Kamai, R. (2014). Summary of the ASK14 ground motion relation for active crustal regions. _Earthquake Spectra_, _30_(3), 1025–1055. https://doi.org/10.1193/070913EQS198M 
[^2]: Goda, K., & Atkinson, G. M. (2009). Probabilistic characterization of spatially correlated response spectra for earthquakes in Japan. _Bulletin of the Seismological Society of America_, _99_(5), 3003–3020. https://doi.org/10.1785/0120090007
[^3]: Baker, J. W., & Bradley, B. A. (2017). Intensity Measure Correlations Observed in the NGA-West2 Database, and Dependence of Correlations on Rupture and Site Parameters. _Earthquake Spectra_, _33_(1), 145–156. https://doi.org/10.1193/060716EQS095M
[^4]: Higham, N. J. (2002). Computing the nearest correlation matrix—a problem from finance. _IMA Journal of Numerical Analysis_, _22_(3), 329–343. https://doi.org/10.1093/imanum/22.3.329


CB-quakemodel (approved):
LGTM

PR #11761, opened 2026-09-11
Ported more endpoints to FastAPI

micheles:
Ported validate and run_scenario. Part of https://github.com/gem/oq-engine/issues/11740.

### DIFF
--- a/openquake/server/api.py
+++ b/openquake/server/api.py
@@ -20,6 +20,7 @@
 
 import json
 import logging
+import multiprocessing as mp
 import os
 import re
 import secrets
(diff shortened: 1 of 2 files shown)


openquake/server/api.py
```diff
@@ -79,6 +81,79 @@ async def v0_worker(
     return getattr(master, action)()
 
 
+async def _validate_uploaded_file(request, field, missing_message):
+    """Validate one uploaded calculation file."""
+    form = await request.form()
+    files = {
+        key: value for key, value in form.multi_items()
+        if hasattr(value, 'file')}
+    from openquake.server.views import get_uploaded_file_path, validate_job
```
micheles, line 90: I will need to fix this import in the future

openquake/server/api.py
```diff
@@ -79,6 +81,79 @@ async def v0_worker(
...
+        rup_id: int, request: Request,
+        x_api_key: str | None = Header(default=None)):
+    """Run a papers scenario calculation."""
+    _check_api_key(x_api_key)
+    form = await request.form()
+    username = form.get('username')
+    if not username:
+        raise HTTPException(status_code=400,
+                            detail='Missing calculation owner')
+    from openquake.server.papers import base as papers
```
micheles, line 130: I will need to fix this import in the future

PR #11765, opened 2026-09-14
use folder unified on requirements files and action workflow

vot4anto:


vot4anto:
@nastasi-oq also the weekly tests are green: https://github.com/gem/oq-engine/actions/runs/34827881825

PR #11767, opened 2026-09-14
Extend site amp logic trees to support disagg

CB-quakemodel:
In https://github.com/gem/oq-engine/pull/11726 we added the ability to specify a logic tree on the amplification model used (to support recent site response methodologies advocating for multiple site amplification models in a logic tree approach).

Per-branch amp contributions are computed by integrating each branch's amplification function on the same bedrock-IML grid the classical convolution uses, then aggregated into the disagg epsilon bins.

Per-realisation disagg is therefore consistent (to within ~0.1%) with hazard curves computed from the same site amplification model logic tree (tested in logictree/case_27).

When `individual_rlzs = true` the per-site disagg CSVs correctly propagate the amp legs — e.g. logictree/case_27 produces 3 rlz columns from 1 SSC × 1 GMM × 3 amp.

epsilon_star and mutex sources are unsupported with amplification and rejected early (in oqvalidation.py and disaggregation.compute method respectively), with this being documented here too.


openquake/calculators/tests/logictree_test.py
```diff
@@ -527,6 +527,26 @@ def test_case_26(self):
...
+            for fname in got:
+                self.assertEqualFiles(
+                    'expected/%s%s' % (prefix, strip_calc_id(fname)), fname)
+
+        # Full-enum hcurves must match case_26 (classical vs disagg)
+        self.run_calc(case_27.__file__, 'job.ini')
+        hcurves = self.calc.datastore['hcurves-stats'][:, 0]
+        self.run_calc(case_26.__file__, 'job.ini')
+        expected = self.calc.datastore['hcurves-stats'][:, 0]
+        aac(hcurves, expected, rtol=1e-6)
```
micheles: Does the test become particularly slow? you are repeating the calculation several times

openquake/hazardlib/calc/disagg.py
```diff
@@ -441,7 +536,7 @@ def init(self, magi, src_mutex,
             return sum(self.weights)
         return 1.
 
-    def _disagg6D(self, imldic, g):
+    def _disagg6D(self, imldic, g, rlz=None):
         # returns a 6D matrix of shape (D, Lo, La, E, M, P)
```
micheles, line 573: Does `rlz=None` make sense? In `openquake/hazardlib/calc/disagg.py` investigate the call `mat5 = self._disagg6D(imtls, g)[..., 0]`, is rlz=None causing a failure in `self.amplifier.rlz_ampl_ord[None]`?
You should make `rlz` mandatory in `_disagg6`

openquake/hazardlib/calc/disagg.py
```diff
@@ -232,19 +232,111 @@ def _disaggregate(ctx, mea, std, cmaker, g, iml2, bin_edges, eps4, epsstar, gp,
...
+    :param iml2: log soil IMLs of shape (M, P)
+    :param amplifier: an :class:`Amplifier` for the rlz's amp branch
+    :param ampcode: 2-letter code identifying the site's amplification entry
+    :returns: a disagg matrix (6D array)
+    """
+    if epsstar:
+        # Epsilon is binned on rock predictions, so with amplification the
+        # binning corresponds to pre-amp bedrock rather than post-amp soil;
+        # supporting epsilon_star here would misrepresent per-bin contribution
+        raise NotImplementedError(
```
micheles: This is fine for usage in hazardlib, but in the engine we want to fail early, i.e. the combination `epsilon_star` and `amplifier` must be rejected in `oqvalidation.py`, *before* starting the calculation.

openquake/hazardlib/calc/disagg.py
```diff
@@ -452,12 +547,30 @@ def _disagg6D(self, imldic, g):
             imlog2[m] = to_distribution_values(iml2[m], imt)
         mea, std = self.mea[self.magi], self.std[self.magi]
         gp = self.src_mutex.get('grp_probability', 1.)
+        amp = None
+        if self.amplifier is not None:
+            if self.src_mutex:
+                raise NotImplementedError(
```
micheles: This also can be checked before starting the disaggregation

micheles:
For every RLZ/IMT/IML, summing the disaggregation matrix over M/D/Lon/Lat/Eps must give the same hazard rate as the classical calculation with the same amplification model. Do we have a test for that?

CB-quakemodel:
> For every RLZ/IMT/IML, summing the disaggregation matrix over M/D/Lon/Lat/Eps must give the same hazard rate as the classical calculation with the same amplification model. Do we have a test for that?

A check on this has been added to case_27 - now this QA test checks that summing the disagg matrix reproduces same PoEs obtained from classical calc with same amplification logic tree

openquake/hazardlib/calc/disagg.py
CB-quakemodel: I've added a similar check in oqvalidation to prevent conventional uses (a regular disagg calc) from having to wait until here to fail. I can retain the check previously here (same for presence of mutex sources) if you think best to have a hazardlib-level check for each non-permitted combination as well.

openquake/hazardlib/calc/disagg.py
CB-quakemodel: Please see my comment about checks on epsilon star being moved to oqvalidation above.

openquake/calculators/tests/logictree_test.py
CB-quakemodel: Now the calculation is not repeated. The full enumeration disagg is stored and then checked without repetition against the classical results.

micheles (approved):

PR #11775, opened 2026-09-15
Add banimahd turkiye 2026

banimahd:
This PR adds a new ANN-based ground-motion model for Turkiye (Banimahd et al., 2026).

Summary:
- New GSIM module: `banimahd_turkiye_2026.py`
- ONNX model bundled in `banimahd_turkiye_2026_data/onnx_models/`
- Standard deviations in `banimahd_turkiye_2026_data/stds.csv`
- Unit tests: `banimahd_turkiye_2026_test.py` (2 passed)
- Updated `doc/api-reference/openquake.hazardlib.gsim.rst`
- Updated `CONTRIBUTORS.txt`
- Updated `debian/changelog`

Reference:
Banimahd A, Karimzadeh S, et al. (2026).
"Artificial neural network-based non-parametric ground motion models for multiple intensity measures in Turkiye."
Engineering Applications of Artificial Intelligence.


openquake/hazardlib/tests/gsim/data/BANIMAHD2026TURKIYE/.gitkeep
CB-quakemodel, line 1: please remove this file

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+                period = float(imt_str[3:-1])
+                out_idx = 7 + self._PERIODS.index(period)
+                key = f"ln(PSA={period})"
+            else:
+                raise ValueError(f"IMT {imt_str} not supported")
+
+            mean[m, :] = out[:, out_idx]
+            sig[m, :]  = _PHI_TOTAL[key]     # TOTAL
+            tau[m, :]  = _TAU_INTER[key]     # INTER-EVENT
+            phi[m, :]  = _SIGMA_INTRA[key]   # INTRA-EVENT
```
CB-quakemodel: You need an empty line at the end of the file

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+        # Extract scalar values from ctx
+        mag = np.atleast_1d(ctx.mag).astype(float)[0]
+        hypo_depth = np.atleast_1d(ctx.hypo_depth).astype(float)[0]
+        rake_val = np.atleast_1d(ctx.rake).astype(float)[0]
+
+        rjb_arr = np.atleast_1d(ctx.rjb).astype(float)
+        vs30_arr = np.atleast_1d(ctx.vs30).astype(float)
+
+        # Broadcast to length N
+        if len(rjb_arr) != N:
```
CB-quakemodel: these 4 lines are redundant - rjb and Vs30 will already be length N given always in the ctx is one per site (i.e., one per index position in the ctx object)

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+            rjb_arr = np.resize(rjb_arr, N)
+        if len(vs30_arr) != N:
+            vs30_arr = np.resize(vs30_arr, N)
+
+        # Build input arrays
+        Mw = np.full(N, mag, dtype=float)
```
CB-quakemodel: Again, ctx.mag will already be the correct shape of length N

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+        RJB = rjb_arr
+        Vs30 = vs30_arr
+        FD = np.full(N, hypo_depth, dtype=float)
```
CB-quakemodel: Same for focal depth, already of correct shape as taken from ctx

I would suggest:

```
mw = ctx.mag
rjb = ctx.rjb
vs30 = ctx.vs30
fd = ctx.hypo_depth
```

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+
+        # Run ONNX inference once for all 25 outputs
+        sess = _get_session()
+        input_name = sess.get_inputs()[0].name
+        output_name = sess.get_outputs()[0].name
+        out = sess.run([output_name], {input_name: X})[0]
+
+        # Fill mean/sig/tau/phi for each requested IMT
+        for m, imt in enumerate(imts):
+            imt_str = str(imt)
```
CB-quakemodel: you can use `imt.name` instead. This would avoid the need to split strings on line 218 as well

e.g., you can use:

`if imt.name == "PGA"`, `if imt.name=="SA"` and such to handle lines 212 to 219 instead

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+            vs30_arr = np.resize(vs30_arr, N)
+
+        # Build input arrays
+        Mw = np.full(N, mag, dtype=float)
+        RJB = rjb_arr
+        Vs30 = vs30_arr
+        FD = np.full(N, hypo_depth, dtype=float)
+
+        # Fault mechanism
+        if rake_val < -30.0:
```
CB-quakemodel: Here it would be good to add comments explaining which style of faulting (e.g. strike-slip, reverse) is represented by each condition.

Also, I think this style of faulting scheme does not cover strike slip where strike is close to 180 degrees, it only covers where it's close to zero degrees.

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+        tau : np.ndarray
+            Output array for inter-event stddev, shape (M, N).
+        phi : np.ndarray
+            Output array for intra-event stddev, shape (M, N).
+        """
+        # Number of sites
+        N = mean.shape[1]
+
+        # Extract scalar values from ctx
+        mag = np.atleast_1d(ctx.mag).astype(float)[0]
```
CB-quakemodel: This is a problem - you are in effect flattening a ctx's rupture parameters to only the first value. In some situations, the rupture-specific properties e.g. magnitude, hypocentral depth can vary with each index in the ctx object

This is an issue for all values obtained from line 171 to 177

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+            Output array for mean values, shape (M, N).
+        sig : np.ndarray
+            Output array for total stddev, shape (M, N).
+        tau : np.ndarray
+            Output array for inter-event stddev, shape (M, N).
+        phi : np.ndarray
+            Output array for intra-event stddev, shape (M, N).
+        """
+        # Number of sites
+        N = mean.shape[1]
```
CB-quakemodel: I would use `len(ctx)` as it is a much clearer way of saying "this is the number of sites" than taking shape of the array of mean GMM predictions

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+
+        # Build input arrays
+        Mw = np.full(N, mag, dtype=float)
+        RJB = rjb_arr
+        Vs30 = vs30_arr
+        FD = np.full(N, hypo_depth, dtype=float)
+
+        # Fault mechanism
+        if rake_val < -30.0:
+            FM = np.full(N, 1.0, dtype=float)
```
CB-quakemodel: If you use `dtype=float` you are setting dtype to 64 bit, but you're ONNX file needs float32 (32bit), so it should be float32 as well I think. This will be handled already on line 200 when you create the input matrix anyway

CB-quakemodel (changes requested):
Many thanks for your implementation Amir, I have left some comments that need addressing. Once these are addressed, I can perform a second review.

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+    "AVERAGE_HORIZONTAL",
+    "HORIZONTAL",
+    "RANDOM_HORIZONTAL",
+]
+
+for _name in _IMC_CANDIDATES:
+    if hasattr(const.IMC, _name):
+        _IMC_GMEAN = getattr(const.IMC, _name)
+        break
+else:
```
CB-quakemodel: This is a fallback that should be removed. If the requested horizontal component definition is not one supported, it should raise an error instead of silently falling back to something else. Also, it appears to set the first `IMT` (e.g. PGA) as the `IMC`, which does not make sense.

Also, some of the values in `_IMC_CANDIDATES` are not horizontal component definitions as defined in the engine (e.g. `GMEAN`, `AVERAGE_HORIZONTAL`) - you can see those that actually exist here - https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/const.py#L66

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+"Artificial neural network-based non-parametric ground motion models for
+ multiple intensity measures in Turkiye."
+Engineering Applications of Artificial Intelligence.
+
+Model overview
+--------------
+- Trained on strong-motion data from Turkiye.
+- ML regressor: ensemble of 10 feed-forward neural networks.
+- Inputs (in the order used for training and ONNX):
+    1. FD         : focal depth (km)
```
CB-quakemodel: also, based on my comments below, please either use all caps or all lower case for these input variables in your matrix

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+        if not os.path.exists(_ONNX_FILE):
+            raise IOError(f"Cannot find ONNX file: {_ONNX_FILE}")
+        opt = ort.SessionOptions()
+        opt.intra_op_num_threads = 1
+        opt.inter_op_num_threads = 1
+        _SESSION = ort.InferenceSession(_ONNX_FILE, sess_options=opt)
+    return _SESSION
+
+
+__all__ = ["Banimahd2026Turkiye"]
```
CB-quakemodel: You can remove this line

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+            imt_str = str(imt)
+
+            if imt_str == "PGA":
+                out_idx = 0
+                key = "ln(PGA)"
+            elif imt_str == "PGV":
+                out_idx = 1
+                key = "ln(PGV)"
+            elif imt_str.startswith("SA(") and imt_str.endswith(")"):
+                period = float(imt_str[3:-1])
```
CB-quakemodel: To get the period you can use `imt.period` instead which is much safer

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+                out_idx = 7 + self._PERIODS.index(period)
```
CB-quakemodel, line 162: This line is a bit confusing, I guess something is happening within the ONNX file here with the starting on index of 7. Perhaps using a dict for `_PERIODS` which clearly maps IMTs to the matrix rows (e.g. `{"PGA": 0, "PGV": 1}`) could be cleaner

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,228 @@
...
+_load_stddev_tables()
+
+
+# ---------------------------------------------------------------------
+# ONNX session cache
+# ---------------------------------------------------------------------
+_SESSION = None
+
+
+def _get_session():
```
CB-quakemodel: Please use `openquake.baselib.onnx.PicklableInferenceSession` instead - please see mohammadi_turkiye_2023.py for an example.

You should load the ONNX bytes in `__init__` as `self.session`, and load the stddev CSV there too. That drops the _get_session() funtion and  _SESSION global parameter.

 Also, I would suggest gzipping the ONNX file (.onnx.gz) as done so in mohammadi_turkiye_2023.py as well.

CB-quakemodel:
Also, some tests are failing because of formatting issues. The following issues need resolving:
{7D87D339-3D08-4C1E-885A-06E257CAC256}

banimahd:
Hi Christopher,

Thank you for the review and the detailed comments. I have addressed all of them:

- Removed the unnecessary `.gitkeep` file
- Added an empty line at the end of `banimahd_turkiye_2026.py`
- Removed the redundant broadcasting lines for `rjb` and `vs30`
- Used `len(ctx)` for the number of sites
- Used `ctx` arrays directly (`mw`, `rjb`, `vs30`, `fd`, `rake`)
- Added comments explaining the fault mechanism
- Extended the strike-slip condition to cover rake values near ±180°
- Used `imt.string` instead of `str(imt)`
- Ensured `float32` is used consistently
- Used `const.IMC.GEOMETRIC_MEAN` directly
- Used consistent lower-case for input variable names
- Removed the `__all__` declaration
- Used `imt.period` instead of parsing the string
- Added a comment explaining the ONNX output indexing
- Replaced `_get_session` with `PicklableInferenceSession` as in `mohammadi_turkiye_2023.py`
- Compressed the ONNX file to `.onnx.gz`
- Fixed the docstring formatting
- Fixed the duplicate label issue

The unit tests still pass: `2 passed`.

Please let me know if there is anything else you would like me to address.

Best regards,
Amir

openquake/hazardlib/tests/gsim/banimahd_turkiye_2026_test.py
```diff
@@ -0,0 +1,18 @@
...
+        self.check(
+            "BANIMAHD2026TURKIYE/BANIMAHD2026TURKIYE_MEAN.csv",
+            max_discrep_percentage=0.1,
+        )
+
+    def test_std_total(self):
+        self.check(
+            "BANIMAHD2026TURKIYE/BANIMAHD2026TURKIYE_STD_TOTAL.csv",
+            max_discrep_percentage=0.1,
+        )
```
CB-quakemodel: An empty linespace is needed here too

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,195 @@
...
+    - PGV       : ln(cm/s)
+"""
+
+import os
+import csv
+import numpy as np
+import gzip
+from openquake.baselib.onnx import PicklableInferenceSession
+
+from openquake.hazardlib.gsim.base import GMPE
```
CB-quakemodel, line 36: please group all openquake imports together in the second "chunk" of imports

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,195 @@
...
+    This GSIM wraps an ensemble of 10 feed-forward neural networks exported
+    to a single ONNX file. It returns the mean ln(IM) and standard deviations
+    (intra-event, inter-event, total) for each requested IMT.
+    """
+
+    DEFINED_FOR_TECTONIC_REGION_TYPE = const.TRT.ACTIVE_SHALLOW_CRUST
+
+    DEFINED_FOR_INTENSITY_MEASURE_TYPES = {PGA, PGV, SA}
+
+    DEFINED_FOR_INTENSITY_MEASURE_COMPONENT = _IMC_GMEAN
```
CB-quakemodel: if this is now constant, you can just set:

`DEFINED_FOR_INTENSITY_MEASURE_COMPONENT = const.IMC.GEOMETRIC_MEAN`

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,195 @@
...
+from openquake.baselib.onnx import PicklableInferenceSession
+
+from openquake.hazardlib.gsim.base import GMPE
+from openquake.hazardlib import const
+from openquake.hazardlib.imt import PGA, PGV, SA
+
+# ---------------------------------------------------------------------
+# Choose a suitable IM component, robust across OQ versions
+# ---------------------------------------------------------------------
+_IMC_GMEAN = const.IMC.GEOMETRIC_MEAN
```
CB-quakemodel: Based on my other comment you can remove this

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,195 @@
...
+        for row in reader:
+            key = row["ID"]
+            _SIGMA_INTRA[key] = float(row["Sigma"])
+            _TAU_INTER[key] = float(row["Tau"])
+            _PHI_TOTAL[key] = float(row["Phi"])
+
+
+_load_stddev_tables()
+
+
```
CB-quakemodel: Please remove these many empty lines (you can keep 2 empty lines between line 80 and the GSIM class comment)

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,195 @@
...
+    DEFINED_FOR_REFERENCE_VELOCITY = 760.0
+
+    REQUIRES_RUPTURE_PARAMETERS = {"mag", "hypo_depth", "rake"}
+    REQUIRES_DISTANCES = {"rjb"}
+    REQUIRES_SITES_PARAMETERS = {"vs30"}
+
+    _PERIODS = [0.03, 0.05, 0.075, 0.1, 0.15, 0.2, 0.25, 0.3, 0.4,
+                0.5, 0.75, 1.0, 1.5, 2.0, 2.5, 3.0, 3.5, 4.0]
+    def __init__(self, **kwargs):
```
CB-quakemodel, line 83: Please add a space here

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,195 @@
...
+
+    DEFINED_FOR_STANDARD_DEVIATION_TYPES = {
+        const.StdDev.INTRA_EVENT,
+        const.StdDev.INTER_EVENT,
+        const.StdDev.TOTAL,
+    }
+
+    DEFINED_FOR_REFERENCE_VELOCITY = 760.0
+
+    REQUIRES_RUPTURE_PARAMETERS = {"mag", "hypo_depth", "rake"}
```
CB-quakemodel, line 74: please use consistently one line space between each attribute defined in the class

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,195 @@
...
+
+# ---------------------------------------------------------------------
+# Choose a suitable IM component, robust across OQ versions
+# ---------------------------------------------------------------------
+_IMC_GMEAN = const.IMC.GEOMETRIC_MEAN
+
+# ---------------------------------------------------------------------
+# Paths
+# ---------------------------------------------------------------------
+_DATA_DIR = os.path.join(
```
CB-quakemodel: Can `_DATA_DIR`'s path fit on one line?

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,195 @@
...
+
+# ---------------------------------------------------------------------
+# Load standard deviation tables from stds.csv
+# ---------------------------------------------------------------------
+_SIGMA_INTRA = {}
+_TAU_INTER = {}
+_PHI_TOTAL = {}
+
+
+def _load_stddev_tables():
```
CB-quakemodel: Any files required for a GSIM should be loaded at instantiation time of the GSIM class, so this should be done within the `__init__` method. Please see here for an example - https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/gsim/mohammadi_turkiye_2023.py#L217

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,195 @@
...
+import csv
+import numpy as np
+import gzip
+from openquake.baselib.onnx import PicklableInferenceSession
+
+from openquake.hazardlib.gsim.base import GMPE
+from openquake.hazardlib import const
+from openquake.hazardlib.imt import PGA, PGV, SA
+
+# ---------------------------------------------------------------------
```
CB-quakemodel: Given my comments in this second review, I think all of these comments that use three lines can be removed for conciseness

CB-quakemodel:
Dear Amir,

Thanks for addressing my initial comments. I have now performed a second review with some smaller changes to address (most are just formatting/cleaning up). 

Also, the tests on the remove CI are now failing due to the gzipped ONNX file not being found, so please resolve this too:

{61511AC5-E215-43D3-8685-1DF7ED7887BF}

Many thanks,

Christopher

CB-quakemodel:
Hi Amir, thank you for making the requested changes. Before I rerun the tests, I would suggest you pull from master of the oq-engine and then push to this branch (including resolving the changelog conflict it has been updated in master since you modified it in this PR, so you need to adjust it to include these updates too).

Once all the tests on the continuous integration side are green, I can approve the PR.

openquake/hazardlib/gsim/banimahd_turkiye_2026_data/onnx_models/.gitkeep
CB-quakemodel, line 1: Please remove this file - it is not needed

banimahd:
Hi Christopher,

I have addressed all the comments from your second review, including:

- Removed the empty lines
- Used `const.IMC.GEOMETRIC_MEAN` directly
- Removed the `_IMC_GMEAN` block
- Loaded stddev tables in `__init__` instead of a module-level function
- Simplified three-line comments to single-line comments
- Fixed the ONNX file path (moved `GMM_Turkiye_2026.onnx.gz` into `onnx_models/`)
- Pulled from master and resolved the conflict in `debian/changelog`

The tests pass locally: `2 passed`.

Please let me know if there is anything else you would like me to address.

Best regards,
Amir

CB-quakemodel (approved):

CB-quakemodel:
Hi Amir, the tests are still not passing on the CI.

The first problem, is you need to replace 

        input_name = self.session.get_inputs()[0].name
        output_name = self.session.get_outputs()[0].name
        out = self.session.run([output_name], {input_name: X})[0]

with 

        input_name = self.session.get_inputs()[0].name
        out = self.session.run(None, {input_name: X})[0]
        
This is causing the actual unit tests to fail, so I'm not sure how they are passing locally for you. This fix makes them pass locally for me using your GMM.

Secondly, you need to delete lines 9 and 10 of the GSIM module - the "Reference" label is a duplicate causing the documentation tests to fail.

banimahd:
Hi Christopher,

Thank you for catching these issues. I have addressed both:

1. Replaced the `session.run` call with:
   input_name = self.session.get_inputs()[0].name
   out = self.session.run(None, {input_name: X})[0]

2. Removed the duplicate `Reference` label from the GSIM module.

The tests pass locally: 2 passed.

Please let me know if there is anything else.

Best wishes,
Amir

CB-quakemodel:
Hi Amir, everything looks good now, but we are experiencing an error (on our side) that we need to resolve - it is because you are pushing from a fork.

Could you please do us a favour, and git pull from master then push to this branch, so we can see if we can debug this using a small change just made to the CI tests.

banimahd:
Hi Christopher,

I have synced my branch with master. GitHub shows "This branch is not behind the upstream. No new commits to fetch."

Please let me know if there is anything else I should do.

Best regards,
Amir

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,168 @@
...
+            if imt_str == "PGA":
+                out_idx = 0
+                key = "ln(PGA)"
+            elif imt_str == "PGV":
+                out_idx = 1
+                key = "ln(PGV)"
+            elif imt_str.startswith("SA(") and imt_str.endswith(")"):
+                period = imt.period
+                # The ONNX model returns 25 outputs in this order:
+                #   0: PGA, 1: PGV, 2: Ia, 3: D5-75, 4: D5-95, 5: Tm, 6: CAV
```
CB-quakemodel, line 160: The GMM supports arias intensity, significant duration and CAV, which are supported IMTs for OpenQuake.

If you only want this implementation to support PGA, PGV and SA, I suggest you either add a note in the class docstring e.g. `NOTE: This implementation only supports PGA, PGV and SA, and not the other IMTs it supports as outlined in the GMM's paper` , or preferably, you provide support for these intensity measures too (which would mean adding them to DEFINED_FOR_INTENSITY_MEASURE_TYPES, propagating this to the retrieval of the `key` variable, any necessary unit conversion handling and adding unit tests for these additional intensity measures too). Please note that `Tm` (mean period) is not a supported IMT in the OpenQuake Engine, but that CAV, significant duration and arias intensity are.

If you choose the latter, you can find how the other IMTs are represented in https://github.com/gem/oq-engine/blob/master/openquake/hazardlib/imt.py

openquake/hazardlib/tests/gsim/banimahd_turkiye_2026_test.py
```diff
@@ -0,0 +1,19 @@
...
+class Banimahd2026TurkiyeTestCase(BaseGSIMTestCase):
+    GSIM_CLASS = Banimahd2026Turkiye
+
+    def test_mean(self):
+        self.check(
+            "BANIMAHD2026TURKIYE/BANIMAHD2026TURKIYE_MEAN.csv",
+            max_discrep_percentage=0.1,
+        )
+
+    def test_std_total(self):
```
CB-quakemodel, line 14: If your GMM provides inter-event and intra-event standard deviations, there should be checks for these too as for the total

banimahd:
Hi Christopher,

Thank you for the comments. I have addressed both:

1. Added a note in the class docstring clarifying that this implementation only supports PGA, PGV, and SA.

2. Added tests for the inter-event and intra-event standard deviations:
   - `BANIMAHD2026TURKIYE_STD_INTRA.csv` with `INTRA_EVENT_STDDEV`
   - `BANIMAHD2026TURKIYE_STD_INTER.csv` with `INTER_EVENT_STDDEV`
   - Added `test_std_intra` and `test_std_inter` to the test module.

All 4 tests pass locally: `4 passed`.

Please let me know if there is anything else.

Best regards,
Amir

openquake/hazardlib/gsim/banimahd_turkiye_2026.py
```diff
@@ -0,0 +1,171 @@
...
+    
+    REQUIRES_DISTANCES = {"rjb"}
+    
+    REQUIRES_SITES_PARAMETERS = {"vs30"}
+
+    _PERIODS = [0.03, 0.05, 0.075, 0.1, 0.15, 0.2, 0.25, 0.3, 0.4,
+                0.5, 0.75, 1.0, 1.5, 2.0, 2.5, 3.0, 3.5, 4.0]
+    
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
```
micheles: GMPE does not define `__init__`, and the engine documentation explicitly recommends not calling `super().__init__()` in GMPE subclasses. Since this GSIM has no constructor parameters, could this simply be `def __init__(self):` without `**kwargs` and `super().__init__(**kwargs)`?

banimahd:
Hi Chris,

Just a quick note: my system clock was set incorrectly during part of my work on this PR. I have now corrected it. I am not sure if this is related to the CI issues you are seeing, but I wanted to mention it just in case.

Best regards,
Amir

micheles (approved):

micheles:
The Windows failure is unrelated, the PR is good as it is. Thanks for your hard work.

banimahd:
Hi Michele and Chris,

Thank you very much for your time, patience, and guidance throughout this process. I really appreciate your support. I am glad the implementation is now part of OpenQuake.

Best regards,
Amir

PR #11779, opened 2026-09-16
Use --upgrade-db on Docker

vot4anto:


vot4anto:
Dockerfile does not use install.py

vot4anto:
@micheles :  https://github.com/gem/oq-engine/actions/runs/35144520083

micheles (approved):

PR #11784, opened 2026-09-17
Generate impact reports (adapted to FastAPI)

ptormene:


bin/extract_impact_reports.py
```diff
@@ -0,0 +1,35 @@
+import sys
+import subprocess
+from pathlib import Path
+from openquake.commonlib import datastore
+
+
+def main(calc_id: int):
+    dstore = datastore.read(calc_id)
```
micheles, line 8: Better a `with datastore.read(calc_id) as dstore:` block

openquake/calculators/postrisk/make_impact_reports.py
```diff
@@ -0,0 +1,348 @@
...
+
+
+def _open_dstore(dstore):
+    """
+    Resolve the dstore argument (path/int from CLI, or an open Datastore),
+    returning (dstore, calc_id).
+    """
+    if isinstance(dstore, (str, int)):
+        # NOTE: called from the command line
+        from openquake.commonlib import datastore
```
micheles, line 137: Why not at top-level, as usual?

openquake/calculators/postproc/plots.py
```diff
@@ -472,3 +478,240 @@ def get_assetcol(calc_id):
...
+    return labels
+
+
+@functools.lru_cache(maxsize=1)
+def _read_basemap(basemap_path):
+    import rasterio
+    with rasterio.open(basemap_path) as src:
+        return src.read(), src.transform, src.crs
+
+
```
micheles, line 569: Can you add a comment about what the basemap contains (the whole world?), which format is it, from where you got it, and what it happens if it is missing?

openquake/calculators/postrisk/make_impact_reports.py
```diff
@@ -0,0 +1,348 @@
...
+        disclaimer_txt=DISCLAIMER_TXT,
+        basemap_path=_get_basemap_path(), threshold_deg=threshold_deg,
+        no_uncertainty=no_uncertainty, loss_metric=loss_metric)
+
+    return event_ctx, report_opts, losses_df, iso3_codes, time_of_calc
+
+
+@contextmanager
+def _job_log_handler(calc_id):
+    # Forward report logs to the database while preserving server logs
```
micheles, line 247: I wonder if you are reinventing logs.LogContext; it reeks of ignorance about the existence of
```
with logs.init(dict(job_id=calc_id):
    ...
```
that should do everything you need already.

openquake/calculators/country_impact_report_builder.py
```diff
@@ -0,0 +1,778 @@
...
+            affected = aggregated_gdf
+        bounds = affected.geometry.total_bounds  # (minx, miny, maxx, maxy)
+        return (
+            [bounds[0] - padding_deg, bounds[2] + padding_deg],
+            [bounds[1] - padding_deg, bounds[3] + padding_deg],
+        )
+
+    def _load_country_info(self):
+        try:
+            countries_info_file = config.directory.countries_info_file
```
micheles, line 272: In general I would like an `impact.cfg`  file containing the full list of possible keys (some of them can be empty, like`config.directory.countries_info_file = `)  so that we know what is possible to override. Can go later, because I also want and `impact.cfg` file, a `server.cfg` file, etc, and such files could be used by `install.py`.

openquake/server/views.py
```diff
@@ -195,13 +203,58 @@ def _get_base_url(request):
     return base_url
 
 
-def get_bool_param(request, name, default=False):
-    val = request.GET.get(name)
+def _get_bool_param(obj, name, default=False):
+    val = obj.get(name)
     if val is None:
         return default
     return str(val).lower() in ('1', 'true', 'yes', '')
```
micheles, line 212: NO! This looks like a merge error, you reintroduce a function which hase been moved in services.py!

bin/extract_impact_reports.py
```diff
@@ -0,0 +1,35 @@
...
+        country_group = impact_group[iso3]
+        if "report_pdf" not in country_group:
+            print(f"No report_pdf found for {iso3}")
+            continue
+        pdf_bytes = country_group["report_pdf"][()]
+        fname = Path(f"impact_report_{calc_id}_{iso3}.pdf")
+        with open(fname, "wb") as f:
+            f.write(pdf_bytes)
+        print(f"Wrote {fname}")
+        subprocess.Popen(["xdg-open", str(fname)])
```
micheles, line 26: xdg-open may not exist, so
```python
if shutil.which("xdg-open"):
    subprocess.Popen(["xdg-open", str(fname)])
```
is better.

micheles:
My other comment is that the tests `OQ_APPLICATION_MODE=IMPACT pytest openquake/server/tests/test_impact_mode.py` are way too slow. That must be addressed, but in a different PR.

PR #11791, opened 2026-09-18
Tried to set LOCKDOWN = restricted()

micheles:
`restricted` is an engine/FastAPI concept, so it should go in openquake.cfg, not in settings.py. It is currently *really* hackish (`config.dbserver.host != '127.0.0.1'`), it can be improved, I just want to see what breaks with inferring LOCKDOWN from `openquake.baselib.restricted()`. This will also be useful for our docker.


micheles:
@ptormene points out that this cannot work as I hoped, since the local_settings.py will override LOCKDOWN and the change will be ignored.

openquake/baselib/__init__.py
```diff
@@ -125,13 +125,13 @@ def read(*paths, **validators):
     os.environ['OQ_DISTRIBUTE'] = config.distribution.oq_distribute
 
 
-def multi_user():
+def restricted():
     return config.dbserver.host != '127.0.0.1'
```
micheles, line 129: This is really ugly

PR #11792, opened 2026-09-18
Move server data downloads into pytest fixtures; fix impact behavior converting json ruptures

ptormene:


openquake/server/tests/conftest.py
```diff
@@ -42,8 +45,95 @@ def copy_from_templates_if_needed(tmpldir, ext):
...
+def _download_file(url, target):
+    """Download *url* atomically, removing incomplete files on failure."""
+    if target.is_file() and target.stat().st_size:
+        return
+    target.unlink(missing_ok=True)
+    fd, temporary = tempfile.mkstemp(
+        prefix=f'.{target.name}.', dir=target.parent)
+    os.close(fd)
+    try:
+        subprocess.run(['wget', url, '-O', temporary], check=True)
```
micheles: It is never a good idea to run subprocesses when you can avoid it; in particular in some systems `wget` is not installed. Download the files using `requests` which we know is a dependency of the engine and therefore always available.

openquake/server/tests/test_impact_mode.py
```diff
@@ -319,7 +319,8 @@ def impact_run_then_remove(
                             f'impact_report_{job_id}_{iso3}.{file_format}')
                         self.assertTrue(os.path.isfile(report))
                         with open(report, 'rb') as f:
-                            self.assertTrue(f.read(len(signature)) == signature)
+                            self.assertTrue(
+                                f.read(len(signature)) == signature)
```
micheles: Why not assertEqual? it will give a better error message in case of failure.

openquake/engine/openquake.cfg
```diff
@@ -108,6 +108,12 @@ shared_dir =
...
 # the directory containing the mosaic models
 mosaic_dir =
+# files and directory used to generate impact reports
+countries_info_file =
+world_cities_file =
+admin1_boundaries_file =
+fonts_dir =
+basemap_file =
 # the file containing the geometries of the mosaic model boundaries
 mosaic_boundaries_file =
```
micheles, line 118: In the future we should have a file `impact.cfg` containing these parameters, but for the moment let's keep things are they are.

micheles (approved):

